PR #25, opened 2020-11-02
Allow referencing other Source fields

philippjfr:
Allows referencing variables from another sources in a Source spec, e.g.:

```
sources:
  ae5:
    type: ae5
    url: pyviz.demo.anaconda.com
    kubernetes_api: https://kube-control.pyviz.demo.anaconda.com
    username: anaconda-enterprise
    password: ...
  session_info:
    type: session_info
    urls: "@ae5.deployments.urls"
```

Here urls on the session_info source will be filled in with the `urls` provided by the ae5 source.

### DIFF
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -94,6 +94,24 @@ class Source(param.Parameterized):
         return df
 
     @classmethod
+    def _resolve_reference(cls, reference, sources={}):
+        refs = reference[1:].split('.')
+        if len(refs) != 3:
+            raise ValueError(f"Reference string '{reference} not valid. "
+                             "Must take the form @source.table.field.")
+        sourceref, table, field = refs
+        source = cls.from_spec(sourceref, sources)
+        table_schema = source.get_schema(table)
+        if field not in table_schema:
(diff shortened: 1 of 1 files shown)


lumen/sources/base.py
```diff
@@ -93,6 +93,24 @@ def _filter_dataframe(cls, df, **query):
...
 
+    @classmethod
+    def _resolve_reference(cls, reference, sources={}):
+        refs = reference[1:].split('.')
+        if len(refs) != 3:
+            raise ValueError(f"Reference string '{reference} not valid. "
+                             "Must take the form @source.table.field.")
+        sourceref, table, field = refs
+        source = cls.from_spec(sourceref, sources)
+        table_schema = source.get_schema(table)
```
jbednar, line 104: Can getting a table not fail? Seems like you're only checking for field and enum not being found, but I guess that's because a table not found will just result in an empty set of results (and thus a missing field)? Seems like the field error message might confuse people in that case, as it indicates that the field was not found in the table but doesn't say anything about whether the table itself was found.

PR #57, opened 2020-12-17
Improve Intake Source

philippjfr:


philippjfr:
@yougis Since you are likely the only real user right now. Do you mind if I simply rename `filename` to `uri` here?

lumen/sources/intake.py
```diff
@@ -10,38 +10,59 @@ class IntakeSource(Source):
...
     """
 
-    catalog = param.String(doc="""
-        The name of the catalog; must be provided.""")
-
     dask = param.Boolean(default=False, doc="""
         Whether to return a dask DataFrame.""")
 
-    filename = param.String(doc="Filename of the catalog file.")
+    filename = param.String(doc="Deprecated alias for the URI")
```
yougis, line 16: do you mean  remove this parameter
and the check in the __init__ above (line 26-29)
 ?
if yes , I am totally agree

philippjfr: Thanks for replying so quickly. Didn't want to break your setup but it's early days for the library so would like to keep things clean.

PR #63, opened 2021-01-05
Add support for env and shell variables

philippjfr:
Adds support for templating env and shell variables into the dashboard yaml.

Here is an example using an env var:

```yaml
config:
  title: "AE5 Monitor"
  layout: tabs
sources:
  ae5:
    type: ae5
    hostname: pyviz.demo.anaconda.com
    username: anaconda-enterprise
    password: {{env("AE5_PASSWORD")}}
```

### DIFF
--- a/docs/dashboard.md
+++ b/docs/dashboard.md
@@ -152,11 +152,17 @@ that table to the `urls` parameter of the `WebsiteSource`.
 
 ### Templating
 
-In many cases you do not want to hardcode variables inside the yaml specification instead passing in variables from an environment variable, a shell command or as a CLI argument. This can be achieved using templating syntax:
+In many cases you do not want to hardcode variables inside the yaml specification instead passing in variables from an environment variable, a shell command, a CLI argument, a HTTP request header or cookie, or a OAuth token variable. This can be achieved using the following templating syntax:
 
 - `{{env("USER")}}`: look in the set environment variables for the named variable
 
 - `{{shell("get_login thisuser -t")}}`: execute the command, and use the output as the value. The output will be trimmed of any trailing whitespace.
+
+- `{{cookie("USER")}}`: look in the HTTP request cookies of the served application
+
+- `{{header("USER")}}`: look in the HTTP request headers of the served application
+
+- `{{oauth("USER")}}`: look in the OAuth user token
 
 - `{{USER}}`: Arguments passed in using `--template-vars="{'USER': 'lumen_user'}"` when using `lumen serve` on the commandline.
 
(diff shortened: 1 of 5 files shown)


docs/dashboard.md
```diff
@@ -150,6 +150,16 @@ The `@csv.websites.url` syntax will look up a `Source` called 'csv',
 request a table called 'websites' and then feed the 'url' column in
 that table to the `urls` parameter of the `WebsiteSource`.
 
+### Templating
+
+In many cases you do not want to hardcode variables inside the yaml specification instead passing in variables from an environment variable, a shell command or as a CLI argument. This can be achieved using templating syntax:
+
+- `{{env("USER")}}`: look in the set environment variables for the named variable
+
+- `{{shell("get_login thisuser -t")}}`: execute the command, and use the output as the value. The output will be trimmed of any trailing whitespace.
```
jbednar, line 159: This looks very useful, but do we now need to offer a safe_load option for the YAML that disables features like this, so that people could safely write a web app for dynamically constructing Lumen apps from an editable (and thus abusable) YAML spec?

lumen/util.py
```diff
@@ -82,3 +85,54 @@ def parse_timedelta(time_str):
         if param:
             time_params[name] = int(param)
     return dt.timedelta(**time_params)
+
+
+def _j_getenv(x):
+    if isinstance(x, Undefined):
+        x = x._undefined_name
+    return os.getenv(x, '')
```
jbednar, line 93: Should the default value be something recognizable to a user, e.g. "UNDEFINED_LUMEN_ENV_VARIABLE", so that they can detect when the environment has not been set up properly to run this app?

jbednar (approved):

lumen/util.py
jbednar: Or maybe "UNDEFINED ENVVAR {x}", so that they can tell what it was looking for?

philippjfr: In most cases they would not detect this either. I've mostly followed intake semantics here.

docs/dashboard.md
philippjfr: Yes, especially during live edits I should not allow this.

philippjfr: Hmm, during live edits is access to environment variables safe? Seems like I should not allow any of these during live editing.

philippjfr: How would you suggest we toggle this outside of live editing? A commandline argument?

philippjfr: Will merge for now, but I agree this is a very valid concern and we have to consider how to communicate this and toggling the behavior.

jbednar: Yes, I'm assuming a commandline argument, because whether such support is appropriate is about where you are launching the server more than about the contents of the dashboard.

jbednar: But I think it should also be available from the Python API interface to lumen, so that a Panel app that allows .yml to be edited can call lumen in safe mode to render it.

PR #71, opened 2021-01-19
Add auth field to dashboard specification

philippjfr:
Adds an `auth` key to the dashboard specification which is checked against the `user_info` provided by the auth provider configured when launching the server. As an example the GitHub OAuth provider returns the `login` of the user that is visiting the dashboard. If we add the following field to the yaml:

```yaml
auth:
  login: [philippjfr]
```

It will check the current user against all users listed here. For a more generic auth mechanism auth providers such as Okta make it possible to return a list of groups a user belongs to in which case you could list the allowed groups to the `auth` field.

An unauthorized user will see this:

Screen Shot 2021-01-19 at 6 14 55 PM

An authorized user will see:

Screen Shot 2021-01-19 at 6 43 19 PM

### DIFF
--- a/docs/dashboard.md
+++ b/docs/dashboard.md
@@ -121,6 +121,29 @@ targets: This is the list of targets to monitor
     ...: Additional parameters passed to the Card layout(s), e.g. width or height
 ```
 
+### `auth`
+
+The `auth` field may provide a dictionary of any number of fields
+which are validated against the user information provided the the Auth
+provider, which is made available by Panel in the
+`panel.state.user_info` dictionary. To discover how to configure an
+Auth provider with Panel/Lumen see the [Panel documentation](https://panel.holoviz.org/user_guide/Authentication.html).
+
+As an example the GitHub OAuth provider returns the login of the user
+that is visiting the dashboard. If we add the following field to the
+yaml:
+
+```yaml
+auth:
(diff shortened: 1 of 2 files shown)


philippjfr:
@jbednar In case you have opinions. Should a (unauthorized) user be able to see their own auth information? Secondly, should a unauthorized user be able to see the specific auth fields that didn't match?

jbednar:
> An authorized user will see:

In most websites it just shows the username, not "User: <username>", presumably because a user's username is generally very recognizable as their own username, so it's not ambiguous.

> Should a (unauthorized) user be able to see their own auth information? Secondly, should a unauthorized user be able to see the specific auth fields that didn't match?

Nowadays it seems like the world has decided to reveal as little information as possible, generally not even acknowledging that an unauthorized resource exists. I assume that's to stop leaking information that helps an attacker, but personally I find it maddening for a legitimate user who is blocked for technical reasons, as it makes debugging your setup very difficult. In this case there's clearly some leakage of the title of the dashboard, which one could imagine being sensitive (E.g. "Reasons why we need to put our competitor out of business"). But if we assume a dashboard it meant to be secure mainly inside an organization's firewall, maybe leaking that isn't so important. In any case we should probably not go out of our way to list other data that could help an attacker.

Issue #81, opened 2021-02-24
Template is now private property of Dashboard


yougis:
in line 341 template object should be access by self._template and not self.template.

philippjfr:
Don't see this issue.

Issue #77, opened 2021-02-10
date as widget filter : datetime fomat from schema  is not recognized properly


yougis:
When we use  filter widget with date as field, the format property in schema dict is 'datetime'.

This is not recognized under the _string_type method. The 2 format option is 'date" or 'date-time' but there is not 'datetime' as possible format option.

Then, the method does also not find  'inclusiveMinimum' and 'inclusiveMaximum' property to set the range.

philippjfr:
Thanks @yougis! Will try to resolve that asap, unless you get a chance before I do.

PR #82, opened 2021-02-24
fix bugs

yougis:
date as widget filter : datetime fomat from schema is not recognized properly #77
Template is now private propertie of Dashboard #81


lumen/dashboard.py
```diff
@@ -72,7 +72,7 @@ def construct_template(self):
         params = {'title': self.title, 'theme': self.theme}
         if self.logo:
             params['logo'] = self.config.logo
-        return self.template(**params)
```
philippjfr, line 75: Undo this, the `Config` object has a public template property.

PR #88, opened 2021-03-09
Combine sort and facet specs on Target

philippjfr:
Declaring sorting and faceting separately made no real sense since you can only sort over the facets. Therefore you now declare your facet fields and sort fields in a single `facet` spec on the `Target`.

So what was previously this:

```yaml
targets:
  - title: Example
    filters:
      - field: name
        label: Deployment
        type: facet
    sort:
      fields:
        - memory 
        - cpu
        - restarts
        - session_duration
        - render_duration
    reverse: true
```

is now this:

```yaml
targets:
  - title: Example
    facet:
      by:
        - field: name
          label: Deployment
      sort:
        - memory 
        - cpu
        - restarts
        - session_duration
        - render_duration
      reverse: true
```


jbednar:
Cool. Doesn't "reverse" apply indepenently to each sort field, at least in some cases?

PR #101, opened 2021-03-17
Updated diagram for dark themes

Hoxbro:
I could see the diagram.svg in README was not looking good when looking at it with a dark theme. So I cropped some logos and added white background to black logos. I used Inkscape to do the edits.

**Diagram in dark theme (right now):**
image


philippjfr:
Thanks, was planning to do this but didn't get a chance right away.

PR #102, opened 2021-03-17
small copy edits

AlbertDeFusco:


philippjfr:
Thank you, really appreciate it! Will add a getting started section in another PR to avoid clashes.

PR #140, opened 2021-04-20
Add project transform

### LINKED ISSUE #139: Remove GeoViews requirement for the bikes example
The [London Bike Points](https://lumen.holoviz.org/gallery/bikes.html) example currently fails to render if GeoViews is not installed. Compared to the other dependencies, GeoViews is very difficult to install, and indeed when I tried adding GeoViews on Wednesday  after installing a minimal py38 environment for running the other examples, conda failed to solve at all.  I would think we could easily add a transform step calling hv.util.lon_lat_to_easting_northing on the data, at which point it would no longer need GeoViews at all.

philippjfr:
Fixes #139

### DIFF
--- a/examples/bikes/dashboard.yaml
+++ b/examples/bikes/dashboard.yaml
@@ -36,7 +36,6 @@ targets:
         y: lat
         hover_cols: [commonName]
         tiles: Wikipedia
-        geo: true
         responsive: true
         height: 500
         color: bikesCount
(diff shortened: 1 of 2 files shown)


jbednar:
This will be very useful. I hadn't anticipated that you'd put the actual code here in Lumen, since it's already in HoloViews, but I guess you're making sure it's available for non-HoloViews usage, e.g. with raw Bokeh plots?

philippjfr:
>it's available for non-HoloViews usage, e.g. with raw Bokeh plots?

Or any other plotting library, altair, vega, perspective, matplotlib......

jbednar:
Do the other plotting libraries also make tile sources available but require manual projection into web mercator?  I've only noticed that on Bokeh, but if it applies to other libraries too, then sure.

PR #192, opened 2021-09-30
Avoid calling filter.query twice

maximlt:
Just saw that while having a look at the code, it's not fixing anything so feel free to close this PR if not interested in this change!


philippjfr:
Shouldn't be an expensive operation but this also isn't harmful, thanks!

PR #194, opened 2021-10-05
Do not attach control callbacks when rerender is the consequence of a filter event

maximlt:
The _rerender callback is attached to the filter values. Each rerender triggered an _update_views, itself attaching a _rerender callback to the controls values. So changing the values of the filters and the callbacks meant that more and more of _rerender callbacks were added, without never being cleared.

I noticed that the callbacks in _update_views should not be attached if events arrive (i.e. filters being changed), so I just added that condition.

Would feel more comfortable with a review and some tests :)

### DIFF
--- a/lumen/target.py
+++ b/lumen/target.py
@@ -409,8 +409,11 @@ class Target(param.Parameterized):
                     if not view.controls:
                         continue
                     view_controls.append(view.control_panel)
-                    cb = partial(self._rerender, invalidate_cache=False)
-                    view.param.watch(cb, view.controls)
+                    # Attach controls callback only when there are events, i.e. new
+                    # filter values set.
+                    if not events:
+                        cb = partial(self._rerender, invalidate_cache=False)
+                        view.param.watch(cb, view.controls)
             prev_views = views
             if card is None:
                 continue
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,6 @@ builtdocs/
 discover/
 
 node_modules/
+
+# Examples temp data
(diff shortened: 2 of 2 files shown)


lumen/target.py
```diff
@@ -392,29 +392,37 @@ def get_filter_panel(self, skip=None):
...
+                # Only the controls for the first facet is shown so link
+                # the other facets to the controls of the first
+                for v1, v2 in zip(prev_views, views):
+                    v1.param.watch(partial(self._sync_view, v2), v1.controls)
+            else:
+                # Re-render target when controls update
+                for view in views:
+                    view.param.watch(rerender, view.controls)
 
         self._view_controls[:] = view_controls
```
maximlt, line 427:
```suggestion
                    view.param.watch(rerender, view.controls)
                prev_views = views

        self._view_controls[:] = view_controls
```

lumen/target.py
```diff
@@ -392,29 +392,37 @@ def get_filter_panel(self, skip=None):
...
                     if not view.controls:
                         continue
                     view_controls.append(view.control_panel)
-                    cb = partial(self._rerender, invalidate_cache=False)
-                    view.param.watch(cb, view.controls)
-            prev_views = views
-            if card is None:
+
+            # Initialize callbacks
+            if not init:
```
maximlt, line 415:
```suggestion
            if not init:
                prev_views = views
```

maximlt (commented):
`prev_views` was never assigned, @philippjfr is that OK for you?

PR #195, opened 2021-10-05
Expose the disabled property of filters

maximlt:
This exposes `disabled`, as done previously with `visible`. `get_data` subsequently ignores filters when they are disabled.


maximlt:
The goal with adding/improving the tests was to test that when a widget is disabled it is effectively ignored. I didn't manage in this attempt however to setup a test with a filter/source spec that creates what I wanted.

philippjfr:
>Oh it almost went in with a default of True for disabled 😨

😆 I too missed that. I'll spend some time this week adding at least a basic set of unit tests.

PR #204, opened 2021-11-22
Fix the bikes example

maximlt:
The linked selection of bikes stations on the map wasn't selecting any sub-dataset on the histograms and table. What happened exactly is unclear to me (Lumen dashboards are not so easy to debug), it might be related to https://github.com/holoviz/geoviews/issues/497 which show that linked_selections don't work (so well) with geoviews. Or due to the fact that the `project_lnglat` (lat/lon to web mercator) transform in the map view created a selection whose coordinates were not in the same system as the tables in the other views. The issue was fixed by moving the transform to the source declaration, by using a `DerivedSource` to refer to the source to transform.

### DIFF
--- a/examples/bikes/dashboard.yaml
+++ b/examples/bikes/dashboard.yaml
@@ -24,9 +24,16 @@ sources:
         - source: occupancy
           table: occupancy
           index: id
+  station_occupancy_projected:
+    type: derived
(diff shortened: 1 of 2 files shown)


philippjfr (approved):
Looks good!

PR #207, opened 2021-11-24
Add ability to sync dashboard state with URL parameters

philippjfr:
Adds a global `sync_query` config option and options on individual components.

- [x] Add (more) tests
- [x] Add documentation
- [ ] Add example

### DIFF
--- a/docs/dashboard.md
+++ b/docs/dashboard.md
@@ -24,6 +24,7 @@ config:
   title: The title of the overall application
   layout: The layout to put the targets in ('grid', 'tabs', 'column')
   logo: A URL or local path to an image file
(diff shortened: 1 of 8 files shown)


philippjfr:
@maximlt What do you think about the naming of `sync_query`? Any better suggestions?

maximlt:
I think it wouldn't hurt too much to add `url` somewhere in the name, to make it clear what the query is about. People may think it's about data querying for instance. Maybe `sync_url_query`?

philippjfr:
Agreed, was also thinking about `sync_with_url`, which one is better?

maximlt:
`sync_with_url`, definitely!

PR #220, opened 2021-12-10
add apply_to method for base transforms

eli-pinkus:
Allows user to use logic imbedded in a `transform` in an ad-hoc context via the following syntax:
```python
df_transformed = MyWackyTransform.apply_to(df, param1='arg_for_param1', param2='arg_for_param2')
```

### DIFF
--- a/lumen/transforms/base.py
+++ b/lumen/transforms/base.py
@@ -78,11 +78,7 @@ class Transform(param.Parameterized):
         -------
         A DataFrame with the results of the transformation.
         """
-        trans = cls()
-        for key, value in kwargs.items():
-            setattr(trans, key, value)
-            
-        return trans.apply(table)
+        return cls(**kwargs).apply(table)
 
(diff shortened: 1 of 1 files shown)


lumen/transforms/base.py
```diff
@@ -64,6 +64,25 @@ def from_spec(cls, spec):
...
+        
+        Returns
+        -------
+        A DataFrame with the results of the transformation.
+        """
+        trans = cls()
+        for key, value in kwargs.items():
+            setattr(trans, key, value)
+            
+        return trans.apply(table)
```
philippjfr, line 85: This won't work?

```suggestion
        return cls(**kwargs).apply(table)
```

eli-pinkus: yea that's definitely cleaner though, we currently use what's there for our `BXDSTransform` class and it works fine.

PR #222, opened 2022-01-03
Implement SQLTransform

eli-pinkus:
Depends on this fix to `intake`
https://github.com/intake/intake/pull/639


lumen/transforms/sql.py
```diff
@@ -10,6 +10,12 @@ class SQLTransform(Transform):
         
     __abstract = True
     
+    def __str__(self):
+        vals = self.param.get_param_values()
+        vals_dict = {k: v for k, v in vals}
+        vals_dict.pop('name', None)
```
TvieiraB, line 16: Does this pip the name from the original param values?

eli-pinkus: `name` is a field inherited all the way from `param.Parametrized`. It usually looks something like MyTransformType10248 (w/ a random and unique id at the end). This interrupts the query based caching b/c this changing random number alters the hash even if no input parameters change. This `str` pops that `name` off and uses the result of the values for the hash.

philippjfr (approved):
Okay this should be good to go, I folded most of the changes back into the base class and added some tests around caching.

PR #225, opened 2022-01-05
Deactivate the loading spinner after rerender

### LINKED ISSUE #224: Hanging loading spinner on a custom view that doesn't implement `_get_params`
To implement a custom *view* one should implement the `get_panel` method that returns a Panel object. In order to avoid rebuilding the whole Panel object/view on every update/render one can implement the private method `_get_params` that should return a dictionary of Parameter values that are used to update the Panel object when need be (with something like `panel_obj.param.set_param(**params)`). When `_get_params` isn't implemented, the custom view inherits from the base `_get_params` method that simply returns `None` to signal that this way of updating an existing Panel object isn't available.

When `_get_params` isn't implemented a filter change effectively triggers a full update of the view, but also adds a loading spinner (via `Dashboard._set_loading`) on the view that is never deactivated and so prevents any further action on the view.

maximlt:
Fixes #224

There may have been other places where to add this line in the code, let me know what you think @philippjfr

### DIFF
--- a/lumen/target.py
+++ b/lumen/target.py
@@ -490,6 +490,9 @@ class Target(param.Parameterized):
         if self._stale or (update_views and rerender):
             self._application._render()
             self._stale = False
+        # Remove the loading spinner set when a target needs to be
+        # rerendered, for instance when a view has not implemented _get_params.
+        self._application._layout.loading = False
 
     ##################################################################
     # Public API
--- a/lumen/target.py
+++ b/lumen/target.py
@@ -490,9 +490,9 @@ class Target(param.Parameterized):
         if self._stale or (update_views and rerender):
             self._application._render()
             self._stale = False
-        # Remove the loading spinner set when a target needs to be
-        # rerendered, for instance when a view has not implemented _get_params.
-        self._application._layout.loading = False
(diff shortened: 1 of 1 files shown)


lumen/target.py
```diff
@@ -490,6 +490,9 @@ def _rerender(self, *events, invalidate_cache=False, update_views=True):
         if self._stale or (update_views and rerender):
             self._application._render()
             self._stale = False
+        # Remove the loading spinner set when a target needs to be
+        # rerendered, for instance when a view has not implemented _get_params.
+        self._application._layout.loading = False
```
philippjfr, line 495: Only needs to happen if it actually had to re-render right, i.e. does it still work if you indent this line?

maximlt: Yep you're right it still works when this is executed in this `if` block (https://github.com/holoviz/lumen/pull/225/commits/985f2ca4f5136281656ea6a095db0f483c2837c2).

maximlt:
The test suite is now passing after the last commit, but note that this is actually untested code. I wonder if the test suite was failing before that commit because of my code of because of the tests not being 100% correctly setup.

philippjfr (approved):
Thanks for the fix!

PR #227, opened 2022-01-14
Fix broken link to HoloViz Discourse

mattkram:
The current link to the HoloViz Discourse on the Lumen documentation has an extra `https://` in it.


jbednar (approved):
Oops; thanks!

PR #228, opened 2022-01-19
Allow launching UI from CLI

philippjfr:
Adds a `lumen gui` CLI command that launches a Bokeh/Panel server running the builder.

- [ ] Add tests
- [x] Decide on naming of builder, is it `ui`, `gui`, or `builder`?


jbednar:
All those names seem fine, but I'd personally vote for Lumen GUI.

maximlt:
In the list I'd choose `builder`, I still see Lumen as a tool that could be used by people who don't have so much IT culture and for these people probably don't know what a GUI is. I'd like to add one candidate: [wizard](https://en.wikipedia.org/wiki/Wizard_(software))

jbednar:
Sure, though I associate "wizard" specifically with a Windows desktop app.

maximlt:
Because UNIX users have no problem installing stuff ;)

So we're not "normal"! (GUI is an even more remote concept to non-English speakers)

jbednar:
Is UI any clearer for non-English speakers?

maximlt:
I'd say UI and GUI are the same level.

philippjfr:
Okay, stuck with `builder`.

PR #242, opened 2022-02-09
Implement Variable API

philippjfr:
It is frequently required to declare variables that will be reused throughout the dashboard specification. Currently we have support for declaring special variables using yaml templating syntax, e.g. to resolve environment variables we can use the jinja2 syntax `{{ env('some_env_var') }}` and this is resolved when the dashboard is loaded. However frequently we have the need to dynamically resolve variables, tie variables to widgets, or specifically in the context of the `Lumen Builder` to wait to resolve the variables until the final dashboard. To empower users to do these things in a very explicit way we add a `Variable` hierarchy to the existing hierarchy of components (i.e. in addition to `Filter`, `Source`, `Transform` and `View`).

Just like all other components users will be able to declare their own `Variable` types, e.g. to load some secret from a custom variable storage system. Some additional design considerations:

1. The `Variable` definitions are the first thing that are resolved ensuring that all other components can make use of the variables. One `Variable` may even reference another `Variable` in its definition as long as they are declared in the appropriate order.

2. To use a variable anywhere in the yaml specification you use the `$` syntax already used by sources, e.g. to reference a `Variable` named `username` you would specify `$variable.username` (alternative suggestions welcome)

3. A `Variable` can be dynamic just like any other component, therefore it may implement a `panel` attribute or property that returns a Panel component which will be rendered in the sidebar.

4. A `Variable` can declare whether it is to be inlined/resolved or not. This is a declaration to the Lumen Builder whether it should resolve the variable and replace it with a `ConstantVariable` or whether it should simply be inherited as is.


jbednar:
This seems like a solid solution to a lot of the problems we're wrestling with. Can you add an example .yml file showing how to use this, if only as a comment here?  Personally I find `$` to indicate variable substitution more clearly than `@`, which has gotten a connotation of referring to a username, but there may be YAML precedents here that lend weight to `@`.

PR #246, opened 2022-02-22
Use SQLTransforms to compute SQL Source schemas on database

philippjfr:
Currently all sources load the data into memory to compute the schema. For many sources this is okay because it can at least leverage dask to lazily load and compute the unique categories or range of a numerical or date value. Unfortunately for most SQL sources this can be extremely inefficient. So for SQL sources we leverage the `SQLTransform`s recently added to Lumen to let the database itself compute the ranges and unique categories for each column.

The approach can be described as follows:

1. Fetch the schema for the SQL query but with `LIMIT 1`
2. Now find all columns with ranges and all columns with enums and compute these values over the whole table using `MIN/MAX` and `DISTINCT` SQL statements respectively

### DIFF
--- a/lumen/sources/intake.py
+++ b/lumen/sources/intake.py
@@ -31,11 +31,11 @@ class IntakeBaseSource(Source):
         for entry in list(self.cat):
             if table is not None and entry != table:
                 continue
-            data = self.get(entry, __dask=True)
-            if self.load_schema:
-                schemas[entry] = get_dataframe_schema(data)['items']['properties']
-            else:
+            elif not self.load_schema:
                 schemas[entry] = {}
+                continue
+            data = self.get(entry, __dask=True)
+            schemas[entry] = get_dataframe_schema(data)['items']['properties']
         return schemas if table is None else schemas[table]
 
     @cached(with_query=False)
--- a/lumen/sources/intake_sql.py
+++ b/lumen/sources/intake_sql.py
@@ -48,11 +48,11 @@ class IntakeBaseSQLSource(IntakeBaseSource):
 
         schemas = {}
         limit = SQLLimit(limit=1)
-        for table in tables:
+        for entry in tables:
             if not self.load_schema:
                 schemas[entry] = {}
                 continue
-            source = self._get_source(table)
+            source = self._get_source(entry)
             data = self._read(self._apply_transforms(source, [limit]))
             schema = get_dataframe_schema(data)['items']['properties']
             enums, min_maxes = [], []
@@ -72,9 +72,9 @@ class IntakeBaseSQLSource(IntakeBaseSource):
             for col in min_maxes:
                 schema[col]['inclusiveMinimum'] = minmax_data[f'{col}_min'].iloc[0]
                 schema[col]['inclusiveMaximum'] = minmax_data[f'{col}_max'].iloc[0]
-            schemas[table] = schema
+            schemas[entry] = schema
         return schemas if table is None else schemas[table]
-            
+
 
 class IntakeSQLSource(IntakeBaseSQLSource, IntakeSource):
     """
--- a/lumen/sources/intake_sql.py
+++ b/lumen/sources/intake_sql.py
@@ -61,11 +61,14 @@ class IntakeBaseSQLSource(IntakeBaseSource):
                     enums.append(name)
                 elif 'inclusiveMinimum' in col_schema:
                     min_maxes.append(name)
-            distinct_data = self._read(
-                self._apply_transforms(source, [SQLDistinct(columns=enums)])
-            )
             for col in enums:
-                schema[col]['enum'] = distinct_data[col].to_list()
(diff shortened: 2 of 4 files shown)


lumen/sources/intake.py
```diff
@@ -31,11 +31,11 @@ def get_schema(self, table=None):
...
             if table is not None and entry != table:
                 continue
-            data = self.get(entry, __dask=True)
-            if self.load_schema:
-                schemas[entry] = get_dataframe_schema(data)['items']['properties']
-            else:
+            elif not self.load_schema:
                 schemas[entry] = {}
+                continue
+            data = self.get(entry, __dask=True)
```
eli-pinkus, line 37: Net effect here is to move data pull s.t. it's not performed ` if not self.load_schema` ?

lumen/sources/intake_sql.py
```diff
@@ -7,29 +9,72 @@ class IntakeBaseSQLSource(IntakeBaseSource):
...
+        schemas = {}
+        limit = SQLLimit(limit=1)
+        for table in tables:
+            if not self.load_schema:
+                schemas[entry] = {}
+                continue
+            source = self._get_source(table)
+            data = self._read(self._apply_transforms(source, [limit]))
+            schema = get_dataframe_schema(data)['items']['properties']
+            enums, min_maxes = [], []
```
TvieiraB, line 58: Edge cases that could possibly come up
1. What happens with an ID field that is a string? There could be an enum that is millions of distinct digits but in reality we don't want as a drop down or as part of anything on the UI
2. What happens if there is a numeric field we want to be an enum in the source data? (Product 1, 2, 3, etc...) (Could we get around this by casting as a string type before reading from source if the user wants to or specify somewhere here?)
3. What about boolean field values?

TvieiraB: Maybe set a cap of how many unique values an enum can have?

lumen/sources/intake_sql.py
```diff
@@ -7,29 +9,72 @@ class IntakeBaseSQLSource(IntakeBaseSource):
...
+                self._apply_transforms(source, [SQLMinMax(columns=min_maxes)])
+            )
+            for col in min_maxes:
+                schema[col]['inclusiveMinimum'] = minmax_data[f'{col}_min'].iloc[0]
+                schema[col]['inclusiveMaximum'] = minmax_data[f'{col}_max'].iloc[0]
+            schemas[table] = schema
+        return schemas if table is None else schemas[table]
+            
 
 class IntakeSQLSource(IntakeBaseSQLSource, IntakeSource):
```
eli-pinkus, line 79: jw, been meaning to ask, what's the reason for having a separate child here?

lumen/sources/intake_sql.py
philippjfr: > Maybe set a cap of how many unique values an enum can have?

Yes, that seems like the best approach, just append a LIMIT 1000 SQLTransform.

lumen/transforms/sql.py
```diff
@@ -90,3 +90,41 @@ def apply(self, sql_in):
         return Template(template, trim_blocks=True, lstrip_blocks=True).render(
             limit=self.limit, sql_in=sql_in
         )
+
+
+class SQLDistinct(SQLTransform):
```
eli-pinkus, line 95: I like this framework and will be nice to have these as SQLTransforms in general

philippjfr:
I actually don't think this will currently work if there's multiple enums since I'm not sure what DISTINCT will do there.

eli-pinkus (commented):
A couple clarification Q's, otherwise, All good from me with mind to the points that @TvieiraB brought up. Appreciate the quick turn on this one.

lumen/sources/intake.py
philippjfr: Right, creating a dask dataframe is usually cheap but why do it at all when it's not used.

lumen/transforms/sql.py
philippjfr: Yep, I'm very happy with `SQLTransform` thanks for getting the ball rolling on those.

lumen/sources/intake_sql.py
philippjfr: `IntakeSource` defines a constructor which assumes that we are loading a catalog from file. The `IntakeDremioSource` needs to inherit a lot of the sql related stuff but doesn't want to inherit the stuff about loading the catalog from a file so we have two distinct classes. So I had to create two distinct baseclasses.

lumen/sources/intake_sql.py
TvieiraB: Thanks Philipp, on 3 makes sense, on 2 we can append a space or word to the number as a new column and just use that as the enum.  For 1, not sure how LIMIT 1000 would work, but something like that would be great so it doesn't blow up but we can still reference it in later transforms.

PR #258, opened 2022-03-23
minor fixes to sql filters

eli-pinkus:


lumen/transforms/sql.py
```diff
@@ -188,7 +188,8 @@ def apply(self, sql_in):
             elif isinstance(val, list):
                 if not val:
                     continue
-                condition = f'{col} IN {tuple(val)!r}'
+                val = [f'\'{v}\'' for v in val]
+                condition = f'''{col} IN ({', '.join(val)})'''
```
philippjfr, line 192: Can you explain this change to me? The resultant string is identical in all cases I can identify.

philippjfr: Nevermind, I see the issue. This also does not seem quite correct though.

philippjfr, line 192:
```suggestion
                condition = f"{col} IN ({', '.join(map(repr, val))})"
```

PR #277, opened 2022-05-12
Ensure Launcher config passes through specs

philippjfr:
Previously launcher configurations weren't correctly passed through to the instance.


philippjfr:
Merging despite failing tests. Have to restore CI later.

PR #281, opened 2022-05-24
Don't set the widget name if None

maximlt:
Panel widgets `name` don't accept `None`.

### DIFF
--- a/lumen/variables/base.py
+++ b/lumen/variables/base.py
@@ -209,7 +209,7 @@ class Widget(Variable):
         if 'value' not in params and default is not None:
             params['value'] = default
         params['name'] = self.label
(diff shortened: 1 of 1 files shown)


lumen/variables/base.py
```diff
@@ -209,7 +209,7 @@ def __init__(self, **params):
         if 'value' not in params and default is not None:
             params['value'] = default
         params['name'] = self.label
-        deserialized = {}
+            params['name'] = self.label
```
philippjfr, line 212: This doesn't look right.

maximlt: VSCode got crazy for a moment not staging the right changes.

PR #282, opened 2022-06-08
Edge case filter bug

eli-pinkus:
This fixes an issue where an invalid filter is rendered in the edge cases of a condition that has `[None]` as one of the values.

`('column', [None])` would render something like `.......AND column in () OR column IS NULL` when we really want `........AND column IS NULL`


philippjfr:
Thanks @eli-pinkus!

PR #300, opened 2022-08-17
Added suggestion for mistyped/wrong parameters

Hoxbro:
It can be hard to know if something is `transform` or `transforms`. I have tried to take account of this by using `difflib` to come up with suggestions if parameters are mistyped/wrong. 

I have chosen to raise an error when an input parameter does not match the class parameters. My reasoning for this is that the error usually will arrive later, making it harder to debug.  

I have used the following example to see if I can provoke an error by adding or removing an `s`. 

``` python

data_url = 'https://raw.githubusercontent.com/rfordatascience/tidytuesday/master/data/2020/2020-07-28/penguins.csv'

pipeline = Pipeline.from_spec({
    'source': {
        'type': 'file',
        'table': {
            'penguins': data_url
        }
    },
    'filters': [
        {'type': 'widget', 'field': 'species'},
        {'type': 'widget', 'field': 'island'},
        {'type': 'widget', 'field': 'sex'},
        {'type': 'widget', 'field': 'year'}
    ],
    'transforms': [
        {'type': 'aggregate', 'method': 'mean', 'by': ['species', 'sex', 'year']}
    ]
})
```

The example will raise an error (try to find out what raises it).

<details>

  <summary>Error message </summary>

Before:

image

After:

image

</details>

### DIFF
--- a/lumen/base.py
+++ b/lumen/base.py
@@ -4,7 +4,8 @@ import param
 
 from .state import state
 from .util import (
-    match_suggestion_message, resolve_module_reference, validate_parameters,
+    SpecificationError, match_suggestion_message, resolve_module_reference,
+    validate_parameters,
 )
 
 
@@ -44,7 +45,7 @@ class Component(param.Parameterized):
         clsname = cls.__name__
         clslower = clsname.lower()
         if component_type is None:
-            raise ValueError(f"No 'type' was provided during instantiation of '{clsname}' component.")
+            raise SpecificationError(f"No 'type' was provided during instantiation of '{clsname}' component.")
         if '.' in component_type:
             return resolve_module_reference(component_type, cls)
         try:
@@ -63,4 +64,4 @@ class Component(param.Parameterized):
 
         msg = f"No '{clslower}' for {clslower}_type '{component_type}' could be found."
         msg = match_suggestion_message(component_type, cls_types, msg)
-        raise ValueError(msg)
+        raise SpecificationError(msg)
--- a/lumen/base.py
+++ b/lumen/base.py
@@ -1,10 +1,11 @@
-from difflib import get_close_matches
 from functools import partial
 
 import param
 
 from .state import state
-from .util import resolve_module_reference, validate_parameters
+from .util import (
+    match_suggestion_message, resolve_module_reference, validate_parameters,
+)
 
 
 class Component(param.Parameterized):
@@ -61,7 +62,5 @@ class Component(param.Parameterized):
                 return component
 
         msg = f"No '{clslower}' for {clslower}_type '{component_type}' could be found."
-        matches = "', '".join(get_close_matches(component_type, cls_types))
-        if matches:
-            msg += f" Did you mean '{matches}'?"
+        msg = match_suggestion_message(component_type, cls_types, msg)
         raise ValueError(msg)
--- a/lumen/base.py
+++ b/lumen/base.py
@@ -63,5 +63,5 @@ class Component(param.Parameterized):
         msg = f"No '{clslower}' for {clslower}_type '{component_type}' could be found."
         matches = "', '".join(get_close_matches(component_type, cls_types))
         if matches:
-            msg += f" Did you mean: '{matches}'?"
+            msg += f" Did you mean '{matches}'?"
         raise ValueError(msg)
--- a/lumen/base.py
+++ b/lumen/base.py
@@ -44,7 +44,7 @@ class Component(param.Parameterized):
         clsname = cls.__name__
         clslower = clsname.lower()
         if component_type is None:
-            raise ValueError(f"'type' for '{clslower}' is not available.")
+            raise ValueError(f"No 'type' was provided during instantiation of '{clsname}' component.")
         if '.' in component_type:
             return resolve_module_reference(component_type, cls)
         try:
--- a/lumen/base.py
+++ b/lumen/base.py
@@ -1,9 +1,10 @@
+from difflib import get_close_matches
 from functools import partial
 
 import param
 
 from .state import state
-from .util import resolve_module_reference
+from .util import resolve_module_reference, validate_parameters
 
 
 class Component(param.Parameterized):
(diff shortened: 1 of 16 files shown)


lumen/base.py
```diff
@@ -39,13 +42,26 @@ def refs(self):
...
+            cls_types.add(cls_type)
+            if cls_type == component_type:
                 return component
-        raise ValueError(f"No {clsname} for {clslower}_type '{component_type}' could be found.")
+
+        msg = f"No '{clslower}' for {clslower}_type '{component_type}' could be found."
+        matches = "', '".join(get_close_matches(component_type, cls_types))
+        if matches:
+            msg += f" Did you mean: '{matches}'?"
+        raise ValueError(msg)
```
jbednar, line 67: Should this code be in Param instead? It doesn't seem specific to Lumen.

Hoxbro: The `validate_parameters` could/should be implemented in Param as an option instead of "only" emitting a warning. https://github.com/holoviz/param/issues/174 

For the code in this function, in particular, I don't think it can be ported to param.

jbednar: You don't think that Param could raise a helpful exception when given an attribute name that is not a parameter, guessing the right parameter name the same as done here? I must be missing something, then.

Hoxbro: Yes! A helpful exception for Param would be a very nice addition. 

But that is not what this function does, which tries to match the input string `component_type` to a lumen component and return it.

lumen/base.py
```diff
@@ -16,6 +18,8 @@ class Component(param.Parameterized):
 
     def __init__(self, **params):
         self._refs = params.pop('refs', {})
+        expected = list(self.param.params())
+        validate_parameters(params, expected, type(self).name)
```
philippjfr, line 22: Reasonable at this level, note though that a lot of components simply collect any parameters that haven't been explicitly declared into `self.kwargs` so this will never be triggered in those cases.

lumen/base.py
```diff
@@ -39,13 +43,24 @@ def refs(self):
     def _get_type(cls, component_type):
         clsname = cls.__name__
         clslower = clsname.lower()
+        if component_type is None:
+            raise ValueError(f"'type' for '{clslower}' is not available.")
```
philippjfr, line 47: Not sure this message is very clear:

```suggestion
            raise ValueError(f"No 'type' was provided during instantiation of '{clsname}' component.")
```

philippjfr: Might use "materialization" over "instantiation" but the difference may be too technical. Materialization refers to the process of instantiating the object from a spec.

lumen/target.py
```diff
@@ -84,6 +84,10 @@ def _construct_layout(self):
     def from_spec(cls, spec, filters=None, pipelines={}):
         spec = dict(spec)
         view_specs = spec.get('views', [])
+        if view_specs is None:
+            # If views is not defined in yaml file
+            raise ValueError("No 'views' was provided during instantiation.")
```
philippjfr, line 89: This branch will never be declared because the get returns an empty list:

```python
view_specs = spec.get('views', [])
```

Hoxbro: Got it with this yaml file
``` yaml
config:
  title: Lumen

sources:
  penguins:
    type: file
    tables:
      penguins: https://raw.githubusercontent.com/rfordatascience/tidytuesday/master/data/2020/2020-07-28/penguins.csv

targets:
  - title: Table
    source: penguins
    view:  # <- Missing an s
      table:
        type: table
        table: penguins
```

Hoxbro:
I have found another `s`-problem with this yaml file
``` yaml
config:
  title: Lumen

sources:
  penguins:
    type: file
    tables:
      penguins: https://raw.githubusercontent.com/rfordatascience/tidytuesday/master/data/2020/2020-07-28/penguins.csv

targets:
  - title: Table
    source: penguins
    views:
      table:
        type: table
        table: penguins
        transform:  # missing an s here
          - type: iloc
            end: 100
```
Which does not raise an error, but the following warning:

``` python-traceback
WARNING:param.Tabulator00842: Setting non-parameter attribute transform=[{'type': 'iloc', 'end': 100}] using a mechanism intended only for parameters
```

lumen/target.py
philippjfr: Oh right, this is on `Card` not on `target`. I'll move the validation somewhere earlier than this. The goal is also to add very specific error messages that can help users find where in the YAML specification the error is. So the goal here would be to have some like this be the output:

```
Ensure the following target declares a `views` field:

    - title: Table
       source: penguins
       view: # <- Missing an s
         table:
           type: table
           table: penguins
```

Mapping it back to the YAML is obviously quite difficult but would be nice. In the absence of that it should at least mention the title for instance.

lumen/target.py
```diff
@@ -344,6 +340,8 @@ def _construct_cards(self):
         controls, all_transforms = [], []
         linked_views = None
         for facet_filters in self.facet.filters:
+            if self.views is None:
+                raise ValueError(f"Ensure that the target '{self.title}' declares a 'views' field.")
```
philippjfr, line 344: Let's just validate that a target has to declare views in `Target.from_spec` and simple skip `_construct_cards` entirely if no views are declared.

Hoxbro: I think [a9106fb](https://github.com/holoviz/lumen/pull/300/commits/a9106fb2227e232772c0864e0ff23e506ebf75d7) does what you are asking for.

Hoxbro:
I think it could make sense to have a custom error like SpecificationError. 

Maybe when this error is raised, it would only show the message and not the traceback.

lumen/target.py
```diff
@@ -522,6 +521,9 @@ def from_spec(cls, spec, **kwargs):
         # Resolve source
         spec = dict(spec)
         views = spec.get('views', [])
+        if views is None:
+            raise ValueError(f"Ensure that the target '{spec['title']}' declares a 'views' field.")
```
philippjfr, line 525:
```suggestion
        if 'views' not in spec:
            raise ValueError(f"Ensure that the target '{spec['title']}' declares a 'views' field.")
        views = spec['views']
```

philippjfr:
>I think it could make sense to have a custom error like SpecificationError.

Fully agree!

philippjfr:
Superseded by #312 (kept all your commits)

PR #302, opened 2022-08-22
Refator Source tests

thuydotm:
Use `pytest.fixture` and `pytest.parameterize` to reorganize tests into groups where tests in a group do similar stuffs.


philippjfr:
Thanks @thuydotm!

PR #308, opened 2022-08-25
Check for maximum value for layout in views

Hoxbro:
Created with 

``` yaml
sources:
  penguins:
    type: file
    tables:
      penguins: https://raw.githubusercontent.com/rfordatascience/tidytuesday/master/data/2020/2020-07-28/penguins.csv

targets:
  - title: Penguins
    source: penguins
    layout: [[0], [1]]
    views:
      - type: hvplot
        tables: penguins
        kind: scatter
```

And gave an IndexError beforehand. 

TODO: 
- [x] Make unit test

### DIFF
--- a/lumen/target.py
+++ b/lumen/target.py
@@ -4,6 +4,7 @@ from functools import partial
 from io import BytesIO, StringIO
 from itertools import product
 
+import numpy as np
 import panel as pn
 import param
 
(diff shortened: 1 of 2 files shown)


lumen/target.py
```diff
@@ -51,6 +52,10 @@ def __panel__(self):
     def _construct_layout(self):
         layout = self.layout
         if isinstance(layout, list):
+            if np.concatenate(layout).max() >= len(self.views):
```
philippjfr, line 55: The indexes may also be strings matching the view names.

lumen/target.py
```diff
@@ -52,16 +51,19 @@ def __panel__(self):
...
                 for index in row_spec:
                     if isinstance(index, int):
-                        view = self.views[index]
+                        if index < view_size:
+                            view = self.views[index]
+                        else:
+                            raise ValueError(
+                                f"Layout for '{self.title}' can only "
+                                f"have values up to {view_size - 1}."
+                            )
```
philippjfr, line 66: Being nit-picky about style here but I generally like writing `if <condition> then <error>` very explicitly for readability:

```suggestion
                        if index >= view_size:
                            raise ValueError(
                                f"Layout for '{self.title}' can only "
                                f"have values up to {view_size - 1}."
                            )
                        view = self.views[index]
```

PR #322, opened 2022-09-13
Notify user of runtime errors

### LINKED ISSUE #321: Notify user of runtime errors
Lumen can fail to update for a number of reasons and currently can fail silently when you interact with the application. We want to make sure that users see helpful error messages should something fail. The places where this is most important include

- `Pipeline._update_data`: This is where most of the updates happen, it fetches data from the source, filters data and applies transforms all of which can fail.
- `Download._table_data` can fail to export downloads.
- ...

A good way of finding all the places where this may be needed might be to grep for `.param.watch` and provide nice error handling for all callbacks.

Hoxbro:
Fixes #321 

The main thing in this PR is the `catch_and_notify` decorator to be able to catch runtime errors, and then both notify the user with a notification and log the exception. 

I'm unsure if I should re-raise the error after this, as the following code could fail because it expects the decorated function to succeed.  

A (staged) example of how this looks.

https://user-images.githubusercontent.com/19758978/189873191-fb7d6a73-a11f-4558-9ff6-377fb82a9e6d.mp4

TODO:
- [x] Check if needed and add a decorator on watched functions.

---
Also, fixing these which are non-related to the original issue:
- The title of the weather example
- Adding +1 to `Loading target ...` so it is one-indexed. This makes sense to me that the last message is `2/2` and not `1/2.

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -7,9 +7,16 @@ import re
 import subprocess
 import sys
 
+from functools import wraps
+from logging import getLogger
+
+import panel as pn
+
 from jinja2 import DebugUndefined, Environment, Undefined
 from pandas.core.dtypes.dtypes import CategoricalDtype
 from panel import state
+
+log = getLogger(__name__)
 
 
 def get_dataframe_schema(df, columns=None):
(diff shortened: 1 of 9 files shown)


lumen/util.py
```diff
@@ -236,3 +243,49 @@ def extract_refs(spec, ref_type=None):
         return refs
     filtered = [ref for ref in refs if ref[1:].startswith(ref_type)]
     return filtered
+
+
+def catch_and_notify(message=None):
```
philippjfr, line 248: Great, a decorator is exactly what I was thinking about.

lumen/util.py
```diff
@@ -236,3 +243,49 @@ def extract_refs(spec, ref_type=None):
...
+
+    Parameters
+    ----------
+    message : str | None
+        The notification message, by default None.
+        None will give this "Error: {e}" where e is the
+        exception message.
+
+    """
+    pn.extension(notifications=True)
```
philippjfr, line 264: Can you see if this is sufficient?

```suggestion
    pn.config.notifications = True
```

lumen/util.py
Hoxbro: Glad to hear!

Hoxbro:
This should be ready for review. 

If the error needs to be reraised after the notification, I want to rename the function, maybe to `notify` or `notify_error`.

PR #323, opened 2022-09-20
Getting started section

Hoxbro:
This should be ready for review:

image

image

image

image

For the octicon I used https://primer.style/octicons/


docs/getting_started/build_an_app.md
```diff
@@ -0,0 +1,378 @@
...
+
+![](../_static/getting_started/build_app_07.png)
+
+
+
+
+## Building the app
+This section will walk through the steps to creating the app.
+Create a file named `penguins.yaml`.
+The first thing which is needed is a source. Here we will add a penguin source from a file that contains a penguin's table.
```
philippjfr, line 12:
```suggestion
The first thing which is needed is a source. Here we will add file source that will fetch data about individual penguins of three different and various measurement about them (such as the length of their bill).
```

docs/getting_started/build_an_app.md
```diff
@@ -0,0 +1,378 @@
...
+
+:::{tab-item} Preview
+:sync: preview
+
+![](../_static/getting_started/build_app_00.png)
+:::
+::::
+
+This returns an empty dashboard because we don't have a targeted view.
+The simplest view to add is a table with the penguin's table in it.
```
philippjfr, line 39: Can you keep paragraph together rather than starting each sentence on a new line.

docs/getting_started/build_an_app.md
```diff
@@ -0,0 +1,378 @@
...
+:::{tab-item} Preview
+:sync: preview
+
+![](../_static/getting_started/build_app_02.png)
+:::
+::::
+
+
+This plot is a bit overwhelming, so instead of plotting everything in one plot, we can plot `bill_length_mm` on the x-axis and `bill_depth_mm` on the y-axis.
+Furthermore, we can color based on the specie and change the kind of the plot to scatter.
```
philippjfr, line 105:
```suggestion
Furthermore, we can color based on the species and change the kind of the plot to scatter.
```

docs/getting_started/build_an_app.md
```diff
@@ -0,0 +1,378 @@
...
+
+This returns an empty dashboard because we don't have a targeted view.
+The simplest view to add is a table with the penguin's table in it.
+
+
+::::{tab-set}
+:::{tab-item} YAML
+:sync: yaml
+``` {code-block} yaml
+:emphasize-lines: 7-12
```
philippjfr, line 46: The emphasis use is great!

docs/getting_started/build_an_app.md
```diff
@@ -0,0 +1,378 @@
...
+
+:::{tab-item} Preview
+:sync: preview
+
+![](../_static/getting_started/build_app_05.png)
+:::
+::::
+
+
+The layout of the views is in a column where we can't even see the last plot, and the dashboard is non-responsive.
```
philippjfr, line 239:
```suggestion
The default layout we get is less than ideal for this case since it cuts off one of our plots, leaves a lot of empty space and does not resize responsively.
```

philippjfr:
Left a few comments, overall this is looking good though!

docs/getting_started/build_an_app.md
Hoxbro: I like having sentences separated by a new line; the output will be the same when converted to HTML. 

I think it has some advantages (even though they are small):
- For version control, it is easier to see what sentences have changed.  
- When staging changes on a large paragraph with multiple changes across sentences, I have a tendency to either accept or discard all the changes, new lines give me more fine-grained control. 
- When using spell checking, I can easily see what sentence has a spelling error based on the line.

docs/getting_started/build_an_app.md
Hoxbro: I will commit your suggestion, just want to note that the way I built up the sentences was to try to match the words in the sentence with what is happening in the YAML specification.

docs/getting_started/build_an_app.md
jlstevens: Those are all good points @Hoxbro though I see a few downsides too:

1. This style hurts the fluidity of readability when looking at the Markdown file.
2. Is this something we would try to enforce throughout or is this something you only want to do for yourself when working on the docs?

As I think this would be hard to enforce throughout, it would be fine with me if @Hoxbro uses this approach for himself as long as there is no expectation that other people will follow this style (and people could rewrap the text if they want). The downside with *that* is inconsistency...

My only other suggestion might be that this style of separating sentences might be a good idea while working on a docs PR but we should rewrap the text in a consistent way before merge. How does that sound?

philippjfr: Right readability and consistency are my main concerns too. I agree with all your points @Hoxbro but if we decide to do this then I think we should do it consistently and it just hurts my eyes a little too much to agree to that.

philippjfr: Let's get @droumis' take here as well.

jlstevens: So for consistency, should we agree on a rewrap width before merging docs PRs? Before the merge, I'm ok with any style...

Hoxbro: I don't think it needs to be as black and white as both of you make it out to be. I think having different "formats" across the markdown files is fine, as long as the final product (the HTML) is consistent.  

For readability, I don't think it is as big of an eyesore, it is different but you get used to it quickly.

philippjfr: Leaving it for now but still leaning towards wrapping the paragraphs.

PR #327, opened 2022-09-26
Add reference guide skeleton

philippjfr:
Adds structure for reference guide with empty files.


droumis:
can this be merged now @philippjfr?

PR #329, opened 2022-09-26
Adding how to on caching 

Hoxbro:
Adding a section on how-to on cache.

Found some bugs when setting up the example in the how-to.
1.  `dask.dataframe` was never imported and therefore was not in `sys.modules`. I now try to import `dask.dataframe` and use that through the code.
2. `FileSource` was set up to not save on parquet files, but this should only be the case if the file is local and not remote like the example I use. 
image

### DIFF
--- a/docs/how_to/data_intake/cache.md
+++ b/docs/how_to/data_intake/cache.md
@@ -2,21 +2,22 @@
 
 :::{admonition} What does this guide solve?
 :class: important
-This guide will go through how to cache large files for easier access.
+This guide will show you how to locally cache data to speed up reloading from a remote source.
 :::
 
 ## Caching file
-When working with large files, it can be very advantageous to cache it and therefore be able to save time when reloading the data.
-To cache a [source](../../reference/source) is done by using `cache_dir` followed by a directory.
+When working with large data in a non-optional file format, creating a local cache can be advantageous to save time when reloading the data.
+Caching a [source](../../reference/source) is done by using `cache_dir` followed by a directory.
 If the directory does not exist, it will be created.
-Below is an example of caching a 370 MB file, and the table consists of almost 12 million rows.
+The data will be saved as [parquet](https://www.databricks.com/glossary/what-is-parquet) files in the cache directory.
+Below is an example of caching a 370 MB file that consists of almost 12 million rows.
 
 :::{warning}
 The initial load can take a couple of minutes as the file needs to be downloaded and cached first.
 :::
 
 ::::{tab-set}
-:::{tab-item} Specification (YAML)
+:::{tab-item} YAML
 :sync: yaml
 ``` {code-block} yaml
 :emphasize-lines: 4
@@ -36,11 +37,9 @@ targets:
       - type: table
         table: large_table
 ```
-_The non-emphasized lines are only there to see the table when running `lumen serve file.yaml`._
-
 :::
 
-:::{tab-item} Pipeline (Python)
+:::{tab-item} Python
 :sync: python
 ``` {code-block} python
 :emphasize-lines: 8
@@ -59,11 +58,10 @@ pipeline = Pipeline.from_spec(
 )
 pipeline.data
 ```
-_The non-emphasized lines are only there to have a table to cache when running the code_
 :::
 ::::
 
 
 :::{note}
-Lumen's [cache]() can be added to all sources found [here](../../reference/source).
+Lumen's [cache]() can be added to all [source types](../../reference/source).
 :::
--- a/docs/how_to/data_intake/cache.md
+++ b/docs/how_to/data_intake/cache.md
@@ -1,0 +1,69 @@
+# How to use cache
+
+:::{admonition} What does this guide solve?
+:class: important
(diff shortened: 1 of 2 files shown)


docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
+# How to use cache
+
+:::{admonition} What does this guide solve?
+:class: important
+This guide will go through how to cache large files for easier access.
+:::
+
+## Caching file
+When working with large files, it can be very advantageous to cache it and therefore be able to save time when reloading the data.
```
droumis, line 9:
```suggestion
When working with large remote data, it can be advantageous to create a local cache to save time when reloading the data.
```

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+To cache a [source](../../reference/source) is done by using `cache_dir` followed by a directory.
```
droumis, line 10:
```suggestion
Caching a [source](../../reference/source) is done by using `cache_dir` followed by a directory.
```

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+If the directory does not exist, it will be created.
+Below is an example of caching a 370 MB file, and the table consists of almost 12 million rows.
```
droumis, line 12:
```suggestion
Below is an example of caching a 370 MB file that consists of almost 12 million rows.
```

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+
+:::{warning}
+The initial load can take a couple of minutes as the file needs to be downloaded and cached first.
+:::
+
+::::{tab-set}
+:::{tab-item} Specification (YAML)
```
droumis, line 19: I don't really like the names `Specification (YAML)` and `Pipeline (Python)`.. there is a concept of specification in Python, and a pipeline in the YAML interface.. so it gets confusing..

I would just call it `YAML` and `Python`

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+      engine: fastparquet
+
+targets:
+  - title: Table
+    source: large_source
+    views:
+      - type: table
+        table: large_table
+```
+_The non-emphasized lines are only there to see the table when running `lumen serve file.yaml`._
```
droumis, line 39: I don't think this is needed.. it's implied by the existence of the emphasis.

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+
+:::
+
+:::{tab-item} Pipeline (Python)
```
droumis, line 43:
```suggestion
:::{tab-item} Python
```

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+            "type": "file",
+            "cache_dir": "cache",
+            "tables": {"large_table": data_url},
+            "kwargs": {"engine": "fastparquet"},
+        },
+    }
+)
+pipeline.data
+```
+_The non-emphasized lines are only there to have a table to cache when running the code_
```
droumis, line 62: not needed

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
...
+:::
+::::
+
+
+:::{note}
+Lumen's [cache]() can be added to all sources found [here](../../reference/source).
```
droumis, line 68:
```suggestion
Lumen's [cache]() can be added to all [source types](../../reference/source).
```

docs/how_to/data_intake/cache.md
```diff
@@ -0,0 +1,69 @@
+# How to use cache
+
+:::{admonition} What does this guide solve?
+:class: important
+This guide will go through how to cache large files for easier access.
```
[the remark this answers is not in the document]
Hoxbro: An important note is cache can also be used on local files in non-parquet file formats. This should maybe be expressed more explicitly. 
The reason this example uses remote files is so the user can run it on his own computer.

docs/how_to/data_intake/cache.md
Hoxbro:
```suggestion
:::{tab-item} YAML
```

docs/how_to/data_intake/cache.md
Hoxbro: Much of the YAML specification is boilerplate text for actually showing something on the dashboard. I like to be very explicit about that so people are not overwhelmed by it.

lumen/sources/base.py
```diff
@@ -27,6 +26,11 @@
 from ..util import get_dataframe_schema, is_ref, merge_schemas
 from ..validation import ValidationError, match_suggestion_message
 
+try:
+    import dask.dataframe as dd
```
philippjfr, line 30: Leaving this for now but may want to delay the import.

PR #330, opened 2022-09-27
Add functionality to manually apply updates

philippjfr:
Adds a `manual_update` option to the global config, each Target and Pipeline to be able to manually update the data using a button click. This is useful when applying a filter or transform is very slow.

apply_update

- [x] Add tests
- [x] Decide on naming of `manual_update` parameter

### DIFF
--- a/docs/how_to/pipeline_python.md
+++ b/docs/how_to/pipeline_python.md
@@ -147,5 +147,8 @@ pn.Row(
 ```
 ![dashboard preview](../_static/pipeline_dash.png)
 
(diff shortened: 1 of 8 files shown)


droumis:
When needing this manual update button, the user is trying to prevent auto-updates, so I could see the parameter being something like `auto_update: False`

philippjfr:
Think I prefer that over `manual_update`, thanks.

maximlt:
Just adding some comments as I recently worked on a similar thing on a Panel app. The app was only doing filtering so I called it `live_filtering`. The UI I thought implementing (I didn't have to in the end) was a combination of a check box and a button, to either enabling live filtering (the button would then be disabled) or disabling it in which case you have to click on the button to execute the filters and update the view.

PR #334, opened 2022-09-27
Add data outtake for downloading data

Hoxbro:
image


droumis:
looks good! I might suggest giving the different methods specific names so we can easily talk about them in other places... something like: sidebar download, target download, view download. I don't love these names but you get the idea. 

The only other thing might be to put the admonition-note above the tab-set, so that no one misses it if they are viewing on a smaller screen.

Hoxbro:
> sidebar download, target download, view download. I don't love these names but you get the idea.
I could not find better names, so I used yours. 

> The only other thing might be to put the admonition-note above the tab-set
I thought about this and would have done it if it was a warning, but because it is "only" a note I chose to put it below so it does not take focus from the interesting part of this how-to.

PR #335, opened 2022-09-27
Improve the logic behind importing a module

Hoxbro:
When a module was not installed, it would raise a very ambiguous error message: 

``` python
Source component specification declared unknown type 'intake'.
```

This is because the `__import__` did not look if the import error was caused by the actually import or one of a dependency inside the file imported.  

When failing, the error message could be improved, but I want some feedback on if this is the way forward.


lumen/base.py
```diff
@@ -363,10 +363,19 @@ def _get_type(cls, component_type, spec=None):
...
-            pass
+            import_name = f'lumen.{base_type.__name__.lower()}s.{component_type}'
+            __import__(import_name)
+        except ImportError as e:
+            if e.name != import_name:
+                msg = (
+                    f"In order to use the {base_type.__name__.lower()} "
+                    f"component '{component_type}', the '{e.name}' package "
+                    "must be installed. It can be installed with:"
+                    f"\n  conda install {e.name}"
```
maximlt, line 376: It's a bit weird to just recommend `conda install`. I'd either not recommend anything, or recommend both `pip install` and `conda install`.

Hoxbro: The inspiration was taken from the error message in hvplot, but I can remove the recommendation.

https://github.com/holoviz/hvplot/blob/master/hvplot/converter.py#L1246-L1249

maximlt: Ah yes but that's specifically for datashader in hvPlot. While this error is for any package If I'm not mistaken.

Issue #339, opened 2022-09-29
Pre-caching proposal


philippjfr:
Pre-caching data is simple when the data is a manageable size and we can perform a single query to get all of it. However when the data is potentially huge and various filters and transforms are applied we need to find a different solution to caching individual queries. Since we now express all filters and transform at the level of a `Pipeline` this is also the appropriate place to declare the pre-caching configuration. There are two possibilities for expressing the space of queries to cache:

1) Express the values for each filter and potentially also options for specific transforms. We could then compute the cross-product of all of those options and execute and cache those queries. The problem with this approach is that cross-products quickly get huge and often we just need to pre-fetch the most commonly viewed combinations.
2) Express each explicit combination of filter, transform and variable values to fetch.

Both approaches have value but for now we will focus on the latter.

## Specification

The proposed specification for defining individual pre-cached values looks like this: 

```yaml
pipelines:
  sql_table:
    - source: sql_source
      table: sql_table
      filters:
        a:
          type: widget
        b:
          type: widget
      sql_transforms:
        - type: aggregate
          by: $variables.groupby
          aggregates:
            c: mean
      precache:
        - a: 1
          b: 2
          variables:
            groupby: day
        - a: 2
          b: 1
          variables:
            groupby: week
        ...
```

To address option 1 we will eventually also support this syntax:

```yaml
pipelines:
  sql_table:
    - source: sql_source
      table: sql_table
      filters:
        a:
          type: widget
        b:
          type: widget
      sql_transforms:
        - type: aggregate
          by: $variables.groupby
          aggregates:
            c: mean
      precache:
        a: [1, 2]
        b: [1, 2]
        variables:
            groupby: [day, week]
```

philippjfr:
This has been merged.

PR #340, opened 2022-09-29
Add precache command

philippjfr:
Implements https://github.com/holoviz/lumen/issues/339

- [ ] Add tests
- [ ] Add docs

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -15,9 +15,12 @@ import panel as pn
 from jinja2 import DebugUndefined, Environment, Undefined
 from pandas.core.dtypes.dtypes import CategoricalDtype
 from panel import state
+from panel.io.cache import _hash_funcs
 
 log = getLogger(__name__)
 
+# Temporarily patch panel _hash_funcs
+_hash_funcs[dt.date] = lambda obj: f'{type(obj)}{obj}'.encode('utf-8')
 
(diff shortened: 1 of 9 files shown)


lumen/util.py
```diff
@@ -15,9 +15,12 @@
 from jinja2 import DebugUndefined, Environment, Undefined
 from pandas.core.dtypes.dtypes import CategoricalDtype
 from panel import state
+from panel.io.cache import _hash_funcs
 
 log = getLogger(__name__)
 
+# Temporarily patch panel _hash_funcs
+_hash_funcs[dt.date] = lambda obj: f'{type(obj)}{obj}'.encode('utf-8')
```
Hoxbro, line 23: Why not just pin panel? Is it because `0.14.1` is needed?

philippjfr: Yes, will release 0.14.1 tomorrow or Monday so this will be removed before the Lumen release but we need this for a demo today.

philippjfr: Demo driven development is of course not ideal but don't have much choice in this case.

Hoxbro: I understand. 

I would properly open up a PR with the removal and a pin, so it is not forgotten before the next release. But maybe that is just me being too cautious.

PR #342, opened 2022-09-30
Example for how to use variables

Hoxbro:
It started with an example of how to use variables but failed with a verification error. This is fixed with 488ee9c5d935788602d1de50f4f8f93032040ee5. 

Though, I still get this warning `WARNING:param.Source: Resolving nested variable references currently not supported.` which I think is why I get no automatic update when changing the variable input. 

Maybe I'm doing something wrong with the specification. 

``` yaml
variables:
  ticker:
    type: widget
    kind: TextInput
    default: https://raw.githubusercontent.com/matplotlib/sample_data/master/aapl.csv

sources:
  stock_data:
    type: file
    tables:
      ticker: $variables.ticker

targets:
- title: Table
  source: stock_data
  views:
    - type: table
      table: ticker
```


Hoxbro:
Noticed that I could get a very bad snippet in validation error, this is fixed with [2d21689](https://github.com/holoviz/lumen/pull/342/commits/2d21689e46a8bf019ec8ee8cb8a28e4298b2abc6)
image

With YAML specifcation
``` YAML
variables:
  ticker:
    type: widget
    kind: TextInput
    default: https://raw.githubusercontent.com/matplotlib/sample_data/master/aapl.csv

sources:
  stock_data:
    type: file
    tables:
      ticker: $variables.ticker

targets:
- title: Table
  source: stock_data
  views:
    - type: table
      table:  $variables.ticker
```

Hoxbro:
Without adding `shell=True` in `check_output` I got a `FileNotFoundError`, when passing in `echo Hello`.

The last thing missing for this to be ready (for review) is an example like the one used in the first post, but without having a `Resolving nested variable references currently not supported` warning.

docs/how_to/variables_and_references.md
```diff
@@ -7,6 +7,157 @@ This guide shows you how to use one of the advanced features of Lumen: Variables
 
 ## Overview
 Variables are one of the most powerful features of Lumen.
-This guide will give you an overview of the different variables and how to reference them in the YAML specification.
+This guide will give you an overview of three different types of variables and how to reference them in the YAML specification.
+
+As a rule of thumb, internal variables are referenced with a starting dollar sign `$`, whereas external references use double curly brackets before and after `{{ }}`.
 
 ## Variables
```
philippjfr, line 14: I see references explained here and template variables but not a single example of an actual variable definition.

philippjfr: Note that's not a great example. I'd probably opt for some widget variable that controls the parameter of a transform.

Hoxbro: This part is a work-in-progress as I'm missing a working example. It is hard to document it when I'm unsure if it works.

Hoxbro: I'm aware that this is missing. Did you read my previous comment? 

I'm not able to get the interactive part of the Variables as `Resolving nested variable references currently not supported` so they are just static.

philippjfr: Still looking into that but I don't think that blocks you from building an example where a variable controls a transform parameter for instance.

Hoxbro:
@philippjfr This should now be ready for review. 

Screenshot 2022-10-11 at 17-00-37 How to use variables — Lumen 0 5 0a66 documentation

PR #343, opened 2022-09-30
Create how to on custom local components

droumis:
- [x] import a component by its module path, e.g. you can do my_library.some_module.MyTransform

- Also made the install page more explicit about python installation after a good talk with @Hoxbro. 
- Also included some minor fixes like syntax highlighting and a couple of links

127 0 0 1_5500_docs__build_html_how_to_local_components html (1)


maximlt:
Maybe this is still WIP but shouldn't this How-to actually describe how to write the code of a custom compoment? I.e. you have to create a class that inherits from a base class, you have to define transform_type to define the name by which you will be referring to that component in the spec, you have to implement the apply method with this signature and returning that type of object, create Parameters, specify _field_params, etc.

droumis:
@maximlt, according to the plan by @philippjfr , there are separate how-to guides planned for 

- Build a Custom Source 
- Build a Custom Filter
- Build a Custom Transform
- Build a Custom View

This PR is much simpler and is intended just to cover how to use those components, once created.

philippjfr:
Yeah, let's keep that separate. However one thing that wasn't documented elsewhere and should be here is that you can import a component by its module path, e.g. you can do `my_library.some_module.MyTransform`>

droumis:
@philippjfr can we merge this now?

PR #376, opened 2022-11-08
Ensure schema is still generated for empty dataframe

philippjfr:
Previously we were not using the dtype information to at least return a basic schema.

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -10,6 +10,7 @@ from functools import wraps
 from logging import getLogger
 from subprocess import check_output
 
+import pandas as pd
 import panel as pn
 
 from jinja2 import DebugUndefined, Environment, Undefined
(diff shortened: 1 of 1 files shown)


lumen/util.py
```diff
@@ -277,10 +286,13 @@ def wrapper(*args, **kwargs):
             try:
                 return func(*args, **kwargs)
             except Exception as e:
-                log.error(
-                    f"{func.__qualname__!r} raised {type(e).__name__}: {e}"
-                )
-                pn.state.notifications.error(message.format(e=e))
+                if pn.state.notifications:
+                    log.error(
```
Hoxbro, line 290: Why does `log.error` depends on `pn.state.notification` here?

philippjfr: I prefer raising an actual error if we can't notify the user.

philippjfr: This would only really occur in a notebook/programmatic context where the user would actually see the error.

Hoxbro: Oh, read the else as part of the try/except and not the if block.

PR #379, opened 2022-11-09
Fix missing navigation bar icons 

Hoxbro:
On the [current dev docs](https://pyviz-dev.github.io/lumen/) the navigation bar icons for Github and Discourse is missing:

Current
image

With this PR
image

I also updated the Discourse URL to refer to the Lumen category on Discourse, as I could see this is what is Panel does.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -48,14 +48,6 @@
     }
 }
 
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks @Hoxbro!

PR #389, opened 2022-11-25
Change state to config in catch_and_notifify

Hoxbro:
Mostly for debugging purposes when working in a Notebook.  

``` python
import panel as pn
pn.extension(notifications=True)

pn.config.notifications=False

print(pn.state.notifications)

```

Will output `NotificationArea()` and therefore not raise an error.


philippjfr:
Makes sense!

PR #391, opened 2022-11-25
Initialize value for a Widget even if it is throttled

Hoxbro:
Noticed that when initializing a Widget variable with `value_throttled` parameter, it would not set the value of the actual widget.


philippjfr:
Looks good, thanks!

PR #395, opened 2022-12-05
fix renamed import of target to layout

droumis:
fixes doc CI build

simply changes target to layout in the generate_reference script


philippjfr:
Sweet, thanks!

PR #422, opened 2023-01-02
Move duckdb import into __init__

Hoxbro:
Duckdb is not a required dependency, so moving the import into `__init__`.

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -1,4 +1,3 @@
-import duckdb
 import param
 
 from ..transforms import Filter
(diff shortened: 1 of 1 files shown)


philippjfr:
Don't see the need for this, `lumen.sources.duckdb` should only be imported if you import it explicitly or if you define a source with:

```
type: duckdb
```

Hoxbro:
I noticed my test discovery failed and wrongly assumed it was because of the first commit.

I have added a new PR, which checks for it in the test suite instead.

philippjfr:
Sweet, thanks!

PR #431, opened 2023-01-17
add card nav for main how to index

droumis:
Attempting to align Lumen How-to format with recent work in Panel docs. 

In addition to adding cards for main how-to index page (and octicons!), I've made each index for the how-to categories redirect to the appropriate section on the main how-to index. It doesn't make sense to have additional index pages with so little content. This allows us not to duplicate efforts and create additional nice-looking index pages for each how-to category. Reminder, we are using these category index pages in the first place so that the sidenav has a nice accordion for how-to categories rather than a flat list of all how-to guides. For instance, when you click on the (1) `Data Intake` category on the sidenav, it will bring you to the (2) `## Data Intake` on the main how-to index, rather than some other category-specific index page.

image


philippjfr:
I'm happy with this for now but I think at some point we may have to reorganize the same way that Panel has done because the number of sections will get unwieldy for a single index page.

philippjfr:
Please rebase and incorporate the new Callbacks how-to.

philippjfr (approved):
Looks good, thank you!

PR #432, opened 2023-01-17
Update holoviz_tasks/install

Hoxbro:
Fixing failing CI for python 3.7. 

DuckDB does not work for Python 3.7. Even if it succeeds in installing I can't get the tests to pass:
image

So I have removed the `-o sql` option for Python 3.7.


philippjfr (approved):
Great, thank you!

PR #430, opened 2023-01-17
Add title option to view

Hoxbro:
``` python
import panel as pn
from lumen.layout import Layout
from lumen.pipeline import Pipeline
from lumen.sources import FileSource
from lumen.views import Table

pn.extension("tabulator")

data_url = "https://datasets.holoviz.org/penguins/v1/penguins.csv"

pipeline = Pipeline(source=FileSource(tables={"penguins": data_url}), table="penguins")
pipeline.add_transform("iloc", end=5)

table1 = Table(pipeline=pipeline, title="Title 1", download="csv")
table2 = Table(pipeline=pipeline, title="Title 2")
table3 = Table(pipeline=pipeline)
table4 = Table(pipeline=pipeline, download="csv")

Layout(views=[table1, table2, table3, table4], title="Tables")
````

image

- [x] Need to write some unit tests
- [ ] Document it somewhere in the docs

PR #433, opened 2023-01-17
create how to add title

droumis:
completes doc task from #430

image


philippjfr:
Thanks for this but I think this is getting a little to granular. We can't add a how-to guide for every single parameter.

droumis:
That's fine. I was asked to add it to the docs so I assumed that more than docstring reference was needed.

philippjfr:
I think we just need to work out what the appropriate scope for a how to section is. I could see this being part of a guide about laying out multiple views for example.

droumis:
I'll close it for now. we can revisit later

PR #438, opened 2023-01-31
Improve Download with Index and Filename parameters

Hoxbro:
This PR does two things:

- Add the ability to save or not to save the Index
   - Was possible before with `kwargs={'index': False}`.
   - I have kept it as `True` to follow the default used by pandas `to_xxx` methods. 
- Add Filename
   -  If this is not defined, it will try `self.view.title` and thereafter `self.view.pipeline.table` and clean either of them up to be a valid filename. I have chosen not to do this with the parameter, as a user will expect the actually input to be the filename. 

Code used for the filename:
``` python
import numpy as np
import pandas as pd
import panel as pn
from lumen.sources import FileSource
from lumen.views import Table
from lumen.views.base import Download

pn.extension("tabulator")

np.random.seed(1)
df = pd.DataFrame({"revenue": np.random.uniform(0, 10000, 30)})
df.to_parquet("data.parq")

source = FileSource(files=["data.parq"])
pipeline = Pipeline(source=source, table="data", name="Pipeline")

table = Table(
    pipeline=pipeline,
    download=Download(format="csv", index=False, filename="test"),
    # download=Download(format="csv", index=False),
    title="This is my ??test!!!!",
)
table
```

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -33,7 +33,7 @@ from ..state import state
 from ..transforms.base import Transform
 from ..transforms.sql import SQLTransform
 from ..util import (
-    VARIABLE_RE, catch_and_notify, is_ref, resolve_module_reference,
+    VARIABLE_RE, catch_and_notify, is_ref, resolve_module_reference, slugify,
 )
 from ..validation import ValidationError
 
@@ -51,6 +51,12 @@ class Download(Component, Viewer):
 
     color = param.Color(default='grey', allow_None=True, doc="""
       The color of the download button.""")
+
+    filename = param.String(default=None, doc="""filenamefil
+      The filename of the downloaded table.
+      File extension is added automatic based on the format.
+      If filename is not added, it will be based on the name of the view.
+      """)
 
     format = param.ObjectSelector(default=None, objects=DOWNLOAD_FORMATS, doc="""
       The format to download the data in.""")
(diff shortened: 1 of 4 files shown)


lumen/views/base.py
```diff
@@ -89,18 +98,22 @@ def _table_data(self) -> IO:
...
+            data.to_parquet(io, index=self.index, **self.kwargs)
         io.seek(0)
         return io
 
     def __panel__(self) -> DownloadButton:
-        filename = f'{self.view.pipeline.table}.{self.format}'
+        if self.filename:
+            filename = self.filename
+        else:
+            filename = slugify(self.view.title or self.view.pipeline.table)
```
philippjfr, line 115: Not sure I'd expect it to use the title here. I get that it's a presumably unique label but it could also be quite long and unsuited to usage as a filename.

Hoxbro: It is more descriptive than `self.view.pipeline.table`, especially when multiple branched pipelines are getting the same base data with the same name. This means that all downloaded files will have the same name even if they come from different views.

Basically, I'm trying to avoid confusion about whether the downloaded data is raw or filtered/transformed. I'm open to suggestions; I agree it is not the most elegant way to do it.

philippjfr: Very much get that point but I think adding the`filename` parameter offers a very explicit escape hatch. The 3 way fallback chain is difficult to explain and can result in pretty silly/garbled filenames.

philippjfr: I've got a suggestion and feel free to shoot it down, how about allowing providing both the filename and extension when setting up the download functionality, i.e. setting `download: my_custom_filename.csv` would get parsed out into the format and the filename.

Hoxbro: Why would I shoot down such a good idea :+1: 

Added that in [369ae81](https://github.com/holoviz/lumen/pull/438/commits/369ae81906afff461cb10d6fe68e470cbb63e91c).

philippjfr (approved):
Looks good, thank you!

PR #445, opened 2023-02-09
Remove distinct SQL limit

philippjfr:
Previously we limited the distinct categories but realistically querying these is relatively cheap.


lumen/sources/intake_sql.py
```diff
@@ -20,6 +20,8 @@ class IntakeBaseSQLSource(IntakeBaseSource):
     # Declare this source supports SQL transforms
     _supports_sql = True
 
+    _distinct_limit = 5000
+
```
philippjfr, line 24:
```suggestion
```

PR #450, opened 2023-03-06
Remove usage of the deprecated `.param.params()`

maximlt:
`param.params()` has been deprecated. This PR sets the minimum version of Param required to `1.9.0`, which was released on the 3rd of April 2019, about 3 years ago. It replaces `list(obj.param.params())` by `list(obj.param)`.


philippjfr (approved):
Thanks!

PR #451, opened 2023-03-09
Small improvements

Hoxbro:
Observed that sometimes `schema['inclusiveMinimum']` and `schema['inclusiveMaximum']` could be a string (sometimes).

Also, correctly set `dynamic` to not be True on Bokeh 3.


philippjfr (approved):
Looks good!

PR #458, opened 2023-03-29
Don't sort keys when saving to yaml

Hoxbro:
Observed that when having `Layout(views={"B": B, "A": A})`. The yaml file would have sorted the keys so A was plotted before B.


philippjfr (approved):
Looks good, thanks!

PR #459, opened 2023-04-04
fix a NameError in the script that generates the API reference

maximlt:
The docs built was failing because of that.


philippjfr (approved):
Looks good thanks!

PR #463, opened 2023-04-13
Sanitize string input in SQLFilter

Hoxbro:
Identical to the sanitization made in https://github.com/holoviz/lumen/pull/446


philippjfr (approved):
Looks good!

Issue #460, opened 2023-04-04
Provide ability to hide verbose error messages


philippjfr:
We show error messages when a particular Lumen YAML fails to execute. In production it is usually not desirable to show a verbose error message so we should provide a way to disable this.

PR #465, opened 2023-06-01
Add show_traceback config

Hoxbro:
Partial fix for #460, as I still think it is a problem if it fails before any load. 

image


philippjfr:
LGTM

philippjfr (approved):

Issue #483, opened 2023-09-05
Limit should have a bounds set


ahuang11:
Under View, limit should have a bounds set to (1, None).

Context: AI keeps setting it to 0 for some reason.

PR #484, opened 2023-09-05
Add bounds to limit

ahuang11:
Closes https://github.com/holoviz/lumen/issues/483

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -70,7 +70,7 @@ class View(MultiTypeComponent, Viewer):
         Whether to display a loading indicator on the View when the
         Pipeline is refreshing the data.""")
 
-    limit = param.Integer(default=None, bounds=(1, None), doc="""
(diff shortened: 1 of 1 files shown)


philippjfr:
It's valid to select zero rows so the correct setting should be `bounds=(0, None)`.

lumen/views/base.py
```diff
@@ -70,7 +70,7 @@ class View(MultiTypeComponent, Viewer):
         Whether to display a loading indicator on the View when the
         Pipeline is refreshing the data.""")
 
-    limit = param.Integer(default=None, doc="""
+    limit = param.Integer(default=None, bounds=(1, None), doc="""
```
philippjfr, line 73:
```suggestion
    limit = param.Integer(default=None, bounds=(0, None), doc="""
```

philippjfr (approved):

Issue #455, opened 2023-03-27
Enabling `sync_with_url` for Altair Seattle Weather raises an error when setting the date


Hoxbro:
Enabling `sync_with_url` for Altair Seattle Weather raises an error when setting the date

Update the config to the following:
``` yaml
config:
  title: Altair Seattle Weather
  reloadable: false
  sync_with_url: true
```

And comment out margin because of #454

``` python-traceback
2023-03-27 12:43:22,884 ERROR: panel.reactive - Callback failed for object named "Date" changing property {'value': (datetime.datetime(2015, 12, 9, 0, 0),
           datetime.datetime(2015, 12, 24, 0, 0))} 
Traceback (most recent call last):
  File "/home/shh/Repos/holoviz/panel/panel/reactive.py", line 378, in _process_events
    self.param.update(**self_events)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 1902, in update
    self_._batch_call_watchers()
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2063, in _batch_call_watchers
    self_._execute_watcher(watcher, events)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2025, in _execute_watcher
    watcher.fn(*args, **kwargs)
  File "/home/shh/Repos/holoviz/panel/panel/reactive.py", line 742, in reverse_link
    setattr(self, reverse_links[event.name], event.new)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 367, in _f
    instance_param.__set__(obj, val)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 369, in _f
    return f(self, obj, val)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 1252, in __set__
    obj.param._call_watcher(watcher, event)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2043, in _call_watcher
    self_._execute_watcher(watcher, (event,))
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 2025, in _execute_watcher
    watcher.fn(*args, **kwargs)
  File "/home/shh/Repos/holoviz/panel/panel/reactive.py", line 727, in link_cb
    setattr(target, links[event.name], event.new)
  File "/home/shh/Repos/holoviz/param/param/parameterized.py", line 367, in _f
    instance_param.__set__(obj, val)
[...]

PR #494, opened 2023-10-31
Ensure WidgetFilter sync_with_url uses Widget parameter to serialize values

philippjfr:
Since `Filter.value` is of type Parameter it does not define the correct serialize methods for the Location sync to correct serialize and deserialize the data values. Therefore we now sync the `Widget.value` rather than the `Filter.value` allowing the value type of the parameter to provide the correct serialization. In future `Location.sync` in Panel should provide a dictionary of serializers and deserializers so we can override these directly.

Fixes https://github.com/holoviz/lumen/issues/455

### DIFF
--- a/lumen/filters/base.py
+++ b/lumen/filters/base.py
@@ -63,6 +63,9 @@ class Filter(MultiTypeComponent):
 
     def __init__(self, **params):
         super().__init__(**params)
+        self._setup_sync()
+
+    def _setup_sync(self):
         if self._sync_with_url:
             pn.state.location.sync(self, {'value': self.field}, on_error=self._url_sync_error)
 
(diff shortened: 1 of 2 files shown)


Hoxbro:
The error you are seeing is because you are syncing twice with the widgets.

philippjfr:
You sure? That's what the `self.widget is not None` check is for.

Hoxbro:
Isn't it called once in `Filter` and once in each widget?

philippjfr:
Yes but in Filter self.widget is undefined so it's skipped.

Hoxbro:
You are right.

philippjfr:
The problem really is just that the widget (unlike the Filter) always has a default value, so it gets synced, even if it's an empty list.

Hoxbro (approved):

PR #498, opened 2023-11-01
Ensure View.kwargs are validated

philippjfr:
Attempt at making sure that we validate when we can but also allow passing custom kwargs to `View` types. A `View` that declares a `_panel_type` and passes the kwargs directly to the underlying Panel object can easily be fully validated.

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -23,6 +23,7 @@ from panel.pane.perspective import (
     THEMES as _PERSPECTIVE_THEMES, Plugin as _PerspectivePlugin,
 )
 from panel.param import Param
+from panel.util import classproperty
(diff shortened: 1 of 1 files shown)


lumen/views/base.py
```diff
@@ -146,6 +147,17 @@ def __init__(self, **params):
...
 
+    @classproperty
+    def _valid_keys_(cls) -> List[str] | None:
+        try:
+            panel_type = cls._panel_type
+        except Exception:
+            panel_type = None
+        if panel_type is None:
+            return None
+        valid = super()._valid_keys_()
```
philippjfr, line 158:
```suggestion
        valid = super()._valid_keys_
```

Hoxbro (approved):

PR #517, opened 2024-01-09
Add support for sql expressions

ahuang11:
Abstracts out `IntakeDremioSource` params + `_get_source` into `IntakeBaseDremioSource` and then I just needed to manually build `self.cat` and calling `source.get("table")` works!

### DIFF
--- a/lumen/sources/intake_dremio.py
+++ b/lumen/sources/intake_dremio.py
@@ -3,15 +3,9 @@ import param  # type: ignore
 from .intake_sql import IntakeBaseSQLSource
 
 
-class IntakeDremioSource(IntakeBaseSQLSource):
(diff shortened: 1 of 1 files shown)


philippjfr:
Why create a fake cat if _get_source can just return a DremioSource dynamically?

ahuang11:
Oh you're right. I just saw self.cat in the parent classes, and since it's not an internal variable, I thought it must be defined for all children classes.

ahuang11:
Oh I guess this too:
image

philippjfr:
Ah suppose the intake source always wraps catalogs. Bad abstraction, but let's just keep your version for now.

philippjfr (approved):

PR #520, opened 2024-03-06
Lumen ai outlines

ahuang11:
I added outlines, but I don't think I'm formatting the input messages correctly, and thus, it's able to run, but sometimes crash, sometimes experience the dropped reference issue, and sometimes just returns the original spec.

I think I much prefer instructor.

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -161,14 +161,14 @@ class LumenBaseAgent(Agent):
     user = param.String(default='Lumen')
 
     def _render_lumen(self, component: Component, message: pn.chat.ChatMessage = None):
-        def _render_component(spec, active):
+        async def _render_component(spec, active):
             if active == 0:
-                return pn.indicators.LoadingSpinner(
+                yield pn.indicators.LoadingSpinner(
                     value=True, name="Rendering component...", height=50, width=50
                 )
             # store the spec in the cache instead of memory to save tokens
             memory["current_spec"] = spec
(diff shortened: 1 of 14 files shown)


lumen/ai/agents.py
```diff
@@ -161,14 +161,14 @@ class LumenBaseAgent(Agent):
     user = param.String(default='Lumen')
 
     def _render_lumen(self, component: Component, message: pn.chat.ChatMessage = None):
-        def _render_component(spec, active):
+        async def _render_component(spec, active):
```
ahuang11, line 164: This doesn't fix the dropping reference issue.

lumen/ai/llm.py
```diff
@@ -169,40 +181,83 @@ class OpenAI(Llm):
...
+        system: str = "",
+        **kwargs,
+    ):
+        from outlines import generate
+
+        generator = generate.json(create, kwargs.pop("response_model"))
+        message = "\n\n".join(
+            [f"System: {system}"]
+            + [f"{message['role']}: {message['content']}" for message in messages],
+        )
```
ahuang11, line 250: It doesn't accept the OpenAI format; I don't think this is how I should be formatting it.

ahuang11: I think I should be applying something like https://huggingface.co/blog/chat-templates but I need to load transformers too...

philippjfr:
Closing for now. Can revive later if needed.

PR #524, opened 2024-03-07
Add table listing

ahuang11:
Builds on top of https://github.com/holoviz/lumen/pull/523/


lumen/ai/agents.py
```diff
@@ -290,6 +294,33 @@ def invoke(self, messages: list | str):
         self._render_lumen(pipeline)
 
 
+class TableListAgent(LumenBaseAgent):
```
ahuang11, line 297: I wonder, do we need an agent every time we need a specific task like this, or is there an easier, intelligent way to utilize the LLM?

philippjfr: Yes, indeed.

PR #525, opened 2024-03-07
Make Lumen AI a bit more robuster

ahuang11:
Makes the Lumen AI slightly more robust by:
1. Making it decide whether a transform is needed through the `Decision` model, before going on to select one.
2. Properly loading schema to pass to SQL and Lumen agents, or else it doesn't have context about what columns are available. If I set `load_schema: True` for all, performance significantly degrades; so I went with targeted toggling load_schema = True/False.
3. Minor changes to prompting (wrapping choices in !r)
4. Littered print statements all over for debugging purposes; still very early stage I think we can keep them.
5. Dropping invalid transforms from memory if it turns out its invalid
6. Adding f string to `prompt = f'Current you have the following items in memory: {list(memory)}'` and then commenting it out because if memory is empty, it crashes, and it doesn't seem like it's 100% necessary

We should allow Lumen to be configured so that even if `notifications=True`, the errors can still reach the Agent's for loop retry mechanism (currently I have notifications=False)


lumen/ai/agents.py
```diff
@@ -444,6 +454,16 @@ def _transform_prompt(
...
+        decision = self.llm.invoke(
+            messages,
+            system="Decide whether a transformation is needed; if it's just getting data, return False.",
+            response_model=Decision,
+            allow_partial=False,
+        )
+        if decision is None or not decision.required:
+            print(f"{self.name} decided that no transformation is needed because {decision}")
+            return
+
```
philippjfr, line 466: Makes sense, although we have to be concerned about executing too many calls and slowing things down. We we need to break out distinct Transform, Filter and SQLTransform agents and have the pipeline agent actually try to break down multi-stage transforms into multiple steps.

ahuang11: Counterpoint is that without this, it'll force itself to choose a transform, when none is required, and sometimes it chooses the most irrelevant one (e.g. project_lnglat, likely because it's on the bottom of the list). This causes it to error, and retry with the long message of description of transforms which runs even slower.

I also tried having an additional option of `- None: no transforms are needed` but it still liked to choose a transform. This runs pretty fast, in like a second since the prompt is short and it is memoryless.

```
[{'role': 'system', 'content': "Decide whether a transformation is needed; if it's just getting data, return False."}, {'role': 'user', 'content': 'Get the ml2 glass dataset'}]

Invoked output: Decision(required=False)
```

But yes, I agree we should be concerned about too many steps. Part of me thinks langgraph might be useful here https://github.com/langchain-ai/langgraph/blob/main/examples/multi_agent/multi-agent-collaboration.ipynb https://python.langchain.com/docs/langgraph#reflection--self-critique

I also think I should read https://jxnl.github.io/instructor/tutorials/2-tips/#defining-multiple-entities in detail to better implement some of these models

philippjfr: Oh yes this is absolutely needed just wanted to make sure we're mindful of the cost.

Langraph indeed looks interesting, as we discovered with langchain some of it seems a little over engineered but if nothing else we should look at it for inspiration.

ahuang11: I just saw this https://jxnl.github.io/instructor/examples/planning-tasks/#planning-a-query-plan

PR #529, opened 2024-03-27
Fix the CI

Hoxbro:
Fixing the test CI and aligning it with our other repos.


lumen/tests/conftest.py
```diff
@@ -8,13 +9,20 @@
 import pytest
 
 from bokeh.document import Document  # type: ignore
+from hvplot.tests.util import makeMixedDataFrame
```
Hoxbro, line 12: This is the laziest way I could do it.

PR #535, opened 2024-04-15
Lumen AI instructor v1

ahuang11:
Instructor recently released v1. However, llama-cpp-python is not yet supported; I raised an issue on the repo.


ahuang11:
Stale.

PR #536, opened 2024-04-16
Add async, remove String model, and fix bug with duplicate messages

ahuang11:
Simplifies the streaming; no need for `String.output`; simply remove `response_model`

Then, fixes bug with messages being streamed twice behind the scenes (no need to de-duplicate).

Lastly, makes this program async.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -70,9 +70,10 @@ class Agent(Viewer):
             import traceback
 
             traceback.print_exc()
+            last_message = self.interface.objects[-1]
             self.interface.send(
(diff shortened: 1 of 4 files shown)


lumen/ai/agents.py
```diff
@@ -123,8 +123,8 @@ def _link_code_editor(self, value, callback, language):
         return tabs
 
     def _chat_invoke(self, contents: list | str, user: str, instance: ChatInterface):
-        print("-" * 50)
-        return self.invoke(contents)
```
ahuang11, line 127: Should not return if interface streamed already.

philippjfr (approved):

PR #542, opened 2024-04-25
Allows for multiple transforms to be created in one go and fix pivot

ahuang11:
Previously, the LLM could only generate one transform at a time; this releases that constraint by properly typing and handling multiple transforms.

Then, there was an issue where if pivot `self.values is None`, it took it literally, and said it couldn't find None, so if the value is None, I pop it.

Lastly, `pn.widgets.Tabulator` multi index columns is broken at the moment so I need to flatten the index columns.

`Pivot using t_state as column and t_county as index.`

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -76,6 +76,7 @@ class Agent(Viewer):
                     f"Attempting to fix: {exception}.",
                     user="Exception",
                 )
+                memory.pop("current_pipeline")
             else:
                 self.interface.send(
                     f"Error cannot be resolved: {exception}.",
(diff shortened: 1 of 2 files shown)


lumen/ai/agents.py
```diff
@@ -76,6 +76,7 @@ def _exception_handler(exception):
                     f"Attempting to fix: {exception}.",
                     user="Exception",
                 )
+                memory.pop("current_pipeline")
```
ahuang11, line 79: This is usually where it errors so let it refresh from start.

lumen/ai/agents.py
```diff
@@ -446,8 +447,10 @@ async def _render_component(spec, active):
             try:
                 yield type(component).from_spec(load_yaml(spec)).__panel__()
             except Exception as e:
+                import traceback
+                traceback.print_exc()
```
philippjfr, line 451:
```suggestion
```

philippjfr (approved):

PR #543, opened 2024-04-26
refactor for different model types

ahuang11:
Refactors llm.py so that it's easy to select specific models for each query.

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -61,8 +61,14 @@ class Llm(param.Parameterized):
 
         output = None
         for r in range(self.retry):
-            output = await self.run_client(model_key, messages, **kwargs)
-            break
+            try:
+                output = await self.run_client(model_key, messages, **kwargs)
+                break
+            except Exception as e:
+                print(f"Error encountered: {e}")
+                if 'response_model' in kwargs:
+                    kwargs['response_model'] = response_model
+                messages = messages + [{"role": "system", "content": f"You just encountered the following error, make sure you don't repeat it: {e}" }]
         print(f"\033[33mInvoked LLM output: {output!r}\033[0m")
         return output
 
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -187,9 +187,6 @@ class OpenAI(Llm):
         return {"temperature": self.temperature}
 
     def get_client(self, model_key: str):
-        if client := pn.state.cache.get(model_key):
-            return client
-
         import openai
 
         model_kwargs = self._get_model_kwargs(model_key)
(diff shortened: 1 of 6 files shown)


lumen/ai/llm.py
```diff
@@ -67,14 +61,8 @@ async def invoke(
...
-            try:
-                output = await client(messages=messages, **kwargs)
-                break
-            except Exception as e:
-                print(f"Error encountered: {e}")
-                if 'response_model' in kwargs:
-                    kwargs['response_model'] = response_model
-                messages = messages + [{"role": "system", "content": f"You just encountered the following error, make sure you don't repeat it: {e}" }]
+            output = await self.run_client(model_key, messages, **kwargs)
+            break
```
philippjfr, line 65: Don't understand this, where is the error handling?

lumen/ai/llm.py
```diff
@@ -168,38 +171,40 @@ class OpenAI(Llm):
...
         if self.organization:
             model_kwargs["organization"] = self.organization
-        self._model = openai.AsyncOpenAI(**model_kwargs)
-        self._raw_client = self._model.chat.completions.create
-        self._client = self._create_client(self._raw_client)
-
+        llm = openai.AsyncOpenAI(**model_kwargs)
+        raw_client = llm.chat.completions.create
+        client = self._create_client(raw_client, model)
+        pn.state.cache[model_key] = client
```
philippjfr, line 206: Why cache?

lumen/ai/llm.py
ahuang11: I implemented in a separate PR:
https://github.com/holoviz/lumen/pull/543/files#diff-d56253257f94b517eb742b1213dec9b7ccb3fafa7416be4044987ccd54ed3c83L91

https://github.com/holoviz/lumen/pull/543/files#diff-d56253257f94b517eb742b1213dec9b7ccb3fafa7416be4044987ccd54ed3c83R69-R79

lumen/ai/llm.py
ahuang11: For consistency across the `get_client` methods, but I guess compared to LlamaCpp, it's not really necessary to cache.

philippjfr: Indeed, seems pointless here.

philippjfr:
There's a merge conflict.

ahuang11:
Fixed

lumen/ai/llm.py
philippjfr: I'm not convinced it is, the exception handler will only catch it when it bubbles all the way up to whatever triggered the execution, so it will never retry right?!

ahuang11: I suppose you're right; it won't retry that specific part.

I added it back.

PR #545, opened 2024-05-07
Add `use_logfire`

ahuang11:
Pretty nifty to see things logged.

image

PR #546, opened 2024-05-08
add_logfire again

ahuang11:
Supersedes https://github.com/holoviz/lumen/pull/545

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -18,6 +18,8 @@ class Llm(param.Parameterized):
 
     retry = param.Integer(default=2)
 
+    use_logfire = param.Boolean(default=False)
+
     # Allows defining a dictionary of default models.
     _models_kwargs = {}
 
@@ -209,6 +211,10 @@ class OpenAI(Llm):
         llm = openai.AsyncOpenAI(**model_kwargs)
         raw_client = llm.chat.completions.create
         client = self._create_client(raw_client, model)
+
+        if self.use_logfire:
+            import logfire
+            logfire.instrument_openai(llm)
         return client
 
 class AzureOpenAI(Llm):


lumen/ai/llm.py
```diff
@@ -209,6 +211,10 @@ def get_client(self, model_key: str):
         llm = openai.AsyncOpenAI(**model_kwargs)
         raw_client = llm.chat.completions.create
         client = self._create_client(raw_client, model)
+
+        if self.use_logfire:
+            import logfire
+            logfire.instrument_openai(llm)
```
philippjfr, line 217: Shouldn't we instrument the underlying client rather than the instructor client wrapper, ensuring that we capture the actual prompts that instructor is generating?

philippjfr: Nvm, I misunderstood things I think.

ahuang11: Yeah I believe this is the underlying client and captures all the prompts:

For example:
```
{
...
  "request_data": {
    "model": "gpt-3.5-turbo",
    "messages": [
      {
        "role": "system",
        "content": "The user may or may not want to chat about a particular dataset. Determine whether the provided user prompt requires access to actual data."
      },
      {
        "role": "user",
        "content": "What datasets do you have?"
      },
      {
        "role": "assistant",
        "content": "I have a dataset on wind turbines. The dataset includes information such as case ID, FAA ORS number, FAA ASN number, USGS PR ID, EIA ID, state, county, project name, project year, turbine number, turbine capacity, turbine manufacturer, turbine model, turbine height, rotor diameter, and geographical coordinates. \n\nIf you're interested in exploring this dataset further or have any specific questions about it, feel free to ask!"
      },
      {
        "role": "user",
        "content": "Can you tell me about this dataset"
      }
    ],
    "functions": [
      {
        "name": "DataRequired",
        "parameters": {
          "type": "object",
          "required": [
            "data_required"
          ],
          "properties": {
            "data_required": {
              "type": "boolean",
              "title": "Data Required",
              "description": "Whether the provided user query requires access to data."
            }
          }
        },
        "description": "Correctly extracted `DataRequired` with all the required parameters with correct types"
      }
    ],
    "temperature": 0.2,
    "function_call": {
      "name": "DataRequired"
    }
  },
...
```

Also, logfire only accepts AsyncOpenAI/OpenAI types

PR #548, opened 2024-05-08
make tabs stretch

ahuang11:
Not sure why `sizing_mode="stretch_both"` doesn't work so I needed to use `styles={"min-width": "100%"}`...

image

vs originally:
image


philippjfr:
Fine with this for now.

PR #550, opened 2024-05-11
Lumen AI Replace transforms with SQL

ahuang11:
Completely replaces Lumen transforms with SQL instead for less error prone, and faster output.

Also, reuses reasoning from GPT-4 downstream in messages

Chain of thought really is amazing.


philippjfr:
Sorry if this wasn't clear before but I'm -1 on removing the `PipelineAgent` entirely. We will definitely need it for some of the custom transforms the end user might need in future.

PR #551, opened 2024-05-13
Fuzzy table matching

ahuang11:
Gets the closest match to table keywords if list of tables > 10.

Also removes popping current_pipeline on error because it's unrecoverable once popped.

Lastly, shortens the schema keys and removes dtypes from current_data since schema contains the type.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -864,7 +864,7 @@ class hvPlotAgent(LumenBaseAgent):
         )
 
         # Instantiate
-        spec = dict(kwargs)
+        spec = {key: val for key, val in dict(kwargs).items() if val is not None}
         spec["responsive"] = True
 
         spec["type"] = "hvplot_ui"
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -1,7 +1,7 @@
 import io
 import textwrap
 
-from typing import Literal
+from typing import Literal, Optional, Type
 
 import pandas as pd
 import panel as pn
@@ -19,7 +19,7 @@ from ..dashboard import load_yaml
 from ..pipeline import Pipeline
 from ..sources import FileSource, InMemorySource, Source
 from ..sources.intake_sql import IntakeBaseSQLSource
-from ..transforms.sql import SQLOverride, Transform
+from ..transforms.sql import SQLOverride, SQLTransform, Transform
 from ..views import hvPlotUIView
 from .embeddings import Embeddings
 from .llm import Llm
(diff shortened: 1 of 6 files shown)


lumen/ai/agents.py
```diff
@@ -157,14 +157,31 @@ def _get_schema(self, source: Source | Pipeline, table: str = None):
         schema = dict(schema)
         for field, spec in schema.items():
             if "enum" in spec:
-                schema[field] = dict(spec, enum=spec["enum"][:5])
+                spec.pop("enum")
```
philippjfr, line 160: How much of an effect does this have?

ahuang11: I decided to pop it because I'm worried that it'll give the LLM context of only 5 values, which it might think are the only ones available.

lumen/ai/agents.py
```diff
@@ -157,14 +157,31 @@ def _get_schema(self, source: Source | Pipeline, table: str = None):
...
         return schema
 
+
+    async def _get_closest_tables(self, messages: list | str, tables: list[str], n: int = 3) -> list[str]:
+        table_keywords = (await self.llm.invoke(
+            messages,
+            system=self.system_prompt,
+            response_model=FuzzyTable,
+            allow_partial=False
+        )).keywords
```
philippjfr, line 174: Doesn't this need some actual instructions to extract the keywords most relevant to the table selection? The description of the keywords parameter on the FuzzyTable model does not really seem sufficient.

philippjfr: Seems like instead of giving it the system_prompt it should be a prompt tailored to keyword extraction.

lumen/ai/agents.py
```diff
@@ -157,14 +157,35 @@ def _get_schema(self, source: Source | Pipeline, table: str = None):
...
+        table_keywords = (await self.llm.invoke(
+            messages,
+            system=system,
+            response_model=FuzzyTable,
+            allow_partial=False
+        )).keywords
+        tables = tuple(table.replace('"', "") for table in tables)
+        closest_tables = difflib.get_close_matches(table_keywords, tables, n=n)
+        if not closest_tables:
+            closest_tables = tables[:n]
```
ahuang11, line 181: I think here is a great opportunity to introduce wait-for-human-input-from-widget.

lumen/ai/agents.py
philippjfr: The formatter that is applied should be inserting an ellipsis to indicate that there are potentially more items.

PR #555, opened 2024-05-14
use model_kwargs as param

ahuang11:
Allow setting custom model kwargs


philippjfr (approved):
This seems like it could have been a completely independent change. Can we try to keep PRs separate in future so it's easier to review?

PR #557, opened 2024-05-21
Limit number of rows to be rendered once

ahuang11:
Took me a while to pin this down to prevent the interface from freezing.

It was hanging in `pipeline.to_spec`. Specifically, when it calls `self.param.values()`, that calls `wrapped` in `lumen/sources/base.py`, which calls the underlying `df = method(self, table, **cache_query)` which loads the data.

My reproduction basically re-samples a dataset n-times, concatenates, and exports as parquet, but I noticed if it's not duckdb, potentially we could `use_dask` for other cases?

### DIFF
--- a/lumen/sources/intake.py
+++ b/lumen/sources/intake.py
@@ -28,7 +28,6 @@ class IntakeBaseSource(Source):
                     self.param.warning(
                         f"Could not load {entry.name!r} table with dask."
                     )
-        print(entry)
         return entry.read()
 
     def get_tables(self):
--- a/lumen/sources/intake.py
+++ b/lumen/sources/intake.py
@@ -20,7 +20,6 @@ class IntakeBaseSource(Source):
     __abstract = True
 
     def _read(self, entry, dask=True):
-        print(entry)
         if self.dask or dask:
             try:
                 return entry.to_dask()
--- a/lumen/sources/intake.py
+++ b/lumen/sources/intake.py
@@ -39,6 +39,7 @@ class IntakeBaseSource(Source):
     ) -> Dict[str, Dict[str, Any]] | Dict[str, Any]:
         schemas = {}
         for entry in list(self.cat):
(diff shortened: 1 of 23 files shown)


ahuang11:
Hmm, nevermind it still gets stuck using not duckdb.

ahuang11:
Okay the issue is that it's doing limit AFTER selecting all from table :/
`SELECT * FROM ( SELECT * FROM table ) LIMIT 100`

ahuang11:
I'm stumped; I've reduced the sql_expr to `SELECT * FROM "table" LIMIT 100`, yet it still doesn't load.

I suppose the server could be overloaded from previous requests from prior tests?

lumen/sources/intake.py
```diff
@@ -20,6 +20,7 @@ class IntakeBaseSource(Source):
     __abstract = True
 
     def _read(self, entry, dask=True):
+        print(entry)
```
ahuang11, line 23:
```suggestion
```

lumen/transforms/sql.py
```diff
@@ -97,13 +98,18 @@ class SQLLimit(SQLTransform):
...
+                SELECT
+                    *
+                FROM ( {{sql_in}} )
+                LIMIT {{limit}}
+            """
+            return self._render_template(template, sql_in=sql_in, limit=self.limit)
+        else:
+            # more efficiently apply limit directly to the query
+            template = "{{sql_in}} LIMIT {{limit}}"
+            return self._render_template(template, sql_in=sql_in, limit=self.limit)
```
ahuang11, line 112:
```suggestion
        template = """
            SELECT
                *
            FROM ( {{sql_in}} )
            LIMIT {{limit}}
        """
        return self._render_template(template, sql_in=sql_in, limit=self.limit)
```

ahuang11:
Stale at this point.

PR #572, opened 2024-06-13
Add retry_llm_output decorator

ahuang11:
With these changes, I was able to do a fairly involved table join, from year,season -> date onto iso_time.

image

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -82,12 +82,12 @@ class Agent(Viewer):
             if self._retries_left > 0:
                 self._retries_left -= 1
                 self.interface.send(
-                    f"Taking a different approach to expertly resolve this issue {str(exception)!r} using world-class knowledge.",
+                    f"Taking a different approach to expertly resolve this issue `{exception}` using world-class knowledge.",
                     user="Exception",
                 )
             else:
                 self.interface.send(
-                    f"Error cannot be resolved: {str(exception)!r}.",
+                    f"Error cannot be resolved: `{exception}`.",
                     user="System",
                     respond=False
                 )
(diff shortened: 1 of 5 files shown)


lumen/ai/agents.py
```diff
@@ -739,34 +787,18 @@ async def answer(self, messages: list | str):
 
         tables_sql_schemas = {
             table: {
-                "schema": self._get_schema(source, table, include_min_max=False),
+                "schema": self._get_schema(source, table),
```
ahuang11, line 790: Decided the min/max is still useful for the LLM to know the format of the values.

lumen/ai/agents.py
```diff
@@ -80,12 +82,12 @@ def _exception_handler(exception):
             if self._retries_left > 0:
                 self._retries_left -= 1
                 self.interface.send(
-                    f"Attempting to fix: {exception}.",
+                    f"Taking a different approach to expertly resolve this issue `{exception}` using world-class knowledge.",
```
ahuang11, line 85: If it's stuck in a trough, different approach might resolve it.

lumen/ai/agents.py
```diff
@@ -712,8 +714,54 @@ async def _render_sql_result(query, active):
...
+            return
+        sql_query = message.object.replace("```sql", "").replace("```", "").strip()
+
+        # check whether the SQL query is valid
+        try:
+            transforms = [SQLOverride(override=sql_query), SQLLimit(limit=1)]
+            pipeline = Pipeline(
+                source=source, table=tables[0], sql_transforms=transforms
+            )
+            pipeline.data
```
ahuang11, line 757: This validates the SQL (ensuring it's runnable) before the rendering, so that the LLM can auto-correct itself.

philippjfr (approved):

PR #575, opened 2024-06-14
remove need for sql expr

ahuang11:
The `read_parquet` part is quite finicky since the LLM isn't used to seeing and writing SQL expressions with `read_parquet` included. This led to work on targeted retries such as: https://github.com/holoviz/lumen/pull/572/files#diff-d56253257f94b517eb742b1213dec9b7ccb3fafa7416be4044987ccd54ed3c83R751-R752

However, I discovered, duckdb actually supports reading parquet files directly IF the file extension ends with `.parquet` instead of `.parq` https://duckdb.org/docs/api/python/data_ingestion.html

image

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -740,7 +740,6 @@ class SQLAgent(LumenBaseAgent):
         tables_sql_schemas = {
             table: {
                 "schema": self._get_schema(source, table, include_min_max=False),
-                 "sql": source.get_sql_expr(table)
             } for table in tables
         }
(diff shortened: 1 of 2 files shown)


philippjfr:
Does this work?

```python
lmai.memory["current_source"] = DuckDBSource(
    tables={'windturbines': "read_parquet('windturbines.parq')"},
    uri=":memory",
    initializers=["INSTALL httpfs;", "LOAD httpfs;"],
)
```

philippjfr (approved):

PR #577, opened 2024-06-14
Refactor utils

ahuang11:
Refactors into `format_schema` and `get_schema` into utils, and uses it in invalidate memory.


ahuang11:
Merging this since nothing too major and fixes a bug:
image

Issue #585, opened 2024-06-28
[Lumen AI] Warn about missing API key


droumis:
Lumen AI. Warn about missing API key rather than showing downstream errors, e.g. https://github.com/holoviz/lumen/issues/581

PR #586, opened 2024-06-29
Fix error handling

ahuang11:
Closes https://github.com/holoviz/lumen/issues/585

Rather than having the LLM try to recover from unrecoverable errors, prevent it from trying to.

image

consolidates retries to prevent redundant code and only keeps the targeted retry while removing the catch-all-exceptions-and-retry since it usually fails to do the right thing and wastes tokens

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -34,8 +34,7 @@ from .models import (
 )
 from .translate import param_to_pydantic
 from .utils import (
-    UNRECOVERABLE_ERRORS, describe_data, get_schema, render_template,
-    retry_llm_output,
+    describe_data, get_schema, render_template, retry_llm_output,
 )
 from .views import LumenOutput, SQLOutput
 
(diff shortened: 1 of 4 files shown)


lumen/ai/agents.py
```diff
@@ -69,24 +69,17 @@ class Agent(Viewer):
...
-                    f"Taking a different approach to expertly resolve this issue `{exception}` using world-class knowledge.",
-                    user="Exception",
-                )
-            else:
-                self.interface.send(
-                    f"Error cannot be resolved: `{exception}`.",
-                    user="System",
-                    respond=False
-                )
+            if str(exception) in self.interface.serialize()[-1]["content"]:
```
ahuang11, line 73: Prevent redundant message sent twice.

lumen/ai/assistant.py
```diff
@@ -296,6 +296,7 @@ async def _create_valid_agent(self, messages, system, agent_model, return_reason
             response_model=agent_model,
             allow_partial=False
         )
+
```
ahuang11, line 299:
```suggestion
```

droumis:
tested and works for me when missing API key

philippjfr (approved):

lumen/ai/agents.py
philippjfr: Can you not just serialize the last message?

droumis (approved):

ahuang11:
Planning to merge first and fix the serialize() once https://github.com/holoviz/panel/pull/6967 goes in.

PR #587, opened 2024-07-01
Add sidebar_widgets and notebook_preamble params

ahuang11:
To allow deploying a notebook session in a custom way, but maybe this can be generalized to a custom action callback.

Closes https://github.com/holoviz/lumen/issues/591

### DIFF
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -122,12 +122,16 @@ class SQLOutput(LumenOutput):
             )
 
         source = memory["current_source"]
-        memory["current_table"] = self.name
         if hasattr(source, "add_table"):
+            # self.name -> SQLOutput02697?
+            memory["current_table"] = self.name
             source.add_table(self.name, self.spec)
+            table = self.name
+        else:
+            table = memory["current_table"]
 
         try:
-            pipeline = Pipeline(source=source, table=self.name)
+            pipeline = Pipeline(source=source, table=table)
             df = pipeline.data
             if len(df) > 0:
                 memory["current_data"] = describe_data(df)
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -122,7 +122,6 @@ class SQLOutput(LumenOutput):
             )
 
         source = memory["current_source"]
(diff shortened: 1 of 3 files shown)


lumen/ai/views.py
```diff
@@ -122,10 +122,16 @@ async def _render_component(self):
...
-        memory["current_table"] = self.name
-        source.add_table(self.name, self.spec)
+        if hasattr(source, "add_table"):
+            # self.name -> SQLOutput02697?
+            memory["current_table"] = self.name
+            source.add_table(self.name, self.spec)
+            table = self.name
+        else:
+            table = memory["current_table"]
+
```
ahuang11, line 132:
```suggestion
        source = memory["current_source"]
        memory["current_table"] = self.name
        source.add_table(self.name, self.spec)
```

lumen/ai/views.py
```diff
@@ -122,10 +122,16 @@ async def _render_component(self):
...
         try:
-            pipeline = Pipeline(source=source, table=self.name)
+            pipeline = Pipeline(source=source, table=table)
```
ahuang11, line 134:
```suggestion
            pipeline = Pipeline(source=source, table=self.name)
```

lumen/ai/views.py
```diff
@@ -121,6 +121,7 @@ async def _render_component(self):
                 value=True, name="Executing SQL query...", height=50, width=50
             )
 
+        source = memory["current_source"]
```
ahuang11, line 124:
```suggestion
```

ahuang11:
Okay I think I'm going with `sidebar_widgets` because it gives the user maximum flexibility.

Initially, I was thinking of copying functionality from [ChatInterface.button_properties](https://github.com/holoviz/panel/blob/86eff50dbee38c8665fca5eb66ddf3b3e889bc84/panel/chat/interface.py#L117), but over time, I found that to be a bit limiting (e.g. what if the user wants to use a ToggleButton or FileDownload widget instead?).

philippjfr (approved):

Issue #583, opened 2024-06-28
Lumen ai dependencies (or suggesstions)


droumis:
To run the code below, I needed to also install:

- pydantic
- instructor
- pyarrow (if you upload a parquet)
- if using llm.Llama()
  - huggingface_hub
  - CMAKE_ARGS="-DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_APPLE_SILICON_PROCESSOR=arm64 -DLLAMA_METAL=on" pip install --upgrade --verbose --force-reinstall --no-cache-dir llama-cpp-python
<details> <summary> Code </summary>

```python
import pathlib

import lumen as lm
import lumen.ai as lmai
import panel as pn

from lumen.sources.duckdb import DuckDBSource

pn.extension("tabulator", "codeeditor", inline=False, template="fast")

llm = lmai.llm.OpenAI()

lmai.memory["current_source"] = DuckDBSource(
    tables=["read_parquet('windturbines.parq')"],
    uri=":memory:",
    initializers=["INSTALL httpfs;", "LOAD httpfs;"],
)

assistant = lmai.Assistant(
    llm=llm,
    agents=[
        lmai.agents.SourceAgent,
        lmai.agents.TableAgent,
        lmai.agents.TableListAgent,
        lmai.agents.SQLAgent,
        lmai.agents.PipelineAgent,
        lmai.agents.hvPlotAgent,
        lmai.agents.ChatAgent,
    ],
)
assistant.servable("Lumen.ai")
assistant.controls().servable(area="sidebar")
```

</details>

PR #588, opened 2024-07-01
add ai extra setup

ahuang11:
Closes https://github.com/holoviz/lumen/issues/583

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,15 @@ dependencies = [
 ]
 
 extras_require = {
+    'ai': [
+        'nbformat',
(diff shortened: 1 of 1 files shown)


droumis (approved):
LGTM

Maybe we should have `'huggingface_hub'` be in another item like `extras_require["ai_local"]`, and then any provider-specific item like `'ai_llama'` could include `'ai_local'`... just a suggestion, feel free to ignore.

ahuang11:
No I think it's a good suggestion!

droumis (approved):

Issue #584, opened 2024-06-28
Lumen AI chat UI overlapping hvPlot Explorer


droumis:
![image](https://github.com/holoviz/lumen/assets/6613202/47f73221-9a1a-4edd-8f40-bbb4c01bf4c0)

PR #590, opened 2024-07-01
Prevent overlap of plot and downstream messages

ahuang11:
Closes https://github.com/holoviz/lumen/issues/584
image


droumis:
LGTM

PR #594, opened 2024-07-03
Fix for all obj type df

ahuang11:
If all the types are strings, then min/max will not appear and this will error.
image

```
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 6992, in drop
    raise KeyError(f"{labels[mask].tolist()} not found in axis")
KeyError: "['min', 'max'] not found in axis"
```

### DIFF
--- a/lumen/ai/utils.py
+++ b/lumen/ai/utils.py
@@ -135,7 +135,8 @@ def describe_data(df: pd.DataFrame) -> str:
         if isinstance(df[col].iloc[0], pd.Timestamp):
             df[col] = pd.to_datetime(df[col])
 
-    df_describe_dict = df.describe(percentiles=[]).drop(["min", "max"]).to_dict()
+    df_describe_dict = df.describe(percentiles=[]).drop(
+        ["min", "max"], errors="ignore").to_dict()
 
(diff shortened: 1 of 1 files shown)


lumen/ai/utils.py
```diff
@@ -135,7 +135,8 @@ def format_float(num):
         if isinstance(df[col].iloc[0], pd.Timestamp):
             df[col] = pd.to_datetime(df[col])
 
-    df_describe_dict = df.describe(percentiles=[]).drop(["min", "max"]).to_dict()
+    df_describe_dict = df.describe(percentiles=[]).drop(
+        ["min", "max"], errors="ignore").to_dict()
```
droumis, line 139:
```suggestion      
    describe_df = df.describe(percentiles=[])
    columns_to_drop = ["min", "max"] # present if any numeric
    columns_to_drop = [col for col in columns_to_drop if col in describe_df.columns]
    df_describe_dict = describe_df.drop(columns=columns_to_drop).to_dict()
```

It might be a bit safer not to ignore all errors and to check if these specific cols exist

droumis (approved):
Just one minor optional suggestion.

PR #595, opened 2024-07-03
Fix SQLAgent

ahuang11:
I'm slightly sure that the following...
```
        source = memory["current_source"]
        memory["current_table"] = self.name
        source.add_table(self.name, self.spec)
```
was added for a reason, but it was implemented incorrectly because `self.name` is `SQLOutput02697` so subsequent calls to SQL broke (no table is named `SQLOutput02697`)

This reverts those changes by retrieving the pipeline stored in memory from earlier.

Also, even though the table is still valid, in cases where the Assistant chooses `PipelineAgent`, we need to check if the SQL statement is still valid, and if not refresh it.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -549,6 +549,8 @@ class SQLAgent(LumenBaseAgent):
         if not message:
             return
         sql_query = message.replace("```sql", "").replace("```", "").strip()
+        if not sql_query:
+            raise ValueError("No SQL query was generated.")
 
         # check whether the SQL query is valid
         transforms = [SQLOverride(override=sql_query), SQLLimit(limit=1)]
@@ -556,6 +558,7 @@ class SQLAgent(LumenBaseAgent):
             source=source, table=tables[0], sql_transforms=transforms
         )
         pipeline.data
+        memory["current_pipeline"] = pipeline
         return sql_query
 
     async def answer(self, messages: list | str):
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -10,7 +10,6 @@ from ..downloads import Download
 from ..pipeline import Pipeline
 from ..views.base import Table
 from .memory import memory
-from .utils import describe_data
 
 
 class LumenOutput(Viewer):
@@ -123,14 +122,8 @@ class SQLOutput(LumenOutput):
                 value=True, name="Executing SQL query...", height=50, width=50
             )
 
-        source = memory["current_source"]
-        memory["current_table"] = self.name
-        source.add_table(self.name, self.spec)
+        pipeline = memory["current_pipeline"]
         try:
-            pipeline = Pipeline(source=source, table=self.name)
-            df = pipeline.data
-            if len(df) > 0:
-                memory["current_data"] = describe_data(df)
             table = Table(
                 pipeline=pipeline, pagination='remote',
                 height=458, page_size=12
@@ -144,7 +137,6 @@ class SQLOutput(LumenOutput):
             download_pane.styles = {'position': 'absolute', 'right': '-50px'}
             output = pn.Column(download_pane, table)
             yield output
-            memory["current_pipeline"] = self.component = pipeline
         except Exception as e:
             import traceback
             traceback.print_exc()
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -229,25 +229,31 @@ class Assistant(Viewer):
 
         source = memory.get("current_source")
         spec = get_schema(source, table=table)
-        system = render_template("check_validity.jinja2", table=table, spec=spec)
-        with self.interface.add_step(title="Checking table relevancy...", user="Assistant") as step:
+        sql = memory.get("current_sql")
+        system = render_template("check_validity.jinja2", table=table, spec=spec, sql=sql)
+        with self.interface.add_step(title="Checking memory...", user="Assistant") as step:
             validity = await self.llm.invoke(
                 messages=messages,
                 system=system,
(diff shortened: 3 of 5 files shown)


lumen/ai/views.py
```diff
@@ -123,14 +122,8 @@ async def _render_component(self):
                 value=True, name="Executing SQL query...", height=50, width=50
             )
 
-        source = memory["current_source"]
-        memory["current_table"] = self.name
-        source.add_table(self.name, self.spec)
+        pipeline = memory["current_pipeline"]
```
ahuang11, line 125: This is already set in https://github.com/holoviz/lumen/pull/595/files#diff-d56253257f94b517eb742b1213dec9b7ccb3fafa7416be4044987ccd54ed3c83R561

lumen/ai/agents.py
```diff
@@ -549,13 +549,16 @@ async def _create_valid_sql(self, messages, system, source, tables, sql_expr, er
         if not message:
             return
         sql_query = message.replace("```sql", "").replace("```", "").strip()
+        if not sql_query:
+            raise ValueError("No SQL query was generated.")
 
         # check whether the SQL query is valid
         transforms = [SQLOverride(override=sql_query), SQLLimit(limit=1)]
```
ahuang11, line 556:
```suggestion
        transforms = [SQLOverride(override=sql_query)]
```

lumen/ai/agents.py
```diff
@@ -549,13 +549,16 @@ async def _create_valid_sql(self, messages, system, source, tables, sql_expr, er
...
+        if not sql_query:
+            raise ValueError("No SQL query was generated.")
 
         # check whether the SQL query is valid
-        transforms = [SQLOverride(override=sql_query), SQLLimit(limit=1)]
+        transforms = [SQLOverride(override=sql_query)]
         pipeline = Pipeline(
             source=source, table=tables[0], sql_transforms=transforms
         )
         pipeline.data
```
ahuang11, line 560:
```suggestion
        df = pipeline.data
        if len(df) > 0:
            memory["current_data"] = describe_data(df)
```

philippjfr:
Have to review these changes more closely but the original intent here was to persist the tables to the source so they can all the exported to the notebook.

lumen/sources/duckdb.py
```diff
@@ -88,7 +88,7 @@ def get(self, table, **query):
         if self.filter_in_sql:
             sql_transforms = [SQLFilter(conditions=conditions)] + sql_transforms
         for st in sql_transforms:
-            sql_expr = st.apply(sql_expr.rstrip(";"))
```
ahuang11, line 91: Moved up to `get_sql_expr`

ahuang11:
Okay, this seems to work well when https://github.com/holoviz/panel/pull/6971 is merged.

philippjfr:
Can you explain what type was causing the error in serialization?

ahuang11:
`LumenOutput` and `SqlOutput` results in
```python
  File "/opt/continuum/.conda/envs/default/lib/python3.12/site-packages/panel/chat/feed.py", line 864, in _serialize_for_transformers
    content = str(message)
              ^^^^^^^^^^^^
TypeError: __str__ returned non-string (type LumenOutput)
```

philippjfr:
Seems like the simpler fix there is to implement `__str__` method on the output objects.

philippjfr:
Hmm, maybe that's not quite right, since I don't know what type `message` is above but something is confusing there. It shouldn't error when casting `message` to `str`.

lumen/pipeline.py
```diff
@@ -187,6 +188,11 @@ def _init_callbacks(self):
             self.pipeline.param.watch(self._update_stale, '_stale')
             self.pipeline.param.watch(self._update_data, 'data')
 
+    def _sync_source(self, event):
+        event.old.param.unwatch(self._source_watcher)
```
philippjfr, line 192:
```suggestion
        self.param.unwatch(self._source_watcher)
```

lumen/pipeline.py
```diff
@@ -175,6 +175,7 @@ def _handle_stale(self):
 
     def _init_callbacks(self):
         self.param.watch(self._update_data, ['filters', 'sql_transforms', 'transforms', 'table', 'update'])
+        self._source_watcher = self.param.watch(self._sync_source, 'source')
         self.source.param.watch(self._update_data, self.source._reload_params)
```
philippjfr, line 179:
```suggestion
        self.param.watch(self._sync_source, 'source')
        self._source_watcher = self.source.param.watch(self._update_data, self.source._reload_params)
```

lumen/pipeline.py
```diff
@@ -187,6 +188,11 @@ def _init_callbacks(self):
             self.pipeline.param.watch(self._update_stale, '_stale')
             self.pipeline.param.watch(self._update_data, 'data')
 
+    def _sync_source(self, event):
+        self.param.unwatch(self._source_watcher)
+        self._source_watcher = self.param.watch(self._sync_source, 'source')
```
philippjfr, line 193:
```suggestion
        event.old.param.unwatch(self._source_watcher)
        self._source_watcher = self.source.param.watch(self._update_data, self.source._reload_params)
```

lumen/ai/views.py
```diff
@@ -112,25 +111,25 @@ def __panel__(self):
         return self._tabs
 
 
+    def __str__(self):
```
philippjfr, line 114:
```suggestion
    def __repr__(self):
```

philippjfr:
Will merge. Can follow up later.

philippjfr (approved):

PR #601, opened 2024-07-12
Reuse DuckDB connections and handle table/column special characters

philippjfr:
- Inherits the DuckDB connection
- Fixes various escaping issues with column/table names containing special characters

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import functools
+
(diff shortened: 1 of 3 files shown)


ahuang11 (approved):
I tested it and it works well with uploads!

One thing I wonder if sqlite or duckdb has its own `quote` function that we can borrow, but not a big deal.

philippjfr:
Could you update the tests? I'm running out of time today.

ahuang11:
Yeah I can do that.

PR #602, opened 2024-07-12
Add dialect

ahuang11:
Give context to the LLM which SQL dialect/flavor to use.

### DIFF
--- a/lumen/ai/prompts/sql_query.jinja2
+++ b/lumen/ai/prompts/sql_query.jinja2
@@ -5,6 +5,8 @@ The data for table '{{ table }}' follows the following JSON schema:
 {{ details.schema }}
 ```
 {% endfor %}
+
 Checklist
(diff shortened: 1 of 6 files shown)


lumen/ai/prompts/sql_query.jinja2
```diff
@@ -5,6 +5,8 @@ The data for table '{{ table }}' follows the following JSON schema:
 {{ details.schema }}
 ```
 {% endfor %}
+
 Checklist
 - If it's a date column (excluding individual year/month/day integers), cast as date
 - If no limit is specified, specify 10000 as the limit.
+- Use only `{{ sql_dialect }}` syntax
```
ahuang11, line 12:
```suggestion
- Use only `{{ dialect }}` syntax
```

philippjfr (approved):

PR #607, opened 2024-07-16
Add support for operations and overlays on hvPlotView

philippjfr:
- Adds support for serializing `ParameterizedFunction` instances
- Leverages that support for defining an `operation` on hvPlotView
- Add a `hvOverlayView` that allows defining an overlay composing multiple `hvPlotView`s

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -285,6 +285,7 @@ class View(MultiTypeComponent, Viewer):
                 "full specification for the View."
             )
         spec = spec.copy()
+        view_type = View._get_type(spec.pop('type', None))
         resolved_spec, refs = {}, {}
 
         # Resolve pipeline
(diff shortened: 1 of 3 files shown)


lumen/views/base.py
```diff
@@ -337,6 +336,12 @@ def from_spec(
             if is_ref(value):
                 refs[p] = value
                 value = state.resolve_reference(value)
+            if cls._is_param_function(p):
+                if isinstance(value, dict) and 'type' in value:
+                    func_spec = dict(value)
+                    module_ref = func_spec.pop('type')
+                    func = resolve_module_reference(module_ref)
+                    value = func.instance(**func_spec)
```
ahuang11, line 344: Should there be a comment on what this does?

philippjfr: Sure, if you can find something more descriptive to say than "deserialize".

Issue #606, opened 2024-07-16
LumenAI should output plots/tables to the sidebar


ahuang11:
As to not crowd the chat.

philippjfr:
Not sure I understand this one, plots/tables are the main output of the chat and when screen real estate is most important. Putting them in the sidebar seems to be the worst possible UI.

ahuang11:
If you expand the sidebar width to 800-1000px, it can be something akin to the artifacts sidebar in Claude.

philippjfr:
Think we have this now as part of the explorer UI.

PR #610, opened 2024-07-18
LumenAI Views in Sidebar

ahuang11:
Closes https://github.com/holoviz/lumen/issues/606

https://github.com/user-attachments/assets/3fde43af-7273-4223-9fa4-7677e29e2b5d

This makes expensive views like plots and tables, pop up one at a time, on the sidebar instead of showing up in the chat feed, to prevent it from becoming laggy.

Logically, these views would make more sense to pop up on the right side, but templates only support sidebar on the left. Perhaps in the future, there can be a right_sidebar.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -552,9 +552,7 @@ class SQLAgent(LumenBaseAgent):
         pipeline = Pipeline(
             source=sql_expr_source, table=model.expr_name
         )
-        print("pipeline", pipeline)
         df = pipeline.data
-        print(df)
         if len(df) > 0:
             memory["current_data"] = describe_data(df)
         memory["current_pipeline"] = pipeline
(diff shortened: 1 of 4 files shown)


lumen/ai/agents.py
```diff
@@ -549,7 +552,9 @@ async def _create_valid_sql(self, messages, system, source, tables, errors=None)
         pipeline = Pipeline(
             source=sql_expr_source, table=model.expr_name
         )
+        print("pipeline", pipeline)
         df = pipeline.data
+        print(df)
```
ahuang11, line 557: Remove the prints
```suggestion
        df = pipeline.data
```

ahuang11:
> That said I think we need to carefully think about UI/UX in general and work through some proposals.

I think if we can implement a right-sidebar, that'd be great. Another thought is pushing the view to a modal if we don't want to force user to use a template.

One unaddressed issue with the current PR is exporting it is broken at the moment.

ahuang11:
Superseded by ExplorerUI

PR #611, opened 2024-07-22
Add CustomAnalysisAgent

philippjfr:
Adds an agent that applies custom analysis. An analysis is defined as a `ParameterizedFunction` which accepts a pipeline and outputs a Lumen component which can be rendered. Additionally a `CustomAnalysis` implements an `applies` method that implements whether the analysis can be applied to the current data.

ToDo:

- The agent should somehow be able to tell the assistant which analyses are available.
- Find ways to ensure that the SQL agent doesn't try to perform the analysis itself, e.g. if you ask to find outliers the SQL agent should not try to write a query that does it
- Automatically suggest custom analyses in the UI if they are applicable.

### DIFF
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -16,7 +16,9 @@ from panel.widgets import Button, FileDownload
 from pydantic import create_model
 from pydantic.fields import FieldInfo
 
-from .agents import Agent, ChatAgent, CustomAnalysisAgent
+from .agents import (
+    Agent, ChatAgent, CustomAnalysisAgent, SQLAgent,
+)
 from .export import export_notebook
 from .llm import Llama, Llm
 from .logs import ChatLogs
@@ -361,7 +363,21 @@ class Assistant(Viewer):
             with self.interface.add_step(title="Choosing subagent...") as step:
                 step.stream(f"Assistant decided the {subagent.name[:-5]!r} will provide {', '.join(deps)}.")
                 self._current_agent.object = f"## **Current Agent**: {subagent.name[:-5]}"
-                await subagent.answer(messages)
+
+                if isinstance(subagent, SQLAgent):
+                    custom_messages = messages.copy()
+                    custom_agent = next((agent for agent in self.agents if isinstance(agent, CustomAnalysisAgent)), None)
+                    if custom_agent:
+                        custom_analysis_doc = custom_agent.__doc__.replace("Available analyses include:\n", "")
+                        custom_message = (
+                            f"Avoid doing the same analysis as any of these custom analyses: {custom_analysis_doc} "
+                            f"Most likely, you'll just need to do a simple SELECT * FROM {{table}};"
+                        )
+                        custom_messages.append({"role": "user", "content": custom_message})
+                        print(custom_messages)
+                    await subagent.answer(custom_messages)
+                else:
+                    await subagent.answer(messages)
                 step.success_title = f"Selected {subagent.name[:-5]}"
         return selected
 
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -112,7 +112,7 @@ class Assistant(Viewer):
             )
         else:
             interface.callback = self._chat_invoke
-        interface.callback_exception = "raise"
+        interface.callback_exception = "verbose"
         interface.message_params["reaction_icons"] = {"like": "thumb-up", "dislike": "thumb-down"}
 
         self._session_id = id(self)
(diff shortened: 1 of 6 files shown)


ahuang11:
> Find ways to ensure that the SQL agent doesn't try to perform the analysis itself, e.g. if you ask to find outliers the SQL agent should not try to write a query that does it

I think we need to probably need to do it on the assistant level in this order of operation:

> Assistant
  - Memory Validation
  - Custom Analyses Check
  - Default Agents

> Automatically suggest custom analyses in the UI if they are applicable.

Can you elaborate on this?

philippjfr:
>Can you elaborate on this? Wouldn't this automatically be selected?

The idea behind custom analyses is that you can tailor Lumen.ai to your common use cases. Whenever a new pipeline is defined we check which analyses the user has defined apply to the data (using `.applies()`) and then they can be automatically suggested to the user, much like the initial suggestions feature you implemented. That probably does mean the analysis agent has to be more tightly coupled with the assistant than other agents.

lumen/ai/assistant.py
```diff
@@ -302,6 +306,12 @@ async def _choose_agent(self, messages: list | str, agents: list[Agent], return_
             return agent_names[0]
         self._current_agent.object = "## **Current Agent**: [Lumen.ai](https://lumen.holoviz.org/)"
         agent_model = self._create_agent_model(agent_names)
+
+        for agent in agents:
+            if isinstance(agent, CustomAnalysisAgent):
+                analyses = "\n".join((f"`{analysis.__name__}`: {analysis.__doc__.strip()}" for analysis in agent.analyses))
+                agent.__doc__ = f"Available analyses include:\n{analyses}\nSelect this agent to perform one of these analyses."
+
         system = render_template(
```
ahuang11, line 315: Outputs:
```python

- `TableListAgent`: List all of the available tables or datasets inventory. Not useful if the user requests a specific table.

- `hvPlotAgent`: Generates a plot of the data given a user prompt. If the user asks to plot, visualize or render the data this is your best best.

- `TableAgent`: Displays a single table / dataset. Does not discuss.

- `ChatAgent`: Chats and provides info about high level data related topics, e.g. what datasets are available, the columns of the data or statistics about the data, and continuing the conversation. Is capable of providing suggestions to get started or comment on interesting tidbits. If data is available, it can also talk about the data itself.

- `ChatDetailsAgent`: Responsible for chatting and providing info about details about the table values, e.g. what should be noted about the data values, valuable, applicable insights about the data, and continuing the conversation. Does not provide overviews; only details, meaning, relationships, and trends of the data.

- `CustomAnalysisAgent`: Available analyses include: `CumulativeNinoAnalysis`: Counts the number of consecutive months of any Nino/neutral event. `TotalNinoAnalysis`: Counts the total number of Nino/neutral events. Select this agent to perform one of these analyses.
```

Given: 

```python

class CumulativeNinoAnalysis(CustomAnalysis):

    """
    Counts the number of consecutive months of any Nino/neutral event.
    """

    def __call__(self, pipeline):
        return hvPlotView(pipeline=pipeline, x="year", y="cumulative")


class TotalNinoAnalysis(CustomAnalysis):
    
    """
    Counts the total number of Nino/neutral events.
    """

    def __call__(self, pipeline):
        return hvPlotView(pipeline=pipeline, x="year", y="ntotal")


```

ahuang11:
> Find ways to ensure that the SQL agent doesn't try to perform the analysis itself, e.g. if you ask to find outliers the SQL agent should not try to write a query that does it

Here's one way; I don't think it's the best, but I can't think of another good way.

image


> Automatically suggest custom analyses in the UI if they are applicable.

Will defer this for discussion.

lumen/ai/assistant.py
```diff
@@ -123,10 +125,17 @@ def download_notebook():
 
         llm = llm or self.llm
         instantiated = []
+        self._analyses = []
```
ahuang11, line 128: The assistant has special handling of analyses within the CustomAnalysesAgent and is used throughout to:

1. Show the analyses' `name: docstring` to determine whether to use the CustomAnalysesAgent or not
2. Validate memory--i.e. if the user requests like `Apply YdataProfiling` and `YdataProfiling` in analyses, then assume the data doesn't need refreshing

lumen/ai/assistant.py
```diff
@@ -155,9 +164,15 @@ def download_notebook():
             notebook_button, *self.sidebar_widgets, self._current_agent, Tabs(("Memory", memory))
         )
 
-    def _add_suggestions_to_footer(self, suggestions: list[str], inplace: bool = True):
+    def _add_suggestions_to_footer(
```
ahuang11, line 167: This is now tweaked to show applicable custom analyses to the current pipeline as well.

PR #617, opened 2024-07-25
Hide table list agent if theres only one table

ahuang11:
image


lumen/ai/assistant.py
```diff
@@ -314,6 +314,13 @@ async def _create_valid_agent(self, messages, system, agent_model, return_reason
         return out.agent
 
     async def _choose_agent(self, messages: list | str, agents: list[Agent], return_reasoning: bool = False):
+        current_source = memory.get("current_source")
+        if len(current_source.tables) == 1:
+            agents = [
+                agent for agent in agents
+                if not isinstance(agent, TableListAgent)
```
philippjfr, line 321: Would rather implement an `.applies` method on the agents.

ahuang11:
Added applies; not sure if there's opportunity to add it to some other agents.

image

philippjfr (approved):

PR #625, opened 2024-07-26
Wrap in iframe

ahuang11:
image

I couldn't improve the UX though; when you click on the button `button.loading=True` doesn't trigger for some reason; I tried to create_task/to_thread, but to no avail.


ahuang11:
Okay figured out how to prevent the sync analyses from blocking events: `asyncio.sleep`

https://github.com/user-attachments/assets/888834b4-d939-41ba-8d32-5b55a6ec0976

philippjfr (approved):

philippjfr (approved):

PR #628, opened 2024-07-26
Set load buffer really high

ahuang11:
A temporary fix to make it usable.

Before this PR, it's not able to scroll up at all.

https://github.com/user-attachments/assets/6c246e93-d28b-4e52-97a4-e338f7c44158

### DIFF
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -110,7 +110,7 @@ class Assistant(Viewer):
 
         if interface is None:
             interface = ChatInterface(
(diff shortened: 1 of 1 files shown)


ahuang11:
Actually load_buffer of 5 works better than 10, and is less laggy.

philippjfr:
Quite confused here now, the title doesn't match what was done. Specifically the load_buffer was set to a much LOWER value than the default, which is 50. In my experimentation it behaves much worse with this change.

PR #632, opened 2024-07-31
change max width

ahuang11:
Unlimit the width from
https://github.com/holoviz/panel/blob/310309ed624cb1b65fefbfd28b3521a421585304/panel/chat/message.py#L182

Not sure if I should change the default of ChatMessage, or just do for Lumen AI default, or targeted for Lumen tables. The main reason why it was implemented is because of something like
https://www.freshconsulting.com/insights/blog/uiux-principle-46-text-box-width-should-help-users-read/

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -67,6 +67,8 @@ class Agent(Viewer):
 
     provides = param.List(default=[], readonly=True)
 
+    _max_width = 1200
+
     __abstract = True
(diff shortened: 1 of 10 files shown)


philippjfr:
Suspect we might different max widths for different agents maybe?

ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/634

Issue #626, opened 2024-07-26
[Lumen-AI] Discussion on wrapping processing steps into a single size-constrained and collapsed card


droumis:
Lumen AI often spits out several processing steps per user request. Some of these might occasionally be useful, but they take up a lot of space and aren't the final answer. I wonder if we can take inspiration from Perplexity.ai UI and wrap up processing steps into a smaller accordion area that isn't visually equivalent to the final output (answer):

![image](https://github.com/user-attachments/assets/594fe6d5-358b-4f20-b58e-bf6699dc3d4f)

ahuang11:
Will be closed in https://github.com/holoviz/panel/pull/7022

PR #635, opened 2024-07-31
Add min width to ChatStep

ahuang11:
Inspired by https://github.com/holoviz/lumen/issues/626 I noticed our steps are all jagged. By setting a min width, it looks cleaner.
 
image

But it can expand outward if necessary
image

Requires https://github.com/holoviz/panel/pull/7049

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -26,6 +26,7 @@ from ..state import state
 from ..transforms.sql import SQLOverride, SQLTransform, Transform
 from ..views import VegaLiteView, View, hvPlotUIView
 from .analysis import Analysis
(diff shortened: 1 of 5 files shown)


lumen/ai/agents.py
```diff
@@ -41,6 +41,8 @@
 
 FUZZY_TABLE_LENGTH = 10
 
+pn.chat.ChatStep.min_width = 350
```
philippjfr, line 44: Totally on board with this but I'm not convinced this is the best place to configure this. That said I'm not sure where would be better.

ahuang11: Would a config file be better?

PR #637, opened 2024-07-31
Cache Analysis output

ahuang11:
Upon clicking into another tab without changing the spec, user needs to wait a long time again for the same result, without a cache.

### DIFF
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -78,11 +78,16 @@ class LumenOutput(Viewer):
             value=True, name="Rendering component...", height=50, width=50
         )
 
-        if (self.active != (len(self._tabs)-1)):
+        if (self.active != (len(self._tabs) - 1)):
             return
 
         # store the spec in the cache instead of memory to save tokens
         memory["current_spec"] = self.spec
+        output = pn.state.cache.get(self.spec)
+        if output:
+            yield output
+            return
+
         try:
             self.component = type(self.component).from_spec(load_yaml(self.spec))
             if isinstance(self.component, Pipeline):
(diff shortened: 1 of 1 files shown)


lumen/ai/views.py
```diff
@@ -108,6 +113,7 @@ async def _render_component(self):
                 output = pn.Column(download_pane, table)
             else:
                 output = self.component.__panel__()
+            pn.state.cache[self.spec] = output
```
philippjfr, line 116: This is a giant memory leak.

philippjfr: There was some discussion around having a `state.session_cache` but for now you'll likely have to add a WeakKeyDictionary indexed by the `pn.state.curdoc` somewhere, e.g. `lumen.state`

```
_session_cache = WeakKeyDictionary()
_global_cache = {}
...
def session_cache(self):
    doc = pn.state.curdoc
    if doc is None:
        return self._global_cache
    elif doc not in self._session_cache:
        self._session_cache[doc] = {}
    return self._session_cache[doc]      
```

ahuang11: Can we use the existing memory implementation?
```


class _Memory(Viewer):

    _session_contexts: ClassVar[WeakKeyDictionary[Document, Any]] = WeakKeyDictionary()

    _views: ClassVar[WeakKeyDictionary[Document, Any]] = WeakKeyDictionary()

    _global_view = None

    _global_context = {}

    @property
    def _curcontext(self):
        if state.curdoc:
            if state.curdoc in self._session_contexts:
                context = self._session_contexts[state.curdoc]
            else:
                self._session_contexts[state.curdoc] = context = {}
            return context
        else:
            return self._global_context
```

philippjfr: Hmm, technically yes but I'd still say no, the memory is meant to be used by the LLM, not for our code to cache things.

philippjfr:
This is handled differently now.

PR #638, opened 2024-07-31
Export column analyses

ahuang11:
Maybe the analysis agent should send a short message describing what was generated.

image

image

### DIFF
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -139,7 +139,7 @@ class AnalysisOutput(LumenOutput):
         if config_options:
             options = pn.Param(self.analysis.param, parameters=config_options)
             b = pn.widgets.Button(icon='rocket', name='Run...', on_click=self._rerun, button_type='success', margin=20)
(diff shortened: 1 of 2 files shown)


philippjfr (approved):

lumen/ai/views.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, **params):
         if config_options:
             options = pn.Param(self.analysis.param, parameters=config_options)
             b = pn.widgets.Button(icon='rocket', name='Run...', on_click=self._rerun, button_type='success', margin=20)
-            self._tabs.insert(0, ('Config', pn.Column(options, b)))
+            self._tabs.insert(1, ('Config', pn.Column(options, b)))
```
philippjfr, line 142: Why?

ahuang11: I think at the current moment, the custom serializer expects the spec to be on the first tab

https://github.com/holoviz/lumen/blob/a0a4b9f30d67105b725cc9a05a01bba2cfc28267/lumen/ai/assistant.py#L411-L419

PR #647, opened 2024-08-14
Ensure data provenance even for synthetic tables

philippjfr:
Ensures that we track the provenance of data so it can be serialized. Specifically this implements:

- `DuckDBSource` can now define `mirrors` which mirror existing tables and pipelines into the `DuckDB` database
- `DuckDBSource` can now be declared as synthetic which means the tables do not have provenance and have to be serialized as part of the spec. 
- Pipelines can now chain SQL expressions on top of an existing non-SQL pipeline by mirroring the existing pipeline into duckdb.

In future we may have to figure out if there's some better way to persist the synthetic tables.

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -82,10 +82,18 @@ class DuckDBSource(BaseSQLSource):
                 df = source.get(src_table)
             else:
                 df = mirror.data
+                def update(e, table=table):
+                    self._connection.from_df(e.new).to_view(table)
+                    self._set_cache(e.new, table)
+                mirror.param.watch(update, 'data')
             try:
-                self._connection.from_df(df).to_table(table)
+                self._connection.from_df(df).to_view(table)
             except (duckdb.CatalogException, duckdb.ParserException):
                 continue
+
+    @property
+    def connection(self):
+        return self._connection
 
     @classmethod
     def _recursive_resolve(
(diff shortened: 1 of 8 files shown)


ahuang11:
There seems to be an issue with upload

image

lumen/sources/duckdb.py
```diff
@@ -61,6 +75,63 @@ def __init__(self, **params):
...
+        return spec, refs
+
+    def to_spec(self, context: dict[str, Any] | None = None) -> dict[str, Any]:
+        spec = super().to_spec(context)
+        if 'mirrors' not in spec:
+            return spec
+        mirrors = {}
+        for table, (source, src_table) in spec['mirrors'].items():
+            src_spec = source.to_spec(context=context)
+            if source.synthetic:
```
ahuang11, line 123:
```suggestion
            if hasattr(source, "synthetic") and source.synthetic:
```

lumen/ai/prompts/join_required.jinja2
```diff
@@ -1,6 +1,6 @@
 Determine whether a table join is required to answer the user's query.
 
-The current table `{{ table }}` follows this JSON schema:
+The current table '{{ table }}' follows this JSON schema:
```
ahuang11, line 3: The LLM is using ` in its query, but it's invalid syntax.

philippjfr: Okay please fix on the other PR.

ahuang11: See https://github.com/holoviz/lumen/pull/648

lumen/sources/duckdb.py
ahuang11, line 123:
```suggestion
            if source.synthetic:
```

ahuang11:
Great, looks like its fixed

image

philippjfr:
Will add some tests and fix the Pipeline chaining.

PR #648, opened 2024-08-14
Fix quoting and `read(file.ext)` queries

ahuang11:
It is invalid syntax to use backtick \` but since the LLM was instructed to use it verbatim, it did. So here, it replaces it with `'` instead.

Also fixes `read` queries like `read_csv('test.csv')` so it doesn't become `read_csv('test_csv')`

image

### DIFF
--- a/lumen/ai/prompts/sql_query.jinja2
+++ b/lumen/ai/prompts/sql_query.jinja2
@@ -8,12 +8,15 @@ The data for table '{{ table }}' follows the following JSON schema:
 
 {% if not join_required %}
 It was already determined that no join is required, so only use the existing
-table `{{ table }}` to calculate the result.
+table '{{ table }}' to calculate the result.
 {% endif %}
 
 Checklist
 - Do not write a join if the table already contains all the columns that are needed
 - If it's a date column (excluding individual year/month/day integers), cast as date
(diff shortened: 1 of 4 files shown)


lumen/ai/prompts/sql_query.jinja2
```diff
@@ -8,12 +8,15 @@ The data for table '{{ table }}' follows the following JSON schema:
...
-table `{{ table }}` to calculate the result.
+table '{{ table }}' to calculate the result.
 {% endif %}
 
 Checklist
 - Do not write a join if the table already contains all the columns that are needed
 - If it's a date column (excluding individual year/month/day integers), cast as date
 - If no limit is specified, specify 10000 as the limit.
-- Use table names verbatim; e.g. if table is `read_csv('table.csv')` then use `read_csv('table.csv')` and not `table`
+- Use table names verbatim; e.g. if table is read_csv('table.csv') then use read_csv('table.csv') and not 'table'
```
ahuang11, line 18:
```suggestion
- Use table names verbatim; e.g. if table is read_csv('table.csv') then use read_csv('table.csv') and not 'table'
- If `read_*` is used, use with alias, e.g. read_parquet('table.parq') as table_parq
```

philippjfr (approved):

PR #655, opened 2024-08-16
Make SQL joins a bit more robust

ahuang11:
1. Fix regression with replacing "." with "_"
2. Prompt it to not add extraneous prefixes
3. Left join / based on best matching columns (because one time, it created two SELECT queries)
4. Use slugs instead of names with spaces for expression labels


ahuang11:
There seems to still be issues referencing to previous tables

image

philippjfr (approved):

PR #658, opened 2024-08-19
Improve joins on mismatch and finish select table to join implementation

ahuang11:
Updates prompt so LLM knows to perform nearest join match.

image

On randomized data, it seems to work
image

Another example; runs a bit slow on closest match

image

image

---

This PR also fixes using NestedSelect instead of Upload to join table.

### DIFF
--- a/lumen/ai/prompts/sql_query.jinja2
+++ b/lumen/ai/prompts/sql_query.jinja2
@@ -10,7 +10,7 @@ The data for table '{{ table }}' follows the following JSON schema:
 It was already determined that no join is required, so only use the existing
 table '{{ table }}' to calculate the result.
 {% else %}
-If the join's values do not align 50% of the time, then perform a join based on the
+If the join's values do not align based on the min/max lowest common denominator, then perform a join based on the
 closest match, or resample and aggregate the data to align the values.
 {% endif %}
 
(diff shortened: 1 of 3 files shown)


philippjfr (approved):

lumen/ai/prompts/sql_query.jinja2
```diff
@@ -9,16 +9,19 @@ The data for table '{{ table }}' follows the following JSON schema:
 {% if not join_required %}
 It was already determined that no join is required, so only use the existing
 table '{{ table }}' to calculate the result.
+{% else %}
+If the join's values do not align 50% of the time, then perform a join based on the
```
philippjfr, line 13: How would it know whether it's likely to align?

ahuang11: Potentially schema enums (I'll have to double check dates) or min/max

philippjfr: Should be `min/max`.

ahuang11: Seems to work with min/max
image

ahuang11: Seems to work with string enums too 

image

ahuang11: And lastly for numeric matching min/max

image

philippjfr (approved):

PR #659, opened 2024-08-19
Prevent duplicate analyses from appearing

ahuang11:
A lot of times, I hit `Apply Join` instead of `Use tables`

image

This prevents previous analysis to be the same as current analysis if `_consecutive_calls` is set to False

### DIFF
--- a/lumen/ai/analysis.py
+++ b/lumen/ai/analysis.py
@@ -24,6 +24,8 @@ class Analysis(param.ParameterizedFunction):
 
     _run_button = param.Parameter(default=None)
 
+    _consecutive_calls = False
+
     @classmethod
     def applies(cls, pipeline) -> bool:
         return all(col in pipeline.data.columns for col in cls.columns)
--- a/lumen/ai/analysis.py
+++ b/lumen/ai/analysis.py
@@ -24,7 +24,7 @@ class Analysis(param.ParameterizedFunction):
 
     _run_button = param.Parameter(default=None)
 
-    _consecutive_calls = False
+    _consecutive_calls = False  # whether to allow the analysis to be called multiple times in a row
 
     @classmethod
     def applies(cls, pipeline) -> bool:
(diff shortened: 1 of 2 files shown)


lumen/ai/assistant.py
```diff
@@ -226,9 +226,14 @@ async def run_demo(event):
...
-            analysis for analysis in self._analyses if analysis.applies(pipeline)
-        ]
+        current_analysis = memory.get("current_analysis")
+        applicable_analyses = []
+        for analysis in self._analyses:
+            if analysis.applies(pipeline):
+                if current_analysis and not current_analysis._consecutive_calls:
+                    if analysis.__name__ == current_analysis.__name__:
+                        continue
+                applicable_analyses.append(analysis)
```
philippjfr, line 236: Why compare on `__name__` instead of identity? Also those nested conditionals aren't great, would rather `continue` early.

```suggestion
        current_analysis = memory.get("current_analysis")
        allow_consecutive = getattr(current_analysis, '_consecutive_calls', True)
        applicable_analyses = []
        for analysis in self._analyses:
            if analysis.applies(pipeline) and (allow_consecutive or analysis is not current_analysis):
                applicable_analyses.append(analysis)
```

lumen/ai/analysis.py
```diff
@@ -24,6 +24,8 @@ class Analysis(param.ParameterizedFunction):
 
     _run_button = param.Parameter(default=None)
 
+    _consecutive_calls = False
```
philippjfr, line 27: A comment would be good.

lumen/ai/assistant.py
philippjfr: I did not test this ☝️ to be clear.

ahuang11: It was close :) missing `type`

image

lumen/ai/analysis.py
```diff
@@ -24,6 +24,8 @@ class Analysis(param.ParameterizedFunction):
 
     _run_button = param.Parameter(default=None)
 
+    _consecutive_calls = False  # whether to allow the analysis to be called multiple times in a row
```
philippjfr, line 27:
```suggestion
     # Whether to allow the analysis to be called multiple times in a row
    _consecutive_calls = False
```

PR #678, opened 2024-08-20
Shorten card title

ahuang11:
Long title is ugly

image

Maybe should also handle on the Panel side


philippjfr (approved):
Yes, clearly want left-align and proper wrapping in Panel.

philippjfr (approved):

Issue #671, opened 2024-08-20
Hide output tab until run


ahuang11:
Or at least a better message

<img width="711" alt="image" src="https://github.com/user-attachments/assets/a72ea1b6-361a-4905-8752-8a71c4cbf9b9">

PR #679, opened 2024-08-20
Better error msg when spec is empty and not autorun

ahuang11:
Closes https://github.com/holoviz/lumen/issues/671

image

### DIFF
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -86,11 +86,14 @@ class LumenOutput(Viewer):
         if self.spec in self._last_output:
             yield self._last_output[self.spec]
             return
+        elif self.component is None:
+            yield pn.pane.Alert(
+                "No component to render. Please complete the Config tab.",
+                alert_type="warning",
+            )
+            return
 
         try:
-            if self.component is None or self.spec is None:
-                raise ValueError("No spec provided; please first fill out the Config tab.")
-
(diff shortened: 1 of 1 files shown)


lumen/ai/views.py
```diff
@@ -88,6 +88,9 @@ async def _render_component(self):
             return
 
         try:
+            if self.component is None or self.spec is None:
```
philippjfr, line 91: Seems like `self.spec is None` is already handled above, seems like what actually needs to happen here is for the `self._rendered` check below to be changed.

lumen/ai/views.py
```diff
@@ -88,6 +88,9 @@ async def _render_component(self):
...
+                raise ValueError("No spec provided; please first fill out the Config tab.")
+
             if self._rendered:
```
philippjfr, line 94:
```suggestion
            if self._rendered and self.component is not None:
```

ahuang11: Do we need this if `if self.component is None or...` is going to raise a ValueError?

ahuang11: Oh I just saw your above message; this change doesn't work because it'll just go downstream.

image

ahuang11: image

philippjfr (approved):

PR #683, opened 2024-08-21
Start duckdb implementation of dremio

ahuang11:
Started migrating stuff over to DuckDB source and will start testing.

Based off https://www.dremio.com/blog/using-duckdb-with-your-dremio-data-lakehouse/ 

> With a well-curated semantic layer, Dremio makes access and discovery of data much easier for analysts to pull the data they need and have permission to access using Apache Arrow Flight. They can then proceed locally for ad hoc queries on the subset of the data they pulled from Dremio

So essentially, we still have to read the entire dataset and migrate it into a duckdb source. Previously with intake-dremio, I think we were doing dremio -> intake -> duckdb. now it's dremio -> duckdb

While doing that, would like your feedback on implementation or anything.

I think there are things we can tinker with for optimizing the reads like `read_chunk` and partitioning, etc

```
def dataset(source, schema=None, format=None, filesystem=None,
            partitioning=None, partition_base_dir=None,
            exclude_invalid_files=None, ignore_prefixes=None):
```

### DIFF
--- a/lumen/sources/dremio.py
+++ b/lumen/sources/dremio.py
@@ -1,0 +1,69 @@
+from functools import reduce
+
+from pyarrow import flight
+
+
+class DremioClientAuthMiddleware(flight.ClientMiddleware):
+    """
+    A ClientMiddleware that extracts the bearer token from
+    the authorization header returned by the Dremio
+    Flight Server Endpoint.
+
+    Parameters
+    ----------
+    factory : ClientHeaderAuthMiddlewareFactory
+        The factory to set call credentials if an
+        authorization header with bearer token is
+        returned by the Dremio server.
+    """
+
+    def __init__(self, factory):
+        self.factory = factory
+
+    def received_headers(self, headers):
+        if self.factory.call_credential:
+            return
+
+        auth_header_key = "authorization"
+
+        authorization_header = reduce(
+            lambda result, header: (
+                header[1] if header[0] == auth_header_key else result
+            ),
+            headers.items(),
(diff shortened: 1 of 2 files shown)


lumen/sources/dremio.py
```diff
@@ -0,0 +1,69 @@
+from functools import reduce
+
+from pyarrow import flight
+
+
+class DremioClientAuthMiddleware(flight.ClientMiddleware):
```
ahuang11, line 6: Unsure where I should put this; I pulled it from intake-dremio.

lumen/sources/duckdb.py
```diff
@@ -422,3 +416,53 @@ def _process_uri(self, tls=False, username=None, password=None):
...
+    def _ingest_table(self, table):
+        """
+        Ingests a table from Dremio to DuckDB.
+        """
+        from pyarrow.dataset import dataset as arrow_dataset
+
+        sql_expr = self.get_sql_expr(table)
+        data_table = self._execute_dremio_sql(sql_expr)
+        arrow_ds = arrow_dataset(source=[data_table])
+        self._connection.from_arrow(arrow_ds).to_view(table.replace('"', ''))
```
ahuang11, line 442: Not sure what the best way to handle the quotes. I think dremio needs them to read, but duckdb does not support it, especially in table names.

ahuang11: Okay added quote/unquote methods and the pipeline seems to work well!

PR #692, opened 2024-08-22
Unblock main thread with applies

ahuang11:
If we submit `to_thread` in apply and just run `LagCorrelation.apply(df)` it successfully runs

https://github.com/user-attachments/assets/f8221b08-63d6-4c76-ba0b-00cacf1cbbb8

```python
class LagCorrelation(Transform):

    frequency = param.Selector(default=None, objects=FREQUENCIES)

    lags = param.Integer(default=300)

    parallel = param.Boolean(default=False)

    transform_type = 'lag_correlation'

    def _apply(self, df):
        print(df)
        correlations = {}
        cols = df.columns
        lags = list(range(-self.lags, self.lags))

        for col1, col2 in combinations(cols, 2):
            key = (col1, col2)
            # Compute correlations serially
            results = [corr(df, col1, col2, i, self.frequency) for i in lags]
            correlations[key] = list(zip(lags, results))
        
        sdfs = []
        for (c1, c2), vs in correlations.items():
            sdf = pd.DataFrame(np.array(vs), columns=['Lag', 'Correlation'])
            sdf['Col1'] = c1
            sdf['Col2'] = c2
            sdfs.append(sdf)

        df = pd.concat(sdfs).set_index(['Col1', 'Col2'])
        return df

    async def apply(self, df):
        import asyncio
        result = await asyncio.to_thread(self._apply, df)
        return result

```

However, when I run it through the whole pipeline:

image

That's because `param.parameterized.async_executor(self._watch_data)` doesn't watch/wait...

Also, I tried speeding it up through vectorization:

```python
df_lags = {}
for col in df.columns:
    df_col = df[col]
    for lag in np.arange(-lags, lags):
        df_lag = df_col.shift(lag)
        df_lags[(col, lag)] = df.drop(col, axis=1).corrwith(df_lag)

df2 = (
    pd.DataFrame(df_lags)
    .rename_axis("Col1")
    .T.rename_axis(["Col2", "Lag"])
    .reset_index()
    .melt(["Col2", "Lag"], var_name="Col1", value_name="Correlation")
    .set_index(["Col1", "Col2"])
    .sort_index()
)
```

Even though this one looks semi cleaner, it turned out to be slower in my tests.

### DIFF
--- a/lumen/pipeline.py
+++ b/lumen/pipeline.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
-import asyncio
 import sys
 
+from asyncio import iscoroutinefunction
 from functools import partial
+from inspect import isawaitable
 from itertools import product
 from typing import Any, ClassVar
 
@@ -292,13 +293,23 @@ class Pipeline(Viewer, Component):
 
         # Apply transforms
         for transform in self.transforms:
-            if asyncio.iscoroutinefunction(transform.apply):
-                print(transform, "ASYNC")
-                data = await asyncio.to_thread(transform.apply, data)
+            if iscoroutinefunction(transform.apply):
+                data = await transform.apply(data)
             else:
-                print(transform, "SYNC")
                 data = transform.apply(data)
-        return data
+            if isawaitable(data):
+                data = await data
+
+        if pn_state._unblocked(pn_state.curdoc):
+            with unlocked():
+                self.data = data
+        else:
+            self.data = data
+
+        if state.config and state.config.on_update:
+            pn.state.execute(partial(state.config.on_update, self))
+        self._stale = False
+        self._update_widget.loading = False
 
     def get_schema(self):
         """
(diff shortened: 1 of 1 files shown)


ahuang11:
```python
import panel as pn

pn.extension()

def add_one():
    int_spinner.value = (int_spinner.value + 1)

int_spinner = pn.widgets.Spinner(name='Integer Slider', start=0, end=100000)
int_spinner.show()
# display(int_spinner)
pn.state.add_periodic_callback(add_one, 200)
```


```python
import pandas as pd
from analyses import *

length = 1000000
df = pd.DataFrame({
    "datetime": pd.date_range("2020-01-01", periods=length, freq="s"),
    "value": range(length),
    # "group": ["A", "B", "C", "D"] * int(length / 4)
})

# await LagCorrelation().apply(df)

source = DuckDBSource.from_df({"example": df})

pipeline = Pipeline(source=source, table="example")
await LagAnalysis(pipeline=pipeline);
```

Ehh seems to be closer, but still getting 

```

AttributeError: 'NoneType' object has no attribute 'columns'
```

PR #693, opened 2024-08-23
Do not crash on compute_subset_schema

ahuang11:
image


philippjfr:
Let's just back out of that PR entirely.

PR #695, opened 2024-08-23
Revert speculative cast

ahuang11:
Reverts https://github.com/holoviz/lumen/commit/0365383efcc8c881bd43cfabc10196ba0f036146

### DIFF
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -716,22 +716,6 @@ class BaseSQLSource(Source):
     # Declare this source supports SQL transforms
     _supports_sql = True
 
-    def _compute_subset_schema(self, data):
-        casts = {}
-        for col in data.columns:
-            if not hasattr(data[col], 'str'):
-                continue
-            if data[col].str.isdigit().all():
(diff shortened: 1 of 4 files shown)


philippjfr (approved):

philippjfr:
Thanks, I think what we should do is ensure we limit the number of enum entries. What triggered this change in the first place was that I saw a mistyped float column generating a 100MB JSON schema.

ahuang11:
I am trying to track this down, but I think all of ai/ uses `utils/get_schema` which limits the number of enum entries https://github.com/holoviz/lumen/blob/main/lumen/ai/utils.py#L86-L88

image

My guess is the enum is nested somewhere else in the spec.

philippjfr:
>I am trying to track this down, but I think all of ai/ uses utils/get_schema which limits the number of enum entries https://github.com/holoviz/lumen/blob/main/lumen/ai/utils.py#L86-L88

Sorry, I wasn't talking about Lumen AI, this is for the schema generation of the source itself.

PR #704, opened 2024-09-10
Properly handle joins and propagate invalidation assessment

ahuang11:
I am now properly able to back out of the ephemeral table and consistently get the desired result.

image

image

image

The root issue was that if there were multiple tables within a single source that needed to be joined, the key `a_source_name` was overridden, so no joins really happened.
`sources[a_source_name] = (a_source, a_table)`

I refactored it so the table name is the key; and if there are duplicate it prefixes with the source name and then re-extracts it out when fetching the table.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -363,9 +363,15 @@ class TableAgent(LumenBaseAgent):
 
     @staticmethod
     def _create_table_model(tables):
-        table_model = create_model("Table", table=(Literal[tables], FieldInfo(
-            description="The most relevant table based on the user query; if none are relevant, select the first."
-        )))
+        table_model = create_model(
+            "Table",
+            chain_of_thought=(str, FieldInfo(
+                description="Restate the user query--then reason through which table has the sufficient columns."
+            )),
+            table=(Literal[tables], FieldInfo(
+                description="The most relevant table based on the user query; if none are relevant, select the first."
+            ))
+        )
         return table_model
 
     async def answer(self, messages: list | str):
@@ -376,7 +382,7 @@ class TableAgent(LumenBaseAgent):
             for source in available_sources:
                 for table in source.get_tables():
                     tables_to_source[table] = source
-                    schema = get_schema(source, table, include_min_max=False, include_enum=False, limit=1)
+                    schema = get_schema(source, table, include_min_max=False, include_enum=True, limit=1)
                     tables_schema_str += f"### {table}\n```yaml\n{yaml.safe_dump(schema)}```\n"
         else:
             source = memory["current_source"]
@@ -409,9 +415,10 @@ class TableAgent(LumenBaseAgent):
                         allow_partial=False,
                     )
                     table = result.table
+                    step.stream(result.chain_of_thought)
                 else:
                     table = tables[0]
-                step.stream(f"Selected table: {table}")
+                    step.stream(f"Selected table: {table}")
 
         if table in tables_to_source:
             source = tables_to_source[table]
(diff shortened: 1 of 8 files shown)


lumen/ai/agents.py
```diff
@@ -376,7 +379,7 @@ async def answer(self, messages: list | str):
             for source in available_sources:
                 for table in source.get_tables():
                     tables_to_source[table] = source
-                    schema = get_schema(source, table, include_min_max=False, include_enum=False, limit=1)
+                    schema = get_schema(source, table, include_min_max=False, include_enum=True, limit=1)
```
ahuang11, line 382: An enum can give a bit more context, especially if the column names are ambiguous.

image

lumen/ai/agents.py
```diff
@@ -363,9 +363,12 @@ class TableAgent(LumenBaseAgent):
 
     @staticmethod
     def _create_table_model(tables):
-        table_model = create_model("Table", table=(Literal[tables], FieldInfo(
-            description="The most relevant table based on the user query; if none are relevant, select the first."
-        )))
+        table_model = create_model(
+            "Table",
+            relevant_table=(Literal[tables], FieldInfo(
```
ahuang11, line 368: Being more explicit helps LLM hone in on what it has to do.

lumen/ai/agents.py
```diff
@@ -542,20 +545,24 @@ async def _create_valid_sql(self, messages, system, sources, errors=None):
         if not sql_query:
             raise ValueError("No SQL query was generated.")
 
+        sources = set(tables_to_source.values())
         if len(sources) > 1:
             mirrors = {}
-            for a_source, a_table in sources.values():
+            for a_table, a_source in tables_to_source.items():
```
ahuang11, line 551: Basically reversed it and dropped one layer; I don't think the keys initially were useful.

lumen/ai/agents.py
```diff
@@ -599,50 +606,8 @@ async def _create_valid_sql(self, messages, system, sources, errors=None):
...
-        return sources
-
-    async def answer(self, messages: list | str):
-        source = memory["current_source"]
-        table = memory["current_table"]
-
-        if not hasattr(source, "get_sql_expr"):
-            return None
-
+    async def check_join_required(self, messages, schema, table):
```
ahuang11, line 609: Simple refactoring into a separate util method to reduce code clutter in `answer`

lumen/ai/prompts/check_validity.jinja2
```diff
@@ -1,4 +1,8 @@
-Based on the latest user's query, decide whether the current table and data contains the the required data. If the user is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but there are no location related column the data is invalid.
+Based on the latest user's query, decide whether the current table and data contains the the required data. If the user
+is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to
+whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but
+there are no location related column the data is invalid. However, if the query can be solved through SQL,
+the data is assumed to be valid.
```
philippjfr, line 5: Why add line breaks?
```suggestion
Based on the latest user's query, decide whether the current table and data contains the the required data. If the user is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but there are no location related column the data is invalid. However, if the query can be solved through SQL, the data is assumed to be valid.
```

lumen/ai/prompts/sql_query.jinja2
```diff
@@ -10,6 +10,7 @@ The data for table '{{ table }}' follows the following YAML schema:
 It was already determined that no join is required, so only use the existing
 table '{{ table }}' to calculate the result.
```
philippjfr, line 11:
```suggestion
It was already determined that no join is required, so only use the existing table '{{ table }}' to calculate the result.
```

lumen/ai/prompts/sql_query.jinja2
```diff
@@ -10,6 +10,7 @@ The data for table '{{ table }}' follows the following YAML schema:
...
 {% else %}
+Please perform a join between the necessary tables.
 If the join's values do not align based on the min/max lowest common denominator, then perform a join based on the
 closest match, or resample and aggregate the data to align the values.
```
philippjfr, line 15:
```suggestion
If the join's values do not align based on the min/max lowest common denominator, then perform a join based on the closest match, or resample and aggregate the data to align the values.
```

philippjfr (approved):

PR #705, opened 2024-09-10
Consider count when invalidating

ahuang11:
Previously, if the LLM returns 1 result, and I ask for top 5 results instead, it wouldn't back out since it did contain all the columns. Now I make it consider the number of rows too.

image

### DIFF
--- a/lumen/ai/models.py
+++ b/lumen/ai/models.py
@@ -72,9 +72,11 @@ class Validity(BaseModel):
 
     correct_assessment: str = Field(
         description="""
-        Thoughts on whether the current table meets the requirement
-        to answer the user's query, i.e. table contains all necessary columns,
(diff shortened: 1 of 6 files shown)


lumen/ai/models.py
```diff
@@ -72,9 +72,11 @@ class Validity(BaseModel):
 
     correct_assessment: str = Field(
         description="""
-        Thoughts on whether the current table meets the requirement
-        to answer the user's query, i.e. table contains all necessary columns,
-        unless user explicitly asks for a refresh.
+        Restate the current table and think thru whether the current table meets the requirement
```
philippjfr, line 75:
```suggestion
        Restate the current table and think through whether the current table meets the requirement
```

philippjfr (approved):

PR #707, opened 2024-09-20
Import Application from panel.io.application, drop 3.9, and ruff

ahuang11:
Application was dropped from `panel.io.server` which broke the Lumen imports
https://github.com/holoviz/panel/pull/7262/files#diff-aaf5dc2e57ff38e7660dbe2c7e08cd2b8a34242b382d40f049987b95f605e628L63


ahuang11:
Discovered that Bokeh dropped 3.9 support, therefore Panel too, and so I also had to do drop it for Lumen https://github.com/holoviz/panel/pull/6883

PR #708, opened 2024-09-20
Fix mutation of sources

ahuang11:
I think the issue was two folds: sources mutating locally, and then on dremio, the issue wasn't because it wasn't backing out, but checking whether the source contained the table

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -372,20 +372,17 @@ class TableAgent(LumenBaseAgent):
         return table_model
 
     async def answer(self, messages: list | str):
-        if len(memory["available_sources"]) >= 1:
-            available_sources = memory["available_sources"]
-            tables_to_source = {}
-            tables_schema_str = "\nHere are the table schemas\n"
-            for source in available_sources:
-                for table in source.get_tables():
-                    tables_to_source[table] = source
+        available_sources = memory["available_sources"]
+        tables_to_source = {}
+        tables_schema_str = "\nHere are the tables\n"
+        for source in available_sources:
+            for table in source.get_tables():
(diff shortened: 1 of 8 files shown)


lumen/sources/intake_sql.py
```diff
@@ -40,7 +39,7 @@ def _get_source(self, table):
         try:
             source = self.cat[table]
         except KeyError as e:
-            raise ValidationError(
+            raise KeyError(
```
ahuang11, line 42: When I do `sources = [src for src in available_sources if table in src]` in [`TableAgent`](https://github.com/holoviz/lumen/blob/main/lumen/ai/agents.py#L422)

it crashes because
https://github.com/holoviz/lumen/blob/905ac36c791a8111aef90fa8593eef459b04a6ff/lumen/sources/intake_dremio.py#L41
`__contains__` expects a KeyError

lumen/ai/agents.py
```diff
@@ -372,20 +372,17 @@ def _create_table_model(tables):
...
-                for table in source.get_tables():
-                    tables_to_source[table] = source
+        available_sources = memory["available_sources"]
+        tables_to_source = {}
+        tables_schema_str = "\nHere are the tables\n"
+        for source in available_sources:
+            for table in source.get_tables():
+                tables_to_source[table] = source
+                if isinstance(source, DuckDBSource) and source.ephemeral:
                     schema = get_schema(source, table, include_min_max=False, include_enum=True, limit=1)
```
ahuang11, line 382: Getting schemas for all the tables in a IntakeDremioSource is pretty slow, so limit it to eph sources.

philippjfr (approved):

PR #709, opened 2024-09-20
Add cot to table

ahuang11:
Previously, it would take a detour to select the right table
1. select the wrong table
2. require a join
3. don't perform a join; just use the supposed-to-be-joined table

Before:
image

Now, it directly reasons out the proper table:
image

### DIFF
--- a/lumen/ai/prompts/check_validity.jinja2
+++ b/lumen/ai/prompts/check_validity.jinja2
@@ -1,4 +1,8 @@
-Based on the latest user's query, decide whether the current table and data contains the the required data. If the user is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but there are no location related column the data is invalid. However, if the query can be solved through SQL, the data is assumed to be valid.
+Based on the latest user's query, decide whether the current table and data contains the the required data. If the user
+is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to
+whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but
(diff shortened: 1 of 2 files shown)


lumen/ai/prompts/check_validity.jinja2
```diff
@@ -1,4 +1,8 @@
-Based on the latest user's query, decide whether the current table and data contains the the required data. If the user is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but there are no location related column the data is invalid. However, if the query can be solved through SQL, the data is assumed to be valid.
+Based on the latest user's query, decide whether the current table and data contains the the required data. If the user
+is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to
+whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but
+there are no location related column the data is invalid. If the number of rows is insufficient, the table is invalid.
+However, if the query can be solved through SQL, the data is assumed to be valid.
```
philippjfr, line 5:
```suggestion
Based on the latest user's query, decide whether the current table and data contains the the required data. If the user is referencing "this data" or "that" they probably are referring to the current data. Pay particular attention to whether the data actually contains the columns required to answer the query, e.g. if they are asking for a location but there are no location related column the data is invalid. If the number of rows is insufficient, the table is invalid. However, if the query can be solved through SQL, the data is assumed to be valid.
```

philippjfr: Let's not add pointless linebreaks.

philippjfr (approved):

PR #710, opened 2024-09-20
Sql agent for chat

ahuang11:
Chat feels pretty useless without some SQL backing it

Before:
image

image

After:
image
image


philippjfr (commented):
Why does it have to be SQL? Shouldn't you be able to chat about the current table?

ahuang11:
Okay you're right; it just requires `current_table`, and not necessarily SQL.

PR #719, opened 2024-10-07
Add planning mode for AI Assistant

philippjfr:
Up until now the AI `Assistant` was fairly dumb, it simply tried to pick the agent that would actually satisfy the user query and then resolved which "dependencies" would have to be satisfied to give that agent all the context it needed. Apart from potentially requiring a whole number of calls to the LLM to satisfy all the dependencies this had the drawback that it doesn't actually plan the query, so if some intermediate steps are needed it would never perform them. In the new planning mode it takes a different approach, it gives the LLM the chance to break down the user query into a series of steps each of which will be performed by an agent, the assistant makes a plan to solve the task, gives each agent an instruction on what task it should perform and then we validate that the plan it made correctly provides all the context that is needed. If the plan is invalid we ask it to re-solve it, now with the information on where it went wrong.


lumen/ai/models.py
```diff
@@ -99,3 +100,46 @@ class Topic(BaseModel):
...
+                description="Describe at a high-level how the actions of each expert will solve the user query."
+            ),
+        ),
+        steps=(
+            list[step],
+            FieldInfo(
+                description="A list of tuples of the experts name and instructions to that expert to help him solve the overall task"
+            )
+        ),
+    )
```
ahuang11, line 125: Prefix with `correct_` and mention `refer to your chain_of_thoughts output`. 

```suggestion
    return create_model(
        "Plan",
        chain_of_thought=(
            str,
            FieldInfo(
                description="Describe at a high-level how the actions of each expert will solve the user query."
            ),
        ),
        correct_steps=(
            list[step],
            FieldInfo(
                description="A list of tuples of the experts name and instructions to that expert to help him solve the overall task. Refer to your chain_of_thought output"
            )
        ),
    )
```

lumen/ai/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from . import agents, embeddings, llm  # noqa
 from .agents import Analysis  # noqa
-from .assistant import Assistant  # noqa
+from .assistant import Assistant, Planner  # noqa
```
philippjfr, line 5:
```suggestion
from .assistant import Assistant, PlanningAssistant  # noqa
```

lumen/ai/assistant.py
```diff
@@ -51,6 +55,8 @@ class Assistant(Viewer):
 
     logs_filename = param.String()
 
+    planning = param.Boolean(default=False)
+
```
philippjfr, line 59:
```suggestion
```

PR #720, opened 2024-10-07
Support downloading pdf for YdataProfiler

ahuang11:
Requires weasyprint and PyPDF2 because weasyprint alone had empty pages in between and I couldnt figure out which CSS element was triggering that.

https://github.com/user-attachments/assets/8fe3e749-21b6-4262-b973-e3a96ce7e841

Requires (preferably under conda install section)
  - weasyprint
  - pypdf2

I think this method is preferable over selenium because this includes data/images from HTML tabs

example pdf
[profile_report (7).pdf](https://github.com/user-attachments/files/17285294/profile_report.7.pdf)

And it's also tested on deployment

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -5,6 +5,7 @@ object.
 from __future__ import annotations
 
 import html
+import logging
 import sys
 
 from io import BytesIO, StringIO
@@ -1156,6 +1157,8 @@ class YdataProfilingView(View):
             except ImportError:
                 raise ImportError("weasyprint and PyPDF2 must be installed to save the report as PDF")
 
+            # stop all the long list of warnings from weasyprint
+            logging.getLogger("weasyprint").setLevel(logging.ERROR)
             pdf = weasyprint.HTML(string=report_html).write_pdf()
             with BytesIO(pdf) as buf:
                 reader = PdfReader(buf)
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1147,15 +1147,47 @@ class YdataProfilingView(View):
         df = self.get_data()
         return dict(df=df, **self.kwargs)
 
+    def _save_to_pdf(self, report_html: str):
+        with self._file_download.param.update(loading=True):
+            try:
+                import weasyprint
+
+                from PyPDF2 import PdfReader, PdfWriter
+            except ImportError:
+                raise ImportError("weasyprint and PyPDF2 must be installed to save the report as PDF")
+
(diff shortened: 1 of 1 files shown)


philippjfr:
Was really rather hoping this could happen on the frontend.

ahuang11:
I'm not sure it's possible to do with frontend.

Did some testing with https://docs.profiling.ydata.ai/latest/examples/census/census_report.html

1. using right click -> print, I only see a subset of the dataset profiler

image

2. using HTML focus/print, I run into cross origin issue: 
```
Uncaught SecurityError: Failed to read a named property 'print' from 'Window': Blocked a frame with origin "null" from accessing a cross-origin frame.
    at openAndPrintIframe (testdl.html:17:38)
    at HTMLButtonElement.onclick (testdl.html:10:44)
```
(Same issue if I save the census_report.html locally)

```python
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Iframe Print Example</title>
</head>
<body>
    <iframe id="myIframe" src="https://docs.profiling.ydata.ai/latest/examples/census/census_report.html" width="600" height="400"></iframe>
    <button onclick="openAndPrintIframe()">Open and Print Iframe</button>
    <script>
        function openAndPrintIframe() {
            var iframe = document.getElementById('myIframe');
            console.log(iframe);
            if (iframe) {
                iframe.contentWindow.focus();
                iframe.contentWindow.print();
            } else {
                console.error("Iframe not found");
            }
        }
    </script>
</body>
</html>
```

3. I also try opening the iframe source, and if so, print prompt shows the document correctly

image

However, I can't automate the two step process
```
var newWindow = window.open(iframe.src);
newWindow.onload = function() {
  newWindow.print();
};
```

5. If I use Panel, I also run into shadow root issues.
```python
import panel as pn

pn.extension()

iframe = pn.pane.HTML(
    '<iframe id="myIframe" src="https://docs.profiling.ydata.ai/latest/examples/census/census_report.html" width="600" height="400"></iframe>'
)

button = pn.widgets.Button(name="Print Iframe", button_type="primary")

js_callback = """
var iframe = document.getElementById("myIframe");
if (iframe) {
    var newWindow = window.open(iframe.src);
} else {
    console.error("Iframe not found");
}
"""

button.js_on_click(args={}, code=js_callback)

layout = pn.Column(iframe, button)

layout.show()
```

Would appreciate ideas

PR #726, opened 2024-10-22
Add HtmlPdfDownloadButton panel for YDataProfiler view and fixes analyses button

ahuang11:
Supersedes https://github.com/holoviz/lumen/pull/720

I was initially testing JsPDF, but I discovered it works without it :o

https://github.com/user-attachments/assets/83a1194b-b556-49f9-8687-b6ae18a8a7e4

[Profile Report 2.pdf](https://github.com/user-attachments/files/17468019/Profile.Report.2.pdf)

### DIFF
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -408,7 +408,7 @@ class Assistant(Viewer):
                 step.stream(f"Assistant decided the {agent_name!r} will provide {', '.join(deps)}.")
                 self._current_agent.object = f"## **Current Agent**: {agent_name}"
                 custom_messages = messages.copy()
-                if isinstance(subagent, (TableAgent | SQLAgent)):
+                if isinstance(subagent, (TableAgent, SQLAgent)):
                     custom_agent = next((agent for agent in self.agents if isinstance(agent, AnalysisAgent)), None)
                     if custom_agent:
                         custom_analysis_doc = custom_agent.__doc__.replace("Available analyses include:\n", "")
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -17,7 +17,7 @@ from panel.viewable import Viewer
 from panel.widgets import Button, FileDownload
 
 from .agents import (
-    Agent, AnalysisAgent, ChatAgent, SQLAgent,
+    Agent, AnalysisAgent, ChatAgent, SQLAgent, TableAgent,
 )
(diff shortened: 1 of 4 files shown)


philippjfr (approved):

lumen/ai/assistant.py
```diff
@@ -405,7 +408,7 @@ async def _get_agent(self, messages: list | str):
                 step.stream(f"Assistant decided the {agent_name!r} will provide {', '.join(deps)}.")
                 self._current_agent.object = f"## **Current Agent**: {agent_name}"
                 custom_messages = messages.copy()
-                if isinstance(subagent, SQLAgent):
+                if isinstance(subagent, (TableAgent | SQLAgent)):
```
philippjfr, line 411: What's the benefit over the simpler

```suggestion
                if isinstance(subagent, (TableAgent. SQLAgent)):
```

ahuang11: Ruff complains about the simpler

lumen/panel.py
```diff
@@ -155,3 +156,38 @@ def _button_click(self, event=None):
...
+    _esm = """
+    export function render({ model }) {
+      const button = document.createElement('button');
+      button.textContent = 'Download PDF';
+      button.addEventListener('click', () => {
+        const printWindow = window.open('', '_blank');
+        printWindow.document.write(`
+          <html>
+            <head>
+              <title>Profile Report</title>
```
philippjfr, line 174: We will likely reuse this. No reason to hardcode this.

philippjfr:
Would suggest styling that button to be a little nicer and probably position it to the top right of the output (using absolute or relative positioning).

lumen/panel.py
ahuang11, line 174:
```suggestion
```

lumen/ai/assistant.py
philippjfr: Weird @hoxbro is this now preferred syntax?

ahuang11:
```
lumen/ai/assistant.py:411:20: UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`
```

hoxbro: Disable that rule. It is slower than tuple per Andrews link.

ahuang11:
image

image

PR #733, opened 2024-10-28
add wordllama

ahuang11:
This requires pre-fetching the model which can be tedious to set up in firewalled envs. Otherwise, it's really quick and works a lot better than the original implementation.


ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/764

PR #734, opened 2024-10-28
Drop str support for `messages` kwarg

ahuang11:
We weren't really using string types AFAIK; only very early stage.


ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/738

PR #737, opened 2024-10-29
Various small fixes and improvements

philippjfr:
- Add `LumenAI.show_controls` to toggle sidebar
- Fix `lumen-ai serve` for multiple files
- Allow defining LLM that does not support any form of streaming
- Automatically load font-awesome when using `LumenAI`

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -29,7 +29,10 @@ class Llm(param.Parameterized):
     # Allows defining a dictionary of default models.
     model_kwargs = param.Dict(default={})
 
+    # Whether the LLM supports streaming of any kind
     _supports_stream = True
+
+    # Whether the LLM supports streaming of Pydantic model output
     _supports_model_stream = True
 
     __abstract = True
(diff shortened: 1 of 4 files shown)


lumen/ai/llm.py
```diff
@@ -29,6 +29,7 @@ class Llm(param.Parameterized):
     # Allows defining a dictionary of default models.
     model_kwargs = param.Dict(default={})
 
+    _supports_stream = True
```
ahuang11, line 32: I'm a tad confused about this vs supports_model_stream.

philippjfr: Correct, not convinced we actually need both.

ahuang11: Yes, I believe Mistral doesn't have first class support in instructor.
https://github.com/holoviz/lumen/pull/697/files#diff-dc864a2206353ab4b7498c9a9361f5c45af0c5fd84adbc59fe054515346e55d4R290

ahuang11 (approved):

PR #738, opened 2024-10-29
Rename methods and tweak API

ahuang11:
1. Renames `answer` -> `step_through`
2. Renames `Agent().invoke` -> `respond`
3. Improve typing (messages: list)
4. Use only `step.stream` in `step_through`, and `interface.stream` in `respond`
5. Wrap output from `step_through` in `AgentStepResults` parameterized
6. Migrate agent chain tuples into `AgentChainLink` parameterized
7. Make LLM generate a short title for steps (I was confused why the LLM was doing consecutive SQLAgent calls) and widen the default min_width
8. Prevent plan step from flashing (due to empty string replace=True)

image
image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -68,6 +68,8 @@ class Agent(Viewer):
 
     provides = param.List(default=[], readonly=True)
 
+    _steps_layout = param.ClassSelector(default=None, class_=pn.Column)
+
     _extensions = ()
 
     _max_width = 1200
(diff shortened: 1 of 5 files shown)


philippjfr:
Looking good, but thinking about it again I'm not sure I love `step_through` and I'm also wondering if we actually need both methods or should simplify it to one method with an argument like `render_output`/`render_final`.

ahuang11:
AgentStepResults did seem superfluous besides making the return type the same so I like that idea of merging those two methods

ahuang11:
Maybe render_output can be controlled by the LLM too, e.g. prompt

`Show me the table and the plot of the data`

ahuang11:
Okay, having the LLM decide on whether to render intermediate output works.
image

ahuang11:
Table + Chat example

image

philippjfr (approved):

PR #740, opened 2024-10-29
Add hvPlotAgent as default

ahuang11:
Initially I thought there was a bug with initializing a list of agents https://github.com/holoviz/lumen/pull/739

Then, I was surprised to find that it was not in the default agents given that Lumen is a data viz library (I think?) and depends on hvplot https://github.com/holoviz/lumen/blob/main/pyproject.toml#L31


philippjfr:
The thing here is that both hvPlot and Vega are good choices and in my testing I've found the vega agent more useful.

philippjfr:
I'm gonna go with `VegaLiteAgent` for now.

PR #741, opened 2024-10-31
Add tabulator filter placeholder

ahuang11:
The text input was confusing to me without context, so I added a placeholder
image


philippjfr:
Could you submit a PR in Panel to maybe add placeholders by default if header_filters are enabled?

philippjfr (approved):

PR #744, opened 2024-11-11
Refactor handling of memory and output

philippjfr:
This PR implements two major refactors:

- Memory is no longer exclusively global. Each Agent, Assistant and Control now has a `memory` parameter which can be overridden, either temporarily OR permanently.
- `render_output` now only determines whether `LumenOutput` classes are rendered into the chat interfaces. Otherwise outputs are collected into memory and can be rendered by external components from there. As part of this we also make the `render` method public.

Additionally `memory` now allows declaring callbacks and `rx` expressions to perform actions if a memory item is updated.

### DIFF
--- a/lumen/ai/assistant.py
+++ b/lumen/ai/assistant.py
@@ -622,7 +622,6 @@ class PlanningAssistant(Assistant):
                 'plan_agent.jinja2', agents=list(agents.values()), current_agent=self._current_agent.object,
                 unmet_dependencies=unmet_dependencies, memory=self._memory, table_info=info, tables=available
             )
-            print(system)
             async for reasoning in self.llm.stream(
                 messages=messages,
                 system=system,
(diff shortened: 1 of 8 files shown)


lumen/ai/assistant.py
```diff
@@ -592,17 +611,18 @@ async def _make_plan(
...
             requested.append(next(iter(tables)))
         while reasoning is None or requested:
             info += await self._lookup_schemas(tables, requested, provided, cache=schemas)
             available = [t for t in tables if t not in provided]
             system = render_template(
                 'plan_agent.jinja2', agents=list(agents.values()), current_agent=self._current_agent.object,
-                unmet_dependencies=unmet_dependencies, memory=memory, table_info=info, tables=available
+                unmet_dependencies=unmet_dependencies, memory=self._memory, table_info=info, tables=available
             )
+            print(system)
```
philippjfr, line 625:
```suggestion
```

Issue #735, opened 2024-10-29
Consolidate LumenAI prompting system


ahuang11:
At the current moment, there are multiple entry points for prompting the LLM

1. [docstrings](https://github.com/holoviz/lumen/blob/main/lumen/ai/agents.py#L198-L203) (for assistant to choose agent)
2. [jinja2 template prompts](https://github.com/holoviz/lumen/tree/main/lumen/ai/prompts)
3. system prompt as [parameter](https://github.com/holoviz/lumen/blob/main/lumen/ai/agents.py#L231-L237) (for agents)
4. [on-the-fly system prompts](https://github.com/holoviz/lumen/blob/main/lumen/ai/agents.py#L916) (for analysis agent)
5. [static `BaseModel`s](https://github.com/holoviz/lumen/blob/main/lumen/ai/models.py#L29-L38)
6. [dynamic `BaseModel`s](https://github.com/holoviz/lumen/blob/main/lumen/ai/models.py#L141-L158)

I probably am missing one, but I propose that for maximum customizability, we should migrate all of these to jinja2 templates, and dynamically read the prompts from the jinja2 templates and create models dynamically too.

That way, users can update the prompts to their liking by either passing in `system_prompt_paths` with the corresponding names or updating a `SYSTEM_PROMPTS[key]` dict.

We should also make it easy for users to discover these prompts.

ahuang11:
I looked up all `llm.invoke` and I think I want the prompts to be structured like:

```yaml
prompts/
- assistant:
    - check_validity.jinja2
    - handle_exception.jinja2
    - source_agent.jinja2
    - chat_agent.jinja2
    - chat_details_agent.jinja2
    - table_list_agent.jinja2
    - sql_agent.jinja2
    - hvplot_agent.jinja2
    - analysis_agent.jinja2
    - vega_lite_agent.jinja2
- source_agent:
    - n/a
- chat_agent:
    - require_data.jinja2
    - system.jinja2
- chat_details_agent:
    - choose_topic.jinja2
    - system.jinja2
- table_list_agent:
    - n/a
- sql_agent:
    - require_join.jinja2
    - select_tables.jinja2
    - system.jinja2
- hvplot_agent:
    - fill_plot.jinja2
    - system.jinja2
- analysis_agent:
    - system.jinja2
- vega_lite_agent:
    - system.jinja2
```

Putting that thought aside, I think each agent should be refactored with params like:
```python
class ChatAgent(Agent):

    purpose = param.String(
[...]

PR #751, opened 2024-11-13
Refactor prompting system

ahuang11:
Starts implementing https://github.com/holoviz/lumen/issues/735

Far from finished, but wanted to gather some feedback first on the changes before updating the other agents too.

One issue I found was that if user provides custom `template_paths` strings (i.e. the template themselves and not paths), I'm not sure it's possible to extend it, so I limited it to path types.

Here, I provide an instructions override:

```python
from lumen.ai.agents import ChatAgent, DuckDBSource
from lumen.ai.memory import memory

memory["current_source"] = DuckDBSource(
    tables=["ibtracs.parquet", "nino_oni.parquet"],
    uri=":memory:",
)

agent = ChatAgent(prompt_overrides={"instructions": "Hello"})
memory["current_data"] = "ABC"
print(await agent._render_system_prompt([]))
```

To the following template:
```jinja2
{% extends 'Agent/main.jinja2' %}

{% block instructions %}
Act as a helpful assistant for high-level data exploration, focusing on available datasets and, only if data is
available, explaining the purpose of each column. Offer suggestions for getting started if needed, remaining factual and
avoiding speculation.
{% endblock %}

{% block context %}
{% if tables|length > 1 %}
Available tables:
{{ closest_tables }}
{% elif schema %}
{{ table }} with schema: {{ schema }}
{% endif %}
{% if memory.get('current_data') %}
Here's a summary of the dataset the user just asked about:
```
{{ memory['current_data'] }}
```
{% endif %}
{% endblock %}
```

Which outputs:
"""

Hello

Available tables:
['ibtracs.parquet', 'nino_oni.parquet']

Here's a summary of the dataset the user just asked about:
```
ABC
```
"""

Note "Act as a helpful assistant for high-level..." is now replaced with "Hello", while the context is untouched.


lumen/ai/agents.py
```diff
@@ -201,6 +196,21 @@ async def _select_table(self, tables):
...
+        context["memory"] = self._memory
+        system_prompt = render_template(
+            self.template_paths[prompt_name],
+            prompt_overrides=self.prompt_overrides,
+            **context
+        )
+        return system_prompt
+
+    @abstractmethod
+    async def _render_system_prompt(self, messages: list[Message]) -> str:
```
philippjfr, line 209: By system prompt this refers to the "main" prompt?

ahuang11: Yeah. I suppose I should rename to render_main_prompt

lumen/ai/coordinator.py
```diff
@@ -82,6 +82,17 @@ class Coordinator(Viewer):
     suggestions = param.List(default=GETTING_STARTED_SUGGESTIONS, doc="""
         Initial list of suggestions of actions the user can take.""")
 
+    prompt_templates = param.Dict(default={
```
philippjfr, line 85: Should we factor out a baseclass for Agent and Coordinator with shared methods and parameters?

lumen/ai/agents.py
```diff
@@ -20,6 +19,8 @@
 from pydantic import BaseModel, create_model
 from pydantic.fields import FieldInfo
 
+from lumen.ai.llm import Message
```
philippjfr, line 22: Please use relative imports like everywhere else.

philippjfr (approved):
Soo much better, thank you!

lumen/ai/agents.py
```diff
@@ -798,34 +756,45 @@ class BaseViewAgent(LumenBaseAgent):
...
         )
-        print(f"{self.name} is being instructed that {view_prompt}.")
-
-        # Query
         output = await self.llm.invoke(
             messages,
-            system=system_prompt + view_prompt,
-            response_model=self._get_model(schema),
+            system=system_prompt,
+            response_model=self._output_type,
```
philippjfr, line 797: Not correct, the output_type is the type of the object that is rendered, not the model.

Issue #750, opened 2024-11-13
Create Embeddings and VectorStore


ahuang11:
I am planning to refactor the [existing Embeddings class](https://github.com/holoviz/lumen/blob/main/lumen/ai/embeddings.py#L6).

The purpose is to supply LLMs with up-to-date or private data using retrieval-augmented generation (RAG) because LLMs are trained on static and sometimes outdated datasets, which may not provide accurate information.

Goals:

- Create text embeddings to perform similarity searches.
When a user makes a query, the system compares embeddings to find the most relevant texts, ensuring responses are more contextually relevant.

- Allow users to plug in their own embedding models.
Provide an interface (abstract base classes) so methods stay consistent regardless of which embedding model is used (e.g., OpenAI, HuggingFace, MistralAI, WordLlama).

- Cache embeddings to enable fast lookups.
Use vector stores, like NumPy arrays for in memory storage or DuckDB for persistent storage to store embeddings, speeding up similarity searches and improving overall performance.

---

I propose the following interfaces with minimal methods as to prevent the interface from being too rigid. I did not add `delete` to the interface because some of the stores, like numpy, are not persistent. DuckDb will have it though.

```python
class Embeddings(ABC):
    @abstractmethod
    def embed(self, texts: List[str]) -> List[List[float]]:
        """Generate embeddings for a list of texts."""
        pass


class VectorStore(ABC):
    def __init__(self, embedding_model: 'Embeddings'):
        self.embedding_model = embedding_model

    @abstractmethod
    def add(self, texts: List[str], metadata: Optional[List[Dict]] = None) -> List[int]:
        """
        Add texts and their metadata to the store.
        Returns:
            List[int]: A list of unique text IDs for the added texts.
        """
        pass

    @abstractmethod
    def query(self, text: str, top_k: int = 5) -> List[Dict]:
        """
        Query store for similar texts.
        Returns:
            List[Dict]: List of matching texts with metadata and similarity scores.
        """
        pass
```

Then, I am planning to implement these:

OpenAI Embeddings
MistralAI Embeddings
HuggingFace Embeddings
WordLlama Embeddings

Example:
```python
class OpenAIEmbeddings(Embeddings):
    def __init__(self, api_key: str, model: str = 'text-embedding-3-small'):
        from openai import OpenAI
        self.client = OpenAI()

    def embed(self, texts: List[str]) -> List[List[float]]:
        texts = [text.replace("\n", " ") for text in texts]
[...]

PR #764, opened 2024-11-18
Implement embeddings

ahuang11:
Closes https://github.com/holoviz/lumen/issues/750

Not 100% sure how to integrate embeddings. Currently, I do two queries:
1. embeddings without any filters; e.g. across anything added to the vector store. this however, requires a higher similarity threshold
2. embeddings with a filter for the specific table; lower similarity threshold.

The embeddings are injected to the template, below context.

```python
import lumen.ai as lmai
import panel as pn

from lumen.ai.analysis import Analysis
from lumen.views import YdataProfilingView

pn.extension(
    "tabulator",
    "codeeditor",
    "vega",
    inline=False,
    template="fast",
    notifications=True,
)

class Profiling(Analysis):
    """
    Generates a profiling report for the given dataset.
    """

    def __call__(self, pipeline):
        return YdataProfilingView(pipeline=pipeline)


llm = lmai.llm.OpenAI()

cs = lmai.memory["current_source"] = DuckDBSource(
    tables=["windturbines.parquet"],
    uri=":memory:",
    initializers=["INSTALL httpfs;", "LOAD httpfs;"],
)

custom_agent = lmai.agents.AnalysisAgent(analyses=[Profiling])
vs = lmai.NumpyVectorStore()
vs.add([
    {"text": "The secret number is 88."}
])

assistant = lmai.Planner(
    llm=llm,
    agents=[
        lmai.agents.SourceAgent,
        lmai.agents.TableListAgent,
        lmai.agents.SQLAgent,
        lmai.agents.hvPlotAgent,
        lmai.agents.ChatAgent,
        custom_agent,
    ],
    vector_store=vs,
)
assistant.servable("Lumen.ai")
```

image

```python
from lumen.ai.embeddings import NumpyEmbeddings
from lumen.ai.vector_store import NumpyVectorStore

# Initialize embeddings and vector store
embeddings = NumpyEmbeddings()
vector_store = NumpyVectorStore(embeddings)

# Add items
items = [
    {
        "text": "OpenAI creates innovative AI technologies.",
        "metadata": {"category": "AI", "type": "c"},
    },
    {
        "text": "Python is widely used in data science.",
        "metadata": {"category": "Programming", "type": "b"},
    },
    {
        "text": "Neural networks power modern deep learning systems.",
        "metadata": {"category": "AI", "type": "a"},
    },
]

text_ids = vector_store.add(items)
print("Added Text IDs:", text_ids)

# Query the store
query_text = "Python"
results = vector_store.query(
    query_text, top_k=1, filters={"category": "AI"}
)
print("Query Results:", results)

> Added Text IDs: [1, 2, 3]
Query Results: [{'id': 3, 'text': 'Neural networks power modern deep learning systems.', 'metadata': {'category': 'AI', 'type': 'a'}, 'similarity': 0.0}]
```
(Chooses ID 3 because of the filter)


lumen/ai/agents.py
```diff
@@ -135,36 +132,8 @@ def __panel__(self):
...
-
-        table_keywords = [keyword.lower() for keyword in fuzzy_table.keywords]
-        tables_lower = {table.lower(): table for table in tables}
-
-        # get closest matches
-        closest_tables = set()
-        for keyword in table_keywords:
-            closest_tables |= set(difflib.get_close_matches(keyword, list(tables_lower), n=5, cutoff=0.3))
-        closest_tables = [tables_lower[table] for table in closest_tables]
+        results = self.vector_store.query(messages[-1], top_k=n, filters={"category": "table_list"})
```
ahuang11, line 135: Not sure if I should just use the last message.

lumen/ai/actor.py
```diff
@@ -4,33 +4,65 @@
...
+        The paths to the prompt's jinja2 templates.""",
+    )
 
-    prompt_templates = param.Dict(default={}, doc="""
-        The paths to the prompt's jinja2 templates.""")
+    def _add_embeddings(self, messages: list[Message], context: dict[str, Any]) -> Any:
+        for message in messages:
+            if message["role"] == "user":
+                break
+        content = message["content"]
```
ahuang11, line 37: Not sure if I should just use the last message or past few.

ahuang11:
Hmm doesn't work well with actual data; I think it needs to be chunked.

ahuang11:
Seems to work:

image
image

image

lumen/ai/actor.py
```diff
@@ -4,26 +4,66 @@
...
+        context["embeddings"] = embeddings
+        return context
 
     def _render_prompt(self, prompt_name: str, **context) -> str:
         context["memory"] = self._memory
         prompt = render_template(
             self.prompt_templates[prompt_name],
-            overrides=self.prompt_overrides.get(prompt_name, {}),
-            **context
+            prompt_overrides=self.prompt_overrides.get(prompt_name, {}),
```
ahuang11, line 65:
```suggestion
            overrides=self.prompt_overrides.get(prompt_name, {}),
```

ahuang11, line 65:
```suggestion
            overrides=self.prompt_overrides.get(prompt_name, {}),
```

philippjfr:
Okay, we can still back out of my changes but I think I hit on what I think is a cleaner and more composable design for embeddings. Specifically I did introduce a `Tool` class and added the ability to add a list of tools to the `prompts` dictionary. When we generate a prompt we now evaluate any tools listed therein and use the resulting output as additional context. This makes it possible to add additional tools for any prompt and means we are not hardcoding the mechanism by which tables are looked up.

Some to do items to achieve the previous behavior:

- Potentially re-add the schema to the `ChatAgent` so it can talk about the data
- Potentially allow the user to add a list of documents or text upfront which then automatically adds an EmbeddingLookupTool (or similar) to any prompts that would benefit

ahuang11:
Thanks for introducing tools! I think it's necessary for release.

My only concern with this is that it subclasses Actor, which has a LLM param, but I think most tools do not depend on LLMs. Is this intentional?

philippjfr:
>My only concern with this is that it subclasses Actor, which has a LLM param, but I think most tools do not depend on LLMs. Is this intentional?

I've changed my mind about this. Earlier I suggested we wouldn't need LLM everywhere but I've decided it's not really harmful to have it and a pain to add if it is needed. My plan for next steps would be to add a `FunctionTool` class, which works as follows:

1. Extract the signature of the user function (ideally including the per field docstrings)
2. Any arguments that require values in `memory` are auto-populated
3. Anything else we create a pydantic Model Field for and ask the LLM to populate the values
4. We then invoke the function with arguments provided by the LLM and the memory and provide the result as context

So at minimum the `FunctionTool` will require an LLM to correctly invoke the function. Once we have that the user can simply supply type annotated functions as tools and doesn't actually have to implement their own `Tool` class.

philippjfr:
I wouldn't particularly object having a mixin class to add the LLM but not sure it's needed.

ahuang11:
Okay, I suppose I don't have a concern about having a LLM param.

Initially, I was thinking of Pydantic, where if you don't specify a default/default_factory, it's required.

But with param, as long as it's not an arg in `__init__`, all params default to None if unspecified, e.g.
```
from lumen.ai.actor import Actor
Actor()
```

ahuang11:
Also, Marvin has a nice decorator that turns a function with docstring into a function call
https://www.askmarvin.ai/docs/text/functions/

For reference:
https://github.com/PrefectHQ/marvin/blob/1f04d10e5ba90144000bb87e3a0a80aef3894b03/src/marvin/ai/text.py#L531
https://github.com/PrefectHQ/marvin/blob/main/src/marvin/utilities/python.py#L114

ahuang11:
Okay, I think this is ready now!

I made it so ChatAgent is able to see the schemas for table and also table**s** because I think ChatAgent can provide ideas on what to do across datasets, not just one single one.

We should also unify the language between tables and datasets; I feel like the LLM gets confused on it.

Export-1733518902647

ahuang11:
Oh I just realized `_render_main_prompt` was removed.

Also, I realized, the tool is run after `render_prompt` so devs should not expect the memory created in the tool to be available for use with current agent.

That means, ChatAgent is able to find all the relevant tables, BUT cannot use that info yet. Instead, I force table list agent to provide closest_tables if less than 5 tables, but I'm not convinced this is the best way...

philippjfr:
Sounds good!

philippjfr (approved):

PR #776, opened 2024-11-20
Add SnowflakeSource

philippjfr:
Adds a `SnowflakeSource` using the snowflake-connector-python library.


lumen/sources/snowflake.py
```diff
@@ -0,0 +1,169 @@
...
+
+    filter_in_sql = param.Boolean(default=True, doc="""
+        Whether to apply filters in SQL or in-memory.""")
+
+    sql_expr = param.String(default='SELECT * FROM {table}', doc="""
+        The SQL expression to execute.""")
+
+    tables = param.ClassSelector(class_=(list, dict), doc="""
+        List or dictionary of tables.""")
+
```
jbednar, line 54: Few if any of these parameters seem specific to Snowflake compared to any other authenticated SQL database. I guess it's fine to define them here for now, and move the shared parameters to a base class once support for the next type authenticated SQL is added? I actually can't detect any code in this class that is specific to Snowflake, so maybe this class is actually just a prototype for AuthenticatedSQLSource, tested on Snowflake so far but potentially usable for other systems as well?

ahuang11: We might want to consider using SQLAlchemy as the interface to specific SQL engines.

philippjfr: I'm not convinced an `AuthenticatedSQLSource` will ever really make sense. Authentication mechanisms will differ between different providers, so maybe `user` and `password` could be universal but things like, authenticator, account, token, warehouse etc. are probably not universal.

jbednar: Sounds like it's fine for now and a shared superclass can be created once we observe what's universal in practice.

Issue #752, opened 2024-11-13
"What datasets do you have?" should defer to ChatAgent if only one table


ahuang11:
![Image](https://github.com/user-attachments/assets/36babdaa-56d6-4f63-a1bd-f811cd239844)

PR #779, opened 2024-11-21
Use ChatAgent for single table

ahuang11:
Closes https://github.com/holoviz/lumen/issues/752

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -277,7 +277,7 @@ class ChatAgent(Agent):
         statistics about the data, and continuing the conversation.
 
         Is capable of providing suggestions to get started or comment on interesting tidbits.
-        If data is available, it can also talk about the data itself.
+        It can talk about the data, if available.
         If there is only one table available, use this instead of TableListAgent.
         """)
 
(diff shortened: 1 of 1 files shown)


lumen/ai/agents.py
```diff
@@ -277,7 +277,9 @@ class ChatAgent(Agent):
         statistics about the data, and continuing the conversation.
 
         Is capable of providing suggestions to get started or comment on interesting tidbits.
-        If data is available, it can also talk about the data itself.""")
+        If data is available, it can also talk about the data itself.
```
amaloney, line 280: nit: the original sentence is a weak statement, but it is stronger if you switch the clauses separated by the comma

```suggestion
        It can talk about the data, if available.
```

lumen/ai/agents.py
```diff
@@ -277,7 +277,9 @@ class ChatAgent(Agent):
...
+        If there is only one table available, use this instead of TableListAgent.
```
amaloney, line 281: similar suggestion

```suggestion
        Use this instead of TableListAgent if there is only one table available.
```

amaloney (approved):
language nits

PR #780, opened 2024-11-21
Tweak local model settings

ahuang11:
Pins version and makes minor tweaks to config.

Seems promising with Qwen; couldn't get it running with Mistral 7B


ahuang11:
Prior to this, since Llama.invoke isn't truly async, there was no loading indicator and it was confusing whether Lumen was doing anything, but this PR changes that:
image

image

amaloney (approved):
lgtm

PR #781, opened 2024-11-23
Better error handling

ahuang11:
1. Ensure there are messages to serialize or else `    if str(exception) in self.interface.serialize()[-1]["content"]:
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
IndexError: list index out of range`
2. If user inputs `Path` type instead of string
3. Prevent `load_button` from always `loading=True`, especially on error
4. Allow starting explorer ui without any datasets e.g. `lumen-ai serve` or `lmai.ExplorerUI(llm=llm, agents=[lmai.agents.VegaLiteAgent]).servable()`

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -94,7 +94,8 @@ class Agent(Viewer, Actor):
 
     def __init__(self, **params):
         def _exception_handler(exception):
-            if str(exception) in self.interface.serialize()[-1]["content"]:
+            messages = self.interface.serialize()
+            if messages and str(exception) in messages[-1]["content"]:
                 return
 
             import traceback
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -37,7 +37,7 @@ from .memory import _Memory, memory
 if TYPE_CHECKING:
     from .views import LumenOutput
 
-DataT = str | Source | Pipeline
+DataT = str | Path, Source | Pipeline
 
 
 class UI(Viewer):
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -37,7 +37,7 @@ from .memory import _Memory, memory
 if TYPE_CHECKING:
     from .views import LumenOutput
 
-DataT = str | Path, Source | Pipeline
+DataT = str | Path | Source | Pipeline
 
 
 class UI(Viewer):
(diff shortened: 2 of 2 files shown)


lumen/ai/agents.py
```diff
@@ -94,7 +94,8 @@ class Agent(Viewer, Actor):
 
     def __init__(self, **params):
         def _exception_handler(exception):
-            if str(exception) in self.interface.serialize()[-1]["content"]:
+            messages = self.interface.serialize()
+            if messages and str(exception) in messages[-1]["content"]:
```
amaloney, line 98: what's the magic number `-1` here? Will the exception always be in the last message? Can we be sure of this always?

lumen/ai/ui.py
```diff
@@ -142,7 +143,8 @@ def _resolve_data(self, data: DataT | list[DataT] | None):
                 sources.append(src)
             elif isinstance(src, Pipeline):
                 mirrors[src.name] = src
-            elif isinstance(src, str):
+            elif isinstance(src, (str, Path)):
```
amaloney, line 146: if it is already a string, you don't need to reconvert it to one

```suggestion
            elif isinstance(src, Path):
```

lumen/ai/ui.py
```diff
@@ -360,22 +358,22 @@ def update_source_map(_, __, sources, init=False):
...
-            tabs.objects = explorers + [controls]
-            load_button.loading = False
+            with load_button.param.update(loading=True):
+                explorers = []
+                for table in tables:
+                    source = source_map[table]
+                    if len(memory['available_sources']) > 1:
+                        _, table = table.rsplit(' : ', 1)
+                    pipeline = Pipeline(
+                        source=source, table=table, sql_transforms=[SQLLimit(limit=100_000)]
```
amaloney, line 368: btw +1 for the `_` separating integer groupings

amaloney (approved):
If we are certain that the exception message will always and forever be the last item in the messages list, then this LGTM.

Otherwise my nit here would be to add a comment indicating that we "expect" it to always be the last item in the list.

lumen/ai/ui.py
ahuang11: Calling str(src) is for converting the Path to string so we don't have to separate out the rest of the clause:

```python
                if src.startswith('http'):
                    remote = True
...
```

lumen/ai/agents.py
ahuang11: It's not always the last, but if it is, don't resend the exception.

PR #783, opened 2024-11-23
disable chat and suggestions on busy

ahuang11:
While loading a large dataset in the overview, I tried entering a message, but nothing popped up because the upload was blocking.

To prevent confusion, I think we should block the chat if it's busy.

image

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -97,9 +97,6 @@ class Coordinator(Viewer, Actor):
         logs_filename: str = "",
         **params,
     ):
-        def on_busy(event):
-            self.interface.disabled = event.new
-
         def on_message(message, instance):
             def update_on_reaction(reactions):
                 self._logs.update_status(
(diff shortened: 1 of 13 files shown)


philippjfr (approved):

lumen/ai/coordinator.py
```diff
@@ -247,6 +247,7 @@ async def run_demo(event):
                     button_style="outline",
                     on_click=use_suggestion,
                     margin=5,
+                    disabled=state.param.busy
```
philippjfr, line 250: This worries me a little, `state.param.busy` can rapidly toggle on and off which would result in UI flicker.

ahuang11: These are only on the suggestion buttons that disappear after the first message; shouldn't affect much.

ahuang11: Maybe we can implement [debouncing](https://www.geeksforgeeks.org/debouncing-in-javascript/) in param:

```
with button.param.update(disabled=True, debounce=2000)
```

philippjfr: In general that might be nice but it's somewhat difficult. Also that API won't work since it'd clash with a parameter called `debounce`. For now I'd suggest a difference approach here.

ahuang11: Ok I linked it to interface loading.

philippjfr (approved):

PR #785, opened 2024-11-24
Fix schema enums

ahuang11:
1. when limited to the first 100 rows, and the column had missing data in the first 100, the LLM was stating that columns "were placeholders" or it had no available data, which was misleading
image

Instead, it should mention the data was truncated:
image

2. I changed the limit from 100 to 1000; shouldn't seriously affect performance on modern processors(?)
3. The count was placed in the nested dict vs outer dict.

### DIFF
--- a/lumen/ai/utils.py
+++ b/lumen/ai/utils.py
@@ -172,15 +172,20 @@ async def get_schema(
     for field, spec in schema.items():
         if "enum" not in spec:
             continue
-        elif not include_enum:
-            spec.pop("enum")
-        elif "limit" in get_kwargs and len(spec["enum"]) > get_kwargs["limit"]:
-            spec["enum"].append("...")
 
-    if count and include_count:
-        spec["count"] = count
+        limit = get_kwargs.get("limit")
+        if not include_enum:
+            spec.pop("enum")
+        elif limit and len(spec["enum"]) > limit:
+            spec["enum"].append("...")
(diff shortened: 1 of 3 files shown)


ahuang11:
Seems like a bigger problem than I expected; it won't load the data because it assumes the column is null.
image

lumen/ai/utils.py
```diff
@@ -172,15 +172,20 @@ async def get_schema(
     for field, spec in schema.items():
         if "enum" not in spec:
             continue
-        elif not include_enum:
+
+        limit = get_kwargs.get("limit")
+        if not include_enum:
             spec.pop("enum")
-        elif "limit" in get_kwargs and len(spec["enum"]) > get_kwargs["limit"]:
+        elif limit and len(spec["enum"]) > limit:
```
amaloney, line 179: This will error if a user gives a limit as a string as a kwarg. I'm not sure if there is a check for this somewhere else, but we may want to set a default value when we do a `get` from the dictionary.

amaloney (approved):
one nit, but LGTM

lumen/ai/utils.py
ahuang11: Thus far, ai/utils is used internally so passing a `str` shouldn't be much of an issue. We also set a default value above if the user doesn't input it:
```python
        if "limit" not in get_kwargs:
            get_kwargs["limit"] = 100
```

PR #797, opened 2024-11-27
Replace ChatDetailsAgent with AnalystAgent

ahuang11:
I felt like ChatDetailAgent was a bit redundant to ChatAgent, and it was wasteful making two calls (main + topic), so I thought of remaking it into an AnalystAgent which answers after SQLAgent, like an Analyst/Data Scientist who would summarize the results to some non-technical person.

> To provide a recommendation on where to move to avoid earthquakes, we need to analyze earthquake data to identify regions with the lowest seismic activity. The SQLAgent will query the earthquake dataset to find areas with the least number of recorded earthquakes or the lowest magnitudes. After obtaining the results, the AnalystAgent will interpret the data to highlight the safest locations based on the findings.

image

image


lumen/ai/coordinator.py
```diff
@@ -640,19 +640,6 @@ async def _resolve_plan(self, plan, agents, messages) -> tuple[list[ExecutionNod
...
-                    execution_graph.append(
-                        ExecutionNode(
-                            agent=agents['SQLAgent'],
-                            provides=['current_table'],
-                            instruction='Load the table',
-                            title='Loading table',
-                            render_output=False
-                        )
-                    )
-                table_provided = True
```
ahuang11, line 655: I forget the exact reason of adding this, but it was calling SQLAgent twice in a row which isn't quite correct.
```
<AnalystAgent AnalystAgent02022>
<SQLAgent SQLAgent02024>
<SQLAgent SQLAgent02024>
```

I haven't encountered any errors since testing without it.

philippjfr: There's good reason to add it, but it should indeed not add it twice.

philippjfr:
`AnalystAgent` seems too close to `AnalysisAgent` naming wise. Don't have an immediate suggestion for a better name though.

lumen/ai/coordinator.py
```diff
@@ -448,6 +448,7 @@ async def respond(self, messages: list[Message], **kwargs: dict[str, Any]) -> st
                 self.interface.stream(msg, user='Lumen')
                 return msg
             for node in execution_graph:
+                print(node.agent)
```
amaloney, line 451: If this is for debugging purposes, you should remove the print statement. If it is for logging purposes, then you should make a logger.

amaloney (approved):
Check the `print` statement, otherwise this looks good

amaloney:
How about `DataAnalystAgent`?

ahuang11:
To me, AnalysisAgent isn't intuitive. I think it would make more sense to be renamed ToolboxAgent and keeping AnalystAgent.

lumen/ai/coordinator.py
ahuang11, line 451:
```suggestion
```

lumen/ai/coordinator.py
ahuang11: I think this was only necessary because there was a bug in resolving dependencies, but was fixed in: https://github.com/holoviz/lumen/pull/761?

philippjfr: I don't think that's accurate, the whole point of that fix was to move this block to the correct place. Let's undo this and resolve this problem separately.

philippjfr: Okay, inverted the resolution now, which should avoid the problem.

lumen/ai/prompts/AnalystAgent/main.jinja2
```diff
@@ -0,0 +1,17 @@
+{% extends 'Actor/main.jinja2' %}
+
+{% block instructions %}
+You are a world-class data analyst and subject expert. Your role is to analyze the data and present precise, actionable
```
philippjfr, line 4: This prompt results in fairly "business-y" output, which is okay for now but may not be desirable in all cases.

lumen/ai/coordinator.py
philippjfr: It wasn't super easy to trigger the problem that required this change but it was possible. You also always force it by asking it to skip the SQL agent.

PR #798, opened 2024-11-30
Add getting started with Lumen AI

ahuang11:
Adds docs on how to get started with Lumen AI.

Missing pieces about overriding system prompts components, but other than that, I think it's good to go.


philippjfr (approved):

philippjfr:
Merging as is and picking up again separately.

PR #777, opened 2024-11-20
literal workaround

ahuang11:
Since quotations `'"table"."name"'` is unconventional, the LLMs would hallucinate table names, i.e. drop quotations, and it would crash the Pydantic model, raising an invalid error. So, to workaround that, replaces everything except alphanumeric with `_` which the LLM understands/copies well, then remaps it back to the original table name with all the special characters after the LLM chooses the table.

PR #799, opened 2024-12-03
Change quotes to apostrophes in prompt

ahuang11:
An alternative to: https://github.com/holoviz/lumen/pull/777

Since it's mentioned that: "Table names MUST match verbatim including the quotations, apostrophes, periods, or lack thereof.", given that the prompt has quote, and the table name also has quote, e.g. `"some"."table"`, it becomes `""some".table""` which confuses the LLM to drop the inner layer, `"some.table"`. However, with apostrophes it becomes `'"some"."table"'` which may fix the literal issue.

### DIFF
--- a/lumen/ai/prompts/Planner/main.jinja2
+++ b/lumen/ai/prompts/Planner/main.jinja2
@@ -10,7 +10,7 @@ You are team lead and have to make a plan to solve how to address the user query
 {% endblock %}
 
 {% block context %}
-- {% if 'current_table' in memory %}The result of the previous step was the '{{ memory['current_table'] }}' table. Consider carefully if it contains all the information you need and only invoke the SQL agent if some other calculation needs to be performed.{% endif %}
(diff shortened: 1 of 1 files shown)


philippjfr:
Maybe a backtick is best?

lumen/ai/prompts/Planner/main.jinja2
```diff
@@ -10,7 +10,7 @@ You are team lead and have to make a plan to solve how to address the user query
 {% endblock %}
 
 {% block context %}
-- {% if 'current_table' in memory %}The result of the previous step was the "{{ memory['current_table'] }}" table. Consider carefully if it contains all the information you need and only invoke the SQL agent if some other calculation needs to be performed.{% endif %}
+- {% if 'current_table' in memory %}The result of the previous step was the '{{ memory['current_table'] }}' table. Consider carefully if it contains all the information you need and only invoke the SQL agent if some other calculation needs to be performed.{% endif %}
```
philippjfr, line 13:
```suggestion
- {% if 'current_table' in memory %}The result of the previous step was the `{{ memory['current_table'] }}` table. Consider carefully if it contains all the information you need and only invoke the SQL agent if some other calculation needs to be performed.{% endif %}
```

philippjfr (approved):

PR #811, opened 2024-12-05
Fix calling analysis

ahuang11:
The new debugging thing works :)

I noticed that the `unmet_dependencies` was not actually formatted, and there wasn't explicit instructions about what to do with the missing requirement.

Also, `__doc__` was changed to `purpose`

One thing I'm unsure about is whether `Analysis` subclasses should also define `purpose` for consistency as a param, or keep as a docstring.

Also closes https://github.com/holoviz/lumen/issues/812 (selecting all columns if not specified)

### DIFF
--- a/lumen/ai/prompts/Planner/main.jinja2
+++ b/lumen/ai/prompts/Planner/main.jinja2
@@ -33,6 +33,6 @@ Here's the choice of experts and their uses:
 {% endfor %}
 
 {% if unmet_dependencies %}
-Note that a previous plan was unsuccessful because it did not satisfy the following required pieces of information: {unmet_dependencies!r}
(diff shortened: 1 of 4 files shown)


lumen/ai/prompts/Planner/main.jinja2
```diff
@@ -33,6 +33,6 @@ Here's the choice of experts and their uses:
 {% endfor %}
 
 {% if unmet_dependencies %}
-Note that a previous plan was unsuccessful because it did not satisfy the following required pieces of information: {unmet_dependencies!r}
+Note that a previous plan was unsuccessful because it was missing {{ unmet_dependencies }} so you should ensure that {{ unmet_dependencies }} is provided in the next plan!
```
ahuang11, line 36:
```suggestion
Note that a previous plan was unsuccessful because it was missing {{ unmet_dependencies }} so you should ensure that is provided in the next plan!
```

philippjfr (approved):

PR #813, opened 2024-12-06
Add template_overrides and prompts validation

ahuang11:
I initially thought it was a bug with the Planner using default agent instead of the user provided agent instance so I added some logic to prevent adding the same type of default agents.

However, I realized, I formatted it improperly:
```python
agents = [lmai.agents.ChatAgent(template_overrides={"instructions": "Act like the user's meteorologist, and explain jargon in the format of a weather report."})]
```
But there wasn't an error...

Should be nested like:
```python
{"main": {"instructions": "Act like the user's meteorologist, and explain jargon in the format of a weather report."}}
```

So I added some validation functions.

### DIFF
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -14,14 +14,58 @@ from .utils import log_debug, render_template, warn_on_unused_variables
 class Actor(param.Parameterized):
 
     template_overrides = param.Dict(default={}, doc="""
-        Overrides the template's 'instructions', 'context', or 'examples' jinja2 blocks.
+        Overrides the template's 'instructions', 'context', 'embeddings', or 'examples' jinja2 blocks.
         Is a nested dictionary with the prompt name (e.g. main) as the key
         and the block names as the inner keys with the new content as the
         values.""")
 
     prompts = param.Dict(default={}, doc="""
         A dict of the prompt name, like 'main' as key nesting another dict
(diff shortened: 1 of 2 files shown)


lumen/ai/actor.py
```diff
@@ -65,6 +109,7 @@ def _render_prompt(self, prompt_name: str, **context) -> str:
                     f"please ensure overrides contains the {e} key"
                 ) from e
         else:
+            print(overrides, "OVERRIDES")
```
ahuang11, line 112:
```suggestion
```

lumen/ai/actor.py
```diff
@@ -14,14 +14,58 @@
...
+                raise ValueError(
+                    f"Prompt {prompt_name!r} is not a valid prompt name. "
+                    f"Valid prompt names are {valid_prompt_names}."
+                )
+            for block_name in template_override:
+                if block_name not in ["instructions", "context", "embeddings", "examples"]:
+                    raise ValueError(
+                        f"Block {block_name!r} is not a valid block name. "
+                        f"Valid block names are ['instructions', 'context', 'embeddings', 'examples']."
+                    )
```
ahuang11, line 50: If user provides their own prompt, they should be able to use other block names.
```suggestion
```

philippjfr (approved):

philippjfr (approved):

Issue #810, opened 2024-12-05
Drop `table` and keep `pipeline` in requires?


ahuang11:
Is there a case where table isn't provided in a pipeline?

PR #816, opened 2024-12-06
Simplify requires/provides

ahuang11:
At this moment, if I'm not mistaken, pipeline contains all of sql, data table, so I simplified the requires and provides, so that the LLM has less to juggle.

We can re-introduce the other requirements if TransformAgent is ever introduced back.

With this, the Planner has a higher success rate to satisfy all requirements on first try (from my limited testing)

Closes https://github.com/holoviz/lumen/issues/810

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -447,7 +447,7 @@ class SQLAgent(LumenBaseAgent):
         }
     )
 
-    provides = param.List(default=["pipeline"], readonly=True)
+    provides = param.List(default=["table", "sql", "pipeline", "data"], readonly=True)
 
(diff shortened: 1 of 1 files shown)


lumen/ai/agents.py
```diff
@@ -447,7 +447,7 @@ class SQLAgent(LumenBaseAgent):
         }
     )
 
-    provides = param.List(default=["table", "sql", "pipeline", "data"], readonly=True)
```
philippjfr, line 450: I think the requires part is correct, but agents should list what they provide fully. Right now we don't have agents that would require a table but not a pipeline but in theory we could, so the provides list should list everything.

philippjfr (approved):

PR #818, opened 2024-12-06
Cleanup enums

ahuang11:
Main change is truncating each enum to 100 chars because it was overloading the LLM with not too useful info.

image

Also drops `format_schema` because `get_schema` was the only one that called it.

### DIFF
--- a/lumen/ai/utils.py
+++ b/lumen/ai/utils.py
@@ -198,7 +198,7 @@ async def get_schema(
             spec["enum"].append("...")
         elif limit and len(spec["enum"]) == 1 and spec["enum"][0] is None:
             spec["enum"] = [f"(unknown; truncated to {get_kwargs['limit']} rows)"]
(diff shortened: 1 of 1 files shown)


lumen/ai/utils.py
```diff
@@ -195,22 +177,32 @@ async def get_schema(
...
         if not include_enum:
             spec.pop("enum")
+            continue
         elif limit and len(spec["enum"]) > limit:
             spec["enum"].append("...")
         elif limit and len(spec["enum"]) == 1 and spec["enum"][0] is None:
             spec["enum"] = [f"(unknown; truncated to {get_kwargs['limit']} rows)"]
-
-    schema = format_schema(schema)
+        # truncate each enum to 100 rows
```
ahuang11, line 201:
```suggestion
        # truncate each enum to 100 characters
```

philippjfr (approved):

PR #821, opened 2024-12-06
Ability to break while loops

ahuang11:
1. Before [the bug was fixed](https://github.com/holoviz/lumen/pull/811/files#diff-1051df66413da9671c9d8eab752bf7d4ee358831dd458fd3164878a98db21ac1L35-L37), Planner could go in an endless loop of trying to figure out how to meet the unmet_dependencies, so for future bugs that may arise, let the planner break after 5 attempts.
2. Due to the LLM's chaotic nature, SourceAgent may be called when it's unnecessary, but users have no choice besides uploading a dataset to continue--now, they can press cancel.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -715,7 +715,7 @@ class Planner(Coordinator):
                 else:
                     planned = True
                 if attempts > 5:
-                    raise ValueError(f"Could not find a suitable plan for {messages[-1]["content"]!r}")
+                    raise ValueError(f"Could not find a suitable plan for {messages[-1]['content']!r}")
             self._memory['plan'] = plan
             istep.stream('\n\nHere are the steps:\n\n')
             for i, step in enumerate(plan.steps):
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -695,6 +695,7 @@ class Planner(Coordinator):
         unmet_dependencies = set()
         schemas = {}
         execution_graph = []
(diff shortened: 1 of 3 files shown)


lumen/ai/coordinator.py
```diff
@@ -710,8 +711,11 @@ async def _compute_execution_graph(self, messages: list[Message], agents: dict[s
                 execution_graph, unmet_dependencies = await self._resolve_plan(plan, agents, messages)
                 if unmet_dependencies:
                     istep.stream(f"The plan didn't account for {unmet_dependencies!r}", replace=True)
+                    attempts += 1
                 else:
                     planned = True
+                if attempts > 5:
+                    raise ValueError(f"Could not find a suitable plan for {messages[-1]["content"]!r}")
```
philippjfr, line 718: We should probably start making sure that we provide friendly error messages when things go wrong and develop a general "tone" for how the system interacts with the user. 

Does it speak in first person:

> I could not find a suitable plan to answer your question because I could not find a way to determine the correct <table | data | ...>. Please rephrase your question.

is it apologetic:

> Sorry, I couldn’t come up with a good plan to answer your question—I couldn’t figure out the right <table | data | ...>. Could you try rephrasing it for me?

Does it use passive voice:

> A suitable plan to answer your question could not be found because the correct <table | data | ...> could not be determined. Please rephrase your question.

philippjfr: That's not specific to this error, just an observation.

PR #828, opened 2024-12-09
Limit enums to 100

ahuang11:
Was quite surprised to see I exceeded gpt-4o-mini context length, and realized that we were passing potentially 1000 enums per columns.

Then, if there were two consecutive messages, it basically double stacked it.


amaloney (approved):
lgtm

PR #836, opened 2024-12-10
Add default to log_level cli

ahuang11:
Or else it crashes if unspecified.

### DIFF
--- a/lumen/command/ai.py
+++ b/lumen/command/ai.py
@@ -84,7 +84,7 @@ class LumenAIServe(Serve):
         endpoint = args.provider_endpoint
         temperature = args.temperature
         agents = args.agents
-        log_level = args.log_level or "INFO"
+        log_level = args.log_level
 
         if not provider:
             provider = LLMConfig.detect_provider()
@@ -115,8 +115,7 @@ class LumenAIServe(Serve):
         def build_single_handler_applications(
             paths: list[str], argvs: dict[str, list[str]] | None = None
         ) -> dict[str, Application]:
-            handler = AIHandler(
-                paths,
+            kwargs = dict(
                 provider=provider,
                 api_key=api_key,
                 temperature=temperature,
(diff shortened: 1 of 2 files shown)


lumen/command/ai.py
```diff
@@ -84,7 +84,7 @@ def invoke(self, args: argparse.Namespace) -> bool:
         endpoint = args.provider_endpoint
         temperature = args.temperature
         agents = args.agents
-        log_level = args.log_level
+        log_level = args.log_level or "INFO"
```
amaloney, line 87: The `log_level` can be set as a string or `None` in `AIHandler`. I think we should change the type to only accept `str` if we default to `INFO` always.

https://github.com/holoviz/lumen/blob/086858e3d4f53b723d04941c4819e33f51c7506a/lumen/command/ai.py#L141-L155

As a curiosity, which value takes precedence as the default; the one being fed into `AIHandler` via the command line, or the one in the `config`? We set the `log_level` to `INFO` in the config object

https://github.com/holoviz/lumen/blob/086858e3d4f53b723d04941c4819e33f51c7506a/lumen/command/ai.py#L168-L186

and here in the UI

https://github.com/holoviz/lumen/blob/086858e3d4f53b723d04941c4819e33f51c7506a/lumen/ai/ui.py#L72-L73

and here

https://github.com/holoviz/lumen/blob/086858e3d4f53b723d04941c4819e33f51c7506a/lumen/ai/ui.py#L96

amaloney (changes requested):

lumen/command/ai.py
ahuang11: The hierarchy from my understanding is

LumenAIServe.invoke > LumenAIServe.build_single_handler_applications > AIHandler.__init__ > render_template > ExplorerUI

ahuang11:
```python

```

amaloney (approved):

PR #843, opened 2024-12-11
Tweak message logs

ahuang11:
Hides chat reaction icons + copy icon on the help message.

Also properly serializes the messages into strings, so it's not showing objects reprs.

image

However, I'm not sure how to differentiate across Exploration tabs since it's on coordinator and not explorer ui.

Also, not sure if we need to redesign the database schema, to include a separate steps table.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -67,7 +67,7 @@ class Coordinator(Viewer, Actor):
     interface = param.ClassSelector(class_=ChatInterface, doc="""
         The ChatInterface for the Coordinator to interact with.""")
 
-    messages_db_path = param.String(default=None, doc="""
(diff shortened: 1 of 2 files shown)


lumen/ai/coordinator.py
```diff
@@ -67,8 +67,8 @@ class Coordinator(Viewer, Actor):
     interface = param.ClassSelector(class_=ChatInterface, doc="""
         The ChatInterface for the Coordinator to interact with.""")
 
-    logs_filename = param.String(default=None, doc="""
-        Log file to write to.""")
+    messages_db_path = param.String(default=None, doc="""
```
philippjfr, line 70: Don't love `messages_db_path`, how about `log_db_path`?

philippjfr (approved):

PR #844, opened 2024-12-12
Tweak and fixes

ahuang11:
Available tables was renamed to tables
Sources might not be defined prior to UI
Noticed that Planner was consuming a lot of tokens with enums

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -449,7 +449,7 @@ class Coordinator(Viewer, Actor):
             obj = obj.value
         return str(obj)
 
-    def _process_messages(self, messages, max_user_messages=1):
+    def _process_messages(self, messages, max_user_messages=2):
         """
         Process messages to group consecutive messages from the same user.
         Limit the number of user messages to `max_user_messages`.
@@ -460,6 +460,9 @@ class Coordinator(Viewer, Actor):
         for message in messages[::-1]:
             role = message["role"]
             content = message["content"].strip()
+            if user_count == 0 and role == "assistant":
+                # the first message should be user
+                continue
 
             # join the next message to the previous one if they are from the same user
             if role == previous_role and input_messages:
@@ -485,13 +488,13 @@ class Coordinator(Viewer, Actor):
                     await self.llm.get_client("default")  # caches the model for future use
 
             messages = self._process_messages(
-                self.interface.serialize(custom_serializer=self._serialize, limit=8),
-                max_user_messages=2
+                self.interface.serialize(custom_serializer=self._serialize, limit=10),
+                max_user_messages=3
             )
 
             invalidation_assessment = await self._invalidate_memory(messages)
             if invalidation_assessment:
-                messages = mutate_user_message(f"Please be aware: {invalidation_assessment!r}", messages)
+                messages = mutate_user_message(f"Please be aware: {invalidation_assessment!r}", messages[-3:])
 
             agents = {agent.name[:-5]: agent for agent in self.agents}
             execution_graph = await self._compute_execution_graph(messages, agents)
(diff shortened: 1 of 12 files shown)


lumen/ai/tools.py
```diff
@@ -97,7 +97,7 @@ class TableLookup(VectorLookupTool):
     def __init__(self, **params):
         super().__init__(**params)
         self._memory.on_change('sources', self._update_vector_store)
-        self._update_vector_store(None, None, self._memory["sources"])
+        self._update_vector_store(None, None, self._memory.get("sources", [self._memory.get("source")]))
```
philippjfr, line 100: What if 'source' is None?

lumen/ai/llm.py
```diff
@@ -600,4 +604,8 @@ async def invoke(
             if not messages[i]["content"]:
                 messages[i]["content"] = "--"
 
+        if messages and messages[-1]["role"] == "assistant":
+            # cannot end with assistant
```
philippjfr, line 608: Why not? OpenAI doesn't allow for it?

ahuang11: Anthropic doesn't allow for it https://github.com/holoviz/lumen/issues/849

philippjfr: Seems hacky though, remind me in which cases we start with and end with assistant?

Starting should only happen if our arbitrary 5(?) message cutoff for context starts with an assistant message? In that case can't we just drop that message or go one further back in history?

And for the ending message can we just move it before the user message?

ahuang11: I think we arbitrarily insert assistant messages in coordinator:
```

            if isinstance(subagent, SQLAgent):
                custom_agent = next((a for a in self.agents if isinstance(a, AnalysisAgent)), None)
                if custom_agent:
                    custom_analysis_doc = custom_agent.purpose.replace("Available analyses include:\n", "")
                    custom_message = (
                        f"Avoid doing the same analysis as any of these custom analyses: {custom_analysis_doc} "
                        f"Most likely, you'll just need to do a simple SELECT * FROM {{table}};"
                    )
                    custom_messages.append({"role": "user", "content": custom_message})
            if instruction:
                custom_messages.append({"role": "assistant", "content": instruction})
```

```

            if invalidation_assessment:
                messages.append({"role": "assistant", "content": invalidation_assessment + " so do not choose that table."})
                context_length += 1
```

ahuang11: Okay logging it, I realized big issues with how I'm inserting these messages "--"
image

ahuang11: Okay, might add a `system_guidelines`

ahuang11:
Fixed a lot of bugs with:
1. The context from previous steps was not serializing
2. Combining consecutive assistant messages into one

In doing so, I also improved the debugging experience

image

I also combined the extra instructions into user message instead.

lumen/ai/coordinator.py
```diff
@@ -385,38 +389,34 @@ async def _execute_graph_node(self, node: ExecutionNode, messages: list[Message]
                         f"Avoid doing the same analysis as any of these custom analyses: {custom_analysis_doc} "
                         f"Most likely, you'll just need to do a simple SELECT * FROM {{table}};"
                     )
-                    custom_messages.append({"role": "user", "content": custom_message})
-            if instruction:
```
ahuang11, line 391: Dropped this because I already include the full plan
https://github.com/holoviz/lumen/pull/844/commits/6bfbc2a31c6d0af7cba321f3c2b58f7dda58f9a1#diff-d99fd924b197130083f3cc0db28819f92a929aab5aef9f8d17a72b5a9aa7fbf5R683

ahuang11:
Okay this is ready~

philippjfr (approved):

philippjfr (approved):

PR #846, opened 2024-12-12
Add retry icon

ahuang11:
Allows LLMs to focus on editing the output, using a larger model.

https://github.com/user-attachments/assets/be723415-269f-4a16-b154-aa68e2c5ff84


lumen/ai/views.py
```diff
@@ -243,6 +278,10 @@ async def render(self):
             return
 
         pipeline = self.component
+
+        # questionable if this is best way to handle memory
+        # does NOT work if user updates SQL from earlier in the history
+        self._memory["sql"] = self.spec
```
ahuang11, line 284: I realized, if you edit the spec by hand/retry, the latter messages do not reference it because `memory` isn't updated, so this is my work around, but I don't think it's ideal

ahuang11:
Wondering if there should be a RetryAgent instead.

lumen/ai/views.py
philippjfr: Yeah, we need to handle replay. Technically editing for earlier outputs should either be locked or editing should replay all the subsequent steps.

philippjfr:
Looking great!

philippjfr (approved):
Good improvements so far. Not entirely happy about the additional complexity on the Output classes but will see if I can refactor that in a separate PR.

PR #851, opened 2024-12-13
Factor retry handling out of Output into Agents

philippjfr:
The retry handling PR moved a lot of the logic for retries onto the `LumenOutput` classes but overall it'd be much cleaner if that complexity could be handled by the agents themselves. So in this PR I'm moving the retry handling onto the `LumenBaseAgent`.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -448,7 +448,7 @@ class SQLAgent(LumenBaseAgent):
         }
     )
 
-    provides = param.List(default=["table", "sql", "pipeline", "data", "base_schema"], readonly=True)
+    provides = param.List(default=["table", "sql", "pipeline", "data", "tables_sql_schemas"], readonly=True)
 
     requires = param.List(default=["source"], readonly=True)
 
@@ -693,9 +693,6 @@ class SQLAgent(LumenBaseAgent):
 
         # include min max for more context for data cleaning
         schema = await get_schema(source, table, include_min_max=True)
-        if not source.ephemeral:
-            # store the schema of the original table
-            self._memory["base_schema"] = schema
         join_required = await self._check_requires_joins(messages, schema, table)
         if join_required is None:
             return None
@@ -704,7 +701,7 @@ class SQLAgent(LumenBaseAgent):
         else:
             tables_to_source = {table: source}
 
-        table_schemas = {}
+        tables_sql_schemas = {}
         for source_table, source in tables_to_source.items():
             if source_table == table:
                 table_schema = schema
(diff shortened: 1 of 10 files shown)


ahuang11:
Added alternating warning. Will now fix two user messages in a row.
image

ahuang11:
Dropped user query from the system since I think it's redundant now. Also mutated the original message.

image

Testing vega specs now.

lumen/ai/agents.py
```diff
@@ -375,6 +340,82 @@ async def respond(
...
+        async def retry_invoke(event: param.parameterized.Event):
+            reason = event.new
+            modified_messages = mutate_user_message(
+                f"New feedback: {reason!r}.\n\nThese were the previous instructions to use as reference:",
+                deepcopy(messages), wrap='\n"""\n', suffix=False
+            )
+            with self.param.update(memory=memory):
+                system = await self._render_prompt(
+                    "retry_output", messages=modified_messages, spec=out.spec,
+                    language=out.language
```
philippjfr, line 385: Seems like you removed the language from the template again, if that was intentional remove it here too.

ahuang11: I think you might not have pushed your changes on language(?) I don't see it in https://github.com/holoviz/lumen/pull/851/commits/da41924e4f24930d6cc71a21be2bedbd45bad5ac

ahuang11:
I realized that it can tunnel from the lack of columns context and not get back out:
```
Here's a summary of the dataset the user just asked about; you must reference columns from this dataset in your response:
```

ahuang11:
Besides the tunneling issue (I might loop through all the sources tables & specs), it seems to work well!

https://github.com/user-attachments/assets/adbb9bab-e1b0-4cf3-9fc1-fcf50feebc9f


https://github.com/user-attachments/assets/e8c9228d-9467-4650-9cbd-81c0c8524b81

ahuang11:
Okay I think it's ready

"""

The user is unsatisfied with the output; your goal is to update the specification in sql:

Here is the original specification to be corrected:
```sql
SELECT *
FROM read_csv('olympic_hosts.csv')
WHERE "game_season" = 'Winter'
ORDER BY "game_year" ASC
LIMIT 5
```

Here are YAML schemas for currently relevant tables:

- `read_csv('olympic_hosts.csv')`:
```yaml
game_end_date:
  format: datetime
  max: '2022-02-20T12:00:00'
  min: '1896-04-15T11:39:39'
  type: str
game_location:
  enum:
  - China
  - Japan
  - Republic of Korea
  - Brazil
  - Russian Federation
  - '...'
  type: str
game_name:
  enum:
  - Beijing 2022
  - Tokyo 2020
  - PyeongChang 2018
  - Rio 2016
  - Sochi 2014
  - '...'
  type: str
game_season:
  enum:
  - Winter
  - Summer
  type: str
game_slug:
  enum:
  - beijing-2022
  - tokyo-2020
  - pyeongchang-2018
  - rio-2016
  - sochi-2014
  - '...'
  type: str
game_start_date:
  format: datetime
  max: '2022-02-04T15:00:00'
  min: '1896-04-06T11:38:39'
  type: str
game_year:
  max: 2022
  min: 1896
  type: int

```
Respond in the same format as the original specification. The query may include multiple tasks, your job is merely to correct the specification, ignore any instructions unrelated to that and do not add any comments or code examples.

"""

ahuang11 (approved):

PR #854, opened 2024-12-13
Table names bug

### LINKED ISSUE #853: Table names using URLs
#### Description of expected behavior and the observed behavior

- *Expected:* Lumen AI performs data description when using a remote file.
- *Observed:* Lumen AI errors in the `ChatAgent` indicating a `KeyError`

#### Complete, minimal, self-contained example code that reproduces the issue

```bash
lumen-ai serve "https://datasets.holoviz.org/penguins/v1/penguins.csv" --provider llama --show
```

#### Screenshots or screencasts of the bug in action

Image

- [x] I may be interested in making a pull request to address this

amaloney:
This commit refactors the `respond` method in `ai/tools.py` to allow for table names that may contain a URL.

Resolves #853

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -28,7 +28,7 @@ from ..views import (
     Panel, VegaLiteView, View, hvPlotUIView,
 )
 from .actor import Actor, ContextProvider
-from .config import PROMPTS_DIR
+from .config import PROMPTS_DIR, SOURCE_TABLE_SEPARATOR
 from .controls import SourceControls
 from .llm import Llm, Message
 from .memory import _Memory
@@ -513,9 +513,9 @@ class SQLAgent(LumenBaseAgent):
                     sql_query = sql_query.replace(a_table, renamed_table)
                 else:
                     renamed_table = a_table
-                if "//" in renamed_table:
+                if SOURCE_TABLE_SEPARATOR in renamed_table:
                     # Remove source prefixes from table names
-                    renamed_table = re.sub(r"//[^/]+//", "", renamed_table)
+                    renamed_table = re.sub(r".*?<->", "", renamed_table)
                 sql_query = sql_query.replace(a_table, renamed_table)
                 mirrors[renamed_table] = (a_source, a_table)
             source = DuckDBSource(uri=":memory:", mirrors=mirrors)
@@ -586,7 +586,7 @@ class SQLAgent(LumenBaseAgent):
         multi_source = len(self._memory['sources']) > 1
         if multi_source:
             tables = [
-                f"//{a_source}//{a_table}" for a_source in self._memory["sources"]
+                f"{SOURCE_TABLE_SEPARATOR}{a_source}{SOURCE_TABLE_SEPARATOR}{a_table}" for a_source in self._memory["sources"]
                 for a_table in a_source.get_tables()
             ]
         else:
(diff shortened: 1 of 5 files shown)


amaloney:
The refactor allows for table names to be URLs

image

image

This allows for a user to run the following at the command line.

```bash
lumen-ai serve "https://datasets.holoviz.org/penguins/v1/penguins.csv" --provider llama --show
```

lumen/ai/tools.py
```diff
@@ -113,10 +113,16 @@ def _update_vector_store(self, _, __, sources):
...
+        closest_tables = []
+        results = self.vector_store.query(
+            messages[-1]["content"],
+            top_k=self.n,
+            threshold=self.min_similarity,
+        )
+        for result in results:
+            table_name = result["text"].split("//", 1)[1]
+            closest_tables.append(table_name)
+        self._memory["closest_tables"] = closest_tables
```
ahuang11, line 125: Are these the same; just written in a more readable form? I can't seem to spot the change, and not sure how this would fix the issue. Can you elaborate a bit on what's changed? Thanks!

```python
        closest_tables = []
        results = self.vector_store.query(
            messages[-1]["content"],
            top_k=self.n,
            threshold=self.min_similarity,
        )
        for result in results:
            table_name = result["text"].split("//", 1)[1]
            closest_tables.append(table_name)
        self._memory["closest_tables"] = closest_tables
```

vs
```

        self._memory["closest_tables"] = closest_tables = [
            result["text"].split("//")[1] for result in
            self.vector_store.query(messages[-1]["content"], top_k=self.n, threshold=self.min_similarity)
        ]
```

amaloney: They are close to being the same, except for the `split(...)` part. The original code split the string on all occurrences of `//` characters. The refactored bit splits on the **first** `//`, leaving the rest of the string alone.

In essence the original code would take the string

```
Source1223//https://datasets.holoviz.org/penguins/v1/penguins.csv
```

and split on all `//` and _only_ select the first item in the list. So the string would turn in to

```
['Source1223', 'http:', 'datasets.holoviz.org/penguins/v1/penguins.csv']
```

and select `http:` as the name of the table, which is incorrect if you supply data via the command similar to the following.

```
lumen-ai serve "https://datasets.holoviz.org/penguins/v1/penguins.csv" --provider llama --show
```

The name of the table in Lumen becomes `https://datasets.holoviz.org/penguins/v1/penguins.csv`, which has the `//` character in it that the `respond` method splits on.

As I was debugging I had to break the list comprehension down, which is where the change came from. If you don't like it, I can revert back since the absolutely minimal possible change would be the following.

```python
self._memory["closest_tables"] = closest_tables = [
    result["text"].split("//", 1)[1] for result in
    self.vector_store.query(messages[-1]["content"], top_k=self.n, threshold=self.min_similarity)
]
```

Your call, but I should refactor after the `if not closest_tables:` check doing the same string manipulation in this PR as well.

ahuang11: Oh thanks for the clarification.

No worries, we can keep yours; it's more readable.

I'm thinking of whether we need a new delimiter, like ` : `.

amaloney:
Refactored and changed the source and table separator to `<->`.

ahuang11:
I think you need to find and replace all across the entire repo.
image

lumen/ai/agents.py
```diff
@@ -695,8 +695,8 @@ async def respond(
             table=table,
         )
         if join_required:
-            # Remove source prefixes message, e.g. //<source>//<table>
-            messages[-1]["content"] = re.sub(r"//[^/]+//", "", messages[-1]["content"])
+            # Remove source prefixes message, e.g. SOURCE_TABLE_SEPARATOR<source>SOURCE_TABLE_SEPARATOR<table>
+            messages[-1]["content"] = re.sub(r".*?<->", "", messages[-1]["content"])
```
ahuang11, line 699: Maybe this can be 
`r".*?" + SOURCE_TABLE_SEPARATOR` (if it works)

amaloney: good catch

amaloney: `rf".*?{SOURCE_TABLE_SEPARATOR}"` works

ahuang11 (approved):
Thanks! I think just need to resolve conflict

ahuang11:
Resolved the merge conflict. Will auto merge once tests pass.

PR #866, opened 2024-12-13
Custom data source docs

amaloney:


doc/how_to/ai_config/custom_data_source.md
```diff
@@ -0,0 +1,42 @@
...
+supports the following LLM providers.
+
+- OpenAI
+- Anthropic
+- MistralAI
+- AzureOpenAI
+- AzureMistralAI
+- Llama
+
+Ensure your API key is in the environment of the terminal you run your commands in.
```
philippjfr, line 23: I know this is early but just as a heads up, each how-to will be pretty standalone. You should assume that a user has configured the LLM already, either by reading the getting started or the how-to on configuring the LLM.

amaloney:
@ahuang11 rebased and updated with suggestions and an app example.

philippjfr (changes requested):
Thanks, a good start! I'd like to add a section about configuring a custom data source, e.g. an example like this:

```python
from lumen.sources.snowflake import SnowflakeSource

import lumen.ai as lmai

source = SnowflakeSource(
    account='...',
    authenticator='externalbrowser',
    database='...',
    user='...',
)

lmai.ExplorerUI(source).servable()
```

ahuang11:
I would also add if the filename has an unexpected extension, like `.parq` I don't think duckdb is automatically able to infer it, so you may need `read_parquet("file.parq")`

amaloney:
@ahuang11 are you suggesting that a user will need to do something like the following?

```python
import lumen.ai as lmai
import pandas as pd

lmai.ExploreUI(
    data=[pd.read_parquet("path/to/weird/extension.parket")],
    llm=lmai.llm.Llama(),
    agents=[lmai.agents.SQLAgent],
)
```

ahuang11:
Nah, to clarify, duckdb supports this https://duckdb.org/docs/data/parquet/overview.html natively.

So we need to just pass the directive:
`read_parquet("file.parq")`

doc/how_to/ai_config/custom_data_source.md
```diff
@@ -0,0 +1,127 @@
...
+source = SnowflakeSource(
+    account="...",
+    authenticator="externalbrowser",
+    database="...",
+    user="...",
+)
+lmai.ExplorerUI(source).servable()
+```
+
+### DuckDb
```
ahuang11, line 84: Following their branding:
```suggestion
### DuckDB
```

Also, I think it'd be nice to mention/show it supports more than just parq, like csv.

doc/how_to/ai_config/custom_data_source.md
```diff
@@ -0,0 +1,127 @@
...
+---
+This guide shows you how to configure custom data sources for Lumen AI.
+```
+
+## Overview
+
+We will be using a local LLM to understand how to load custom data sources in Lumen. You do not need
+to use a local LLM, and can instead opt for using one you have an API key for, ensure your API key
+is in the environment of the terminal you run your commands in.
+
```
ahuang11, line 15: Maybe you can also mention directly uploading in the UI

ahuang11 (approved):

doc/how_to/ai_config/custom_data_source.md
```diff
@@ -0,0 +1,141 @@
...
+# Inline catalog
+catalog = {
+    "penguins": LocalCatalogEntry(name="penguins", description="Palmer penguins", args=...),
+    "earthquakes": LocalCatalogEntry(name="earthquakes"),
+}
+# URI to a catalog
+#catalog = "s3://bucket/data/cat*.yaml"
+lmai.ExplorerUI(catalog=catalog).servable
+```
+
```
philippjfr, line 128: I'll drop this, we don't support `catalog` like this and I think we're de-emphasizing Intake going forward.

```suggestion
```

philippjfr (approved):
Looks great, thank you!

PR #868, opened 2024-12-14
Disable past specs and minor tweaks

ahuang11:
1. Hides the retry icon + disable codeeditor
image

2. enum had a bug where if it was numeric, it would crash
3. there was uneven spacing between each icon so I added a spacer.

### DIFF
--- a/lumen/ai/views.py
+++ b/lumen/ai/views.py
@@ -40,6 +40,8 @@ class LumenOutput(Viewer):
 
     title = param.String(allow_None=True)
 
+    disabled = param.Boolean(default=False)
+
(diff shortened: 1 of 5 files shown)


lumen/ai/views.py
```diff
@@ -40,6 +40,8 @@ class LumenOutput(Viewer):
 
     title = param.String(allow_None=True)
 
+    disabled = param.Boolean(default=False)
```
philippjfr, line 43: Could start keeping parameters in alphabetical order, unless there's some overriding reason to group them in some other way.

Issue #871, opened 2024-12-16
Replace with Markitdown


ahuang11:
https://github.com/microsoft/markitdown/tree/main

PR #872, opened 2024-12-16
Add markitdown and tweak

ahuang11:
Closes https://github.com/holoviz/lumen/issues/871

1. uses markitdown
2. fixes file uploads were persisting after "add"
3. allows upload tabs to be closable, in case a doc was accidentally uploaded
4. tweaks system prompts so Planner/ChatAgent knows about the docs and not have source agent trigger

image
image


amaloney (approved):
lgtm

PR #875, opened 2024-12-16
Add Custom Tools docs and default to an LLM

ahuang11:


amaloney (approved):
lgtm

ahuang11:
Added a section about integrating other tools, like from LangChain, into Lumen AI, but I needed it to default to an LLM--else, user would have to additionally define a LLM, like

```python
from langchain_community.tools import DuckDuckGoSearchRun

llm = lmai.OpenAI()
search = DuckDuckGoSearchRun()
tool = lmai.tools.FunctionTool(function=search.invoke, purpose=search.description, llm=llm)
ui = lmai.ExplorerUI(tools=[tool])
```

vs just:

```python
from langchain_community.tools import DuckDuckGoSearchRun

search = DuckDuckGoSearchRun()
tool = lmai.tools.FunctionTool(function=search.invoke, purpose=search.description)
ui = lmai.ExplorerUI(tools=[tool])
```

I think the latter is better because users may forget to pass an LLM and it would crash.

philippjfr (approved):

PR #877, opened 2024-12-17
Add custom agents docs

ahuang11:
Questionable to use those internal methods like `_render_prompt` and `_stream` and also `_memory` property, but I don't think there's another way, unless straying away from the pre-defined structure.

Maybe we should make those public?


philippjfr (approved):

philippjfr:
>Questionable to use those internal methods like _render_prompt and _stream and also _memory property, but I don't think there's another way, unless straying away from the pre-defined structure.

Why is it questionable? Using private methods on self is perfectly fine.

Issue #873, opened 2024-12-16
RAG should consider filename


ahuang11:
I have two completely different docs:

1. receipt
2. org chart

When I asked about the receipt, it found text from org chart, rather than looking at the receipt.

PR #883, opened 2024-12-18
Query by filename

ahuang11:
Closes https://github.com/holoviz/lumen/issues/873

For context, I uploaded two completely separate documents:

1. a receipt
2. an org chart

When I asked `what's in the receipt`, it referenced contents from the org chart instead of the receipt because the receipt doesn't mention it's a receipt in its contents.

This PR adds some weights to the relevancy based on filename

### DIFF
--- a/lumen/ai/controls.py
+++ b/lumen/ai/controls.py
@@ -282,7 +282,10 @@ class SourceControls(Viewer):
             file, file_extension=document_controls.extension
         ).text_content
 
-        metadata = f"Filename: {document_controls.filename} " + document_controls._metadata_input.value
+        metadata = {
+            "filename": document_controls.filename,
+            "provided": document_controls._metadata_input.value
+        }
         document = {"text": text, "metadata": metadata}
         if "document_sources" in self._memory:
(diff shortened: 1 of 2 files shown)


amaloney (approved):
lgtm

philippjfr:
Would like some discussion around the decision to insert the filename as a separate kind of entry vs. simply adding it as a metadata field (which seems conceptually cleaner to me).

philippjfr:
I guess the obvious reason is that the metadata is currently only used for filtering and the embeddings are only computed for the embeddings? In which case I guess my question is whether we should automatically include the embeddings in the metadata somehow.

ahuang11:
If it's added only to the metadata, I believe embeddings can't query it, they can only perform an exact match case filter vs entry, it can do fuzzy search.

ahuang11:
> In which case I guess my question is whether we should automatically include the embeddings in the metadata somehow.

I suppose we could have another database column containing a joined text of everything.

ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/911

PR #896, opened 2024-12-23
Add code agent

ahuang11:
Since LLMs excel at coding and Panel can almost wrap anything, I think it's good to add an arbitrary code executor.

Not sure how to control the formatting though. In the first view spec, the code is properly spaced with new lines. The second view spec is condensed. Ideally, it'd use the `|` yaml syntax for multi line strings, but not sure how to specify that.

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -19,6 +19,8 @@ from panel.viewable import Viewable, Viewer
 from pydantic import BaseModel, create_model
 from pydantic.fields import FieldInfo
 
+from lumen.views.base import ExecPythonView
+
 from ..base import Component
 from ..dashboard import Config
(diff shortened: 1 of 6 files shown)


philippjfr:
Seems very scary to me, what if a user asks the LLM to write code that destroys the system or exfiltrate secrets?

ahuang11:
Yes I thought of that which is why it's not part of the default agents, but I suppose it could add a disclaimer to use at the user's own risk. It could be useful for contained environments.

philippjfr:
Would be nice if this could somehow be run in a sandboxed environment, e.g. pyodide or some other WASM based Python environment.

ahuang11:
Stale

PR #907, opened 2024-12-27
Streamline show table

ahuang11:
Makes finding the user message and removing the need to determine `join` a bit more robust.

Before:
image

After:
image

No longer `startswith` because the planner now mutates the user message with additional instructions.

`3 (u). "Show the table: 'passengers.csv'" -- Here's the current instructions from the multi-step plan: "Generate and execute a query to retrieve all data from the 'passengers.csv' table."`

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -710,7 +710,8 @@ class SQLAgent(LumenBaseAgent):
         tables_to_source = {table: source}
         if join_required is None:
             join_required = await self._check_requires_joins(messages, schema, table)
-            if join_required is None:  # TODO: what is this for? can we remove?
+            if join_required is None:
+                # Bail if query was cancelled or errored out
                 return None
             if join_required:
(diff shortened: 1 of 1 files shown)


lumen/ai/agents.py
```diff
@@ -690,19 +700,20 @@ async def respond(
...
+
+        tables_to_source = {table: source}
         if join_required is None:
-            return None
-        if join_required:
-            tables_to_source = await self.find_join_tables(messages)
-        else:
-            tables_to_source = {table: source}
+            join_required = await self._check_requires_joins(messages, schema, table)
+            if join_required is None:  # TODO: what is this for? can we remove?
```
philippjfr, line 713: I believe it occurs if the `check_requires_joins` is cancelled or errors out.

philippjfr (approved):

philippjfr (approved):

PR #911, opened 2025-01-02
Query by metadata

ahuang11:
Supersedes https://github.com/holoviz/lumen/pull/883 based on:

> I guess the obvious reason is that the metadata is currently only used for filtering and the embeddings are only computed for the contents? In which case I guess my question is whether we should automatically include the embeddings in the metadata somehow.

> I suppose we could have another database column containing a joined text of everything, and have a toggle to query_metadata = True/False

```
Please augment your response with the following context if relevant:
- Receipt Invoice number 1234 Receipt number Date paid 1234 December 15, 2024 Payment method American Express - 1234 OpenAI, LLC 548 Market Street PMB 97273 San Francisco, California 1234 United States ar@openai.com Bill to Me Address United State $22.07 paid on 2024 ... (Relevance: 0.2 - Metadata: {'text': 'Filename: Receipt-2813-2096 '}
```


lumen/ai/tools.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, **params):
     def _update_vector_store(self, _, __, sources):
         for source in sources:
             if not self.vector_store.query(source["text"], threshold=1):
-                self.vector_store.add([{"text": source["text"], "metadata": source.get("metadata", "")}])
+                self.vector_store.add([{"text": source["text"], "metadata": {"text": source.get("metadata", "")}}])
```
ahuang11, line 71:
```suggestion
                self.vector_store.add([{"text": source["text"], "metadata": {"comments": source.get("metadata", "")}}])
```

philippjfr:
Sorry I wasn't clear before. I wasn't saying I prefer this approach, I just wanted to weigh the pros and cons of both approaches.

philippjfr:
Well that isn't quite true, I definitely prefer treating the filename as metadata rather than having a distinct row for it, but as we are discovering it does have drawbacks.

ahuang11:
I personally prefer this approach.

Perhaps it could be more targeted with `query_with_metadata = ["<key>", "<key2>"]` so like `query_with_metadata = ["filename"]`

ahuang11:
Alternatively, could go back to the old approach, but store metadata as a separate table, then joined on the main table.

ahuang11:
Looking at ChromaDB, they have explicit filters for metadata filtering
https://docs.trychroma.com/docs/querying-collections/metadata-filtering

philippjfr:
My main question is whether it's really necessary to store a text and metadata column and compute the additional embeddings. Does always including the metadata in the embedding result in any appreciable performance degredation?

ahuang11:
> Does always including the metadata in the embedding result in any appreciable performance degredation?

Do you mean rather than `text | metadata | text_and_metadata`, simply keep `text_and_metadata | metadata`, and then do postprocessing on `text_and_metadata` on retrieval, e.g. split by some delimiter like metadata tags `||||>`

lumen/ai/vector_store.py
```diff
@@ -27,6 +27,65 @@ class VectorStore(param.Parameterized):
...
+            metadata: A dictionary of metadata associated with the text.
+
+        Returns:
+            A tuple containing the content text and the formatted metadata text.
+        """
+        metadata_items = [
+            f"({key}: {self._format_metadata_value(value)})"
+            for key, value in metadata.items()
+        ]
+        text_and_metadata = " ".join(metadata_items)
```
philippjfr, line 59: So it's just metadata, not `text_and_metadata`?

lumen/ai/vector_store.py
```diff
@@ -378,34 +434,45 @@ def add(self, items: list[dict]) -> list[int]:
...
-            chunks = self._chunk_text(text)
+            content_text, text_and_metadata = self._get_content_and_text_and_metadata(
+                text, metadata
+            )
 
-            for chunk in chunks:
+            content_chunks = self._chunk_text(content_text)
+            for chunk in content_chunks:
                 all_texts.append(chunk)
+                all_text_and_metadatas.append(text_and_metadata)
```
philippjfr, line 450: Seems like we'd be computing embeddings for the same thing a bunch of times?

ahuang11: Yeah, so maybe: https://github.com/holoviz/lumen/pull/918

ahuang11:
Seems to work decently now!

image

ahuang11:
Okay this is now ready.

I also added functionality where if user uploads the same file (perhaps with modified contents), it'll overwrite the existing document.

philippjfr (approved):

philippjfr:
Looks good!

PR #918, opened 2025-01-03
LLM decides metadata

ahuang11:
Builds upon https://github.com/holoviz/lumen/pull/917 and maybe supersedes https://github.com/holoviz/lumen/pull/911

This implementation simply lets the LLM decide what files are the most relevant.

I also tried to do a hybrid content + metadata table ranking in https://github.com/holoviz/lumen/compare/main...meta_weight but it didn't work well and was complex...

image
image

image

image

### DIFF
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -13,6 +13,7 @@ from .actor import Actor, ContextProvider
 from .config import PROMPTS_DIR, SOURCE_TABLE_SEPARATOR
 from .embeddings import NumpyEmbeddings
 from .llm import Message
+from .models import RelevantFiles
 from .translate import function_to_model
 from .vector_store import NumpyVectorStore, VectorStore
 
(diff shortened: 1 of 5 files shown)


lumen/ai/tools.py
```diff
@@ -74,7 +83,17 @@ def _update_vector_store(self, _, __, sources):
...
+        relevant_files_model = self._lookup_prompt_key("main", "response_model")
+        system = await self._render_prompt("main", messages=messages, filenames=filenames)
+        result = await self.llm.invoke(messages, system=system, response_model=relevant_files_model)
+        if result.files:
+            filters = {"filename": result.files}
+            threshold = 0
+        else:
+            filters = None
+            threshold = self.min_similarity
+        results = self.vector_store.query(query, top_k=self.n, threshold=threshold, filters=filters)
```
ahuang11, line 96: Maybe call it twice; once with the file filters and once without

PR #921, opened 2025-01-07
Fix loading indicator on explorations and file uploads

ahuang11:
`exploration=Column(loading=True)` would go on forever because render_output would update the wrong tab because the pre-computed index was outdated. Since it's such a negligible computation, I decided to just re-compute it right before rendering output.

Also, previously, unsupported file uploads would crash the UI, but now it catches the exception gracefully.

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -567,6 +567,7 @@ class ExplorerUI(UI):
 
             async def render_output(_, old, new):
                 added = [out for out in new if out not in old]
+                index = self._explorations.active if len(self._explorations) else -1
                 exploration = self._explorations[index]
                 self._add_outputs(exploration, added, local_memory)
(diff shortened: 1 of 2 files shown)


philippjfr (approved):
Thanks for tracking this down!

lumen/ai/ui.py
```diff
@@ -567,6 +567,7 @@ def sync_available_sources_memory(_, __, sources):
 
             async def render_output(_, old, new):
                 added = [out for out in new if out not in old]
+                index = self._explorations.active if len(self._explorations) else -1
```
philippjfr, line 570: Actually I do not believe this to be correct. Seems like it'll break if you switch tabs while an exploration is running.

philippjfr: Fixed.

PR #922, opened 2025-01-07
Add intro messages

ahuang11:
Provides a little more guidance on what the tabs are. Feel free to update the language as you see fit

image

image


lumen/ai/ui.py
```diff
@@ -48,6 +48,23 @@
...
+- Download the chat into a reproducible notebook to share with others!
+- Check out [Using Lumen AI](https://holoviz-dev.github.io/lumen/lumen_ai/getting_started/using_lumen_ai.html) for more tips & tricks!
+"""
+
+EXPLORATIONS_INTRO = """
+🧪 **Explorations**
+
+- Explorations are queries of your original data.
+- They consist of interactive tables and visualizations.
+- Each exploration is saved as a new tab to allow for easy comparison.
```
philippjfr, line 65:
```suggestion
🧪 **Explorations**

- Explorations are analyses applied to one or more datasets.
- An exploration consists of interactive tables and visualizations.
- New explorations are launched in a new tab for each SQL query result.
- Each exploration maintains its own context, so you can branch off an existing result by navigating to that tab.
```

philippjfr:
>specifically I couldn't get visible=self._explorations.param.objects.rx.bool() working

Shouldn't it be `self._explorations.param.objects.rx.bool().rx.not_()`, i.e. make it visible when there are no explorations?

ahuang11:
Yeah, but I get this error:
```
  File "/Users/ahuang/repos/lumen/lumen/ai/ui.py", line 500, in _table_explorer
    visible=self._coordinator.interface.param.objects.rx.bool().rx.not_()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'rx'
```

philippjfr:
Ah okay, it should be:

```
self._coordinator.interface.param['objects'].rx.bool().rx.not_()
```

The param accessor works like a dataframe column accessors, if the parameter name clashes with an existing method the method takes precedence.

ahuang11:
Okay was able to use rx.

Also added an issue
https://github.com/holoviz/param/issues/1005

philippjfr (approved):

PR #965, opened 2025-01-25
Verify that LLM is working on load

philippjfr:
One of the first things people might encounter is that they haven't configured the LLM correctly. This is the one case where it's potentially useful to surface tracebacks to the user, so we make a call to the LLM on load and then either tell the user loading the LLM was successful or show them the error.

Screenshot 2025-01-25 at 21 04 38

Screenshot 2025-01-25 at 21 05 22


ahuang11:
This broke the suggestion buttons.

image

PR #980, opened 2025-01-28
Better sql error handling

ahuang11:
The error was marked as Python and had poor newlines formatting, which might have made the SQL retry not as good; this PR fixes that and addresses when the SQL query returns 0 rows.

Also pops the spinner on fail.

Clipboard-20250128-034851-395

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -526,7 +526,7 @@ class SQLAgent(LumenBaseAgent):
             errors = '\n'.join(errors)
             system += (
                 f"\n\nYour last query did not work as intended:\n```sql\n{last_query}\n```\n\n"
-                f"Your priority is to expertly revise these errors:\n\n```sql\n{errors}\n```\n\n"
+                f"Your priority is to expertly revise these errors:\n\n```\n{errors}\n```\n\n"
                 f"If the error is `syntax error at or near \")\"`, double check you used "
                 f"table names verbatim, i.e. `read_parquet('table_name.parq')` instead of `table_name`."
             )
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -525,8 +525,8 @@ class SQLAgent(LumenBaseAgent):
                 last_query = sql_code_match.group(1)
             errors = '\n'.join(errors)
             system += (
-                f"\n\nYour last query:\n```sql\n{last_query}\n```\ndid not work as intended, so your task is to "
-                f"expertly revise these errors:\n```python\n{errors}\n```\n\n"
+                f"\n\nYour last query did not work as intended:\n```sql\n{last_query}\n```\n\n"
+                f"Your priority is to expertly revise these errors:\n\n```sql\n{errors}\n```\n\n"
                 f"If the error is `syntax error at or near \")\"`, double check you used "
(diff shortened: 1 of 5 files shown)


lumen/ai/agents.py
```diff
@@ -581,6 +582,8 @@ async def _create_valid_sql(
             pipeline = await get_pipeline(
                 source=sql_expr_source, table=expr_slug, sql_transforms=sql_transforms
             )
+            if len(pipeline.data) == 0:
+                raise ValueError("No data was returned from the SQL query.")
```
philippjfr, line 586: This seems like a condition that can actually happen and be a perfectly valid result. I'll remove and then you can try to convince me why this is needed later.

```suggestion
```

lumen/ai/agents.py
```diff
@@ -751,9 +754,12 @@ async def respond(
...
             # Remove source prefixes message, e.g. SOURCE_TABLE_SEPARATOR<source>SOURCE_TABLE_SEPARATOR<table>
             messages[-1]["content"] = re.sub(rf".*?{SOURCE_TABLE_SEPARATOR}", "", messages[-1]["content"])
-        sql_query = await self._create_valid_sql(messages, system_prompt, tables_to_source, step_title)
-        pipeline = self._memory['pipeline']
-
+        try:
+            sql_query = await self._create_valid_sql(messages, system_prompt, tables_to_source, step_title)
+            pipeline = self._memory['pipeline']
+        except RetriesExceededError:
+            self._memory["outputs"] = self._memory['outputs'] + ["__error__"]
```
philippjfr, line 761: I would much rather have a separate sentinel than write to outputs.

lumen/ai/agents.py
```diff
@@ -525,11 +525,12 @@ async def _create_valid_sql(
                 last_query = sql_code_match.group(1)
             errors = '\n'.join(errors)
             system += (
-                f"Your last query:\n```sql\n{last_query}\n```\ndid not work as intended, so your task is to "
-                f"expertly revise these errors:\n```python\n{errors}\n```\n\n"
+                f"\n\nYour last query did not work as intended:\n```sql\n{last_query}\n```\n\n"
+                f"Your priority is to expertly revise these errors:\n\n```sql\n{errors}\n```\n\n"
```
philippjfr, line 529: I'm confused, errors are the exceptions no? They're neither Python nor SQL.

lumen/ai/ui.py
```diff
@@ -577,6 +577,11 @@ def _add_outputs(self, exploration: Column, outputs: list[LumenOutput], memory:
             else:
                 exploration.insert(0, sql_pane)
 
+        if isinstance(outputs[-1], str) and outputs[-1] == "__error__":
+            outputs.pop()
+            self._explorations.pop(-1)
```
philippjfr, line 582: Did you test this? It breaks things in like 10 different ways.

philippjfr: 1. If we already generated some outputs and then errored it removes those outputs
2. If we were adding to an existing exploration this removes the old exploration
3. This pops the last exploration but we may be adding onto an existing one.
4. This only pops the explorations, there's also the conversations, titles, and contexts that need to be modified, which means everything will be out of sync after this.

philippjfr (approved):

PR #983, opened 2025-01-28
Separate context lookups in Planner

philippjfr:
Due to limitations in the way `instructor` works we cannot currently stream the planning stage. This reduces the interactive feel of running an exploration and makes the user wait on output. To avoid this we now separate the context lookup stage of the planning stage and the actual planning. In other words we first have the `Planner` request tables that it will need and any tools it wants to call to provide context and only when those steps have been resolved do we start the planning.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -711,28 +711,35 @@ class Planner(Coordinator):
             table_info=table_info,
             tools=list(tools.values()),
         )
-        context = await self.llm.invoke(
-            messages=messages,
-            system=system,
-            model_spec=model_spec,
-            response_model=context_model,
-            max_retries=3,
-        )
-        if getattr(context, 'tables', None):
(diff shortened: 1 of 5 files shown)


philippjfr:
This also makes things like this possible:

```python

async def user_clarify(topic_to_clarify: str):
    """
    Allows asking user to clarify, only use this if the question is completely incomprehensible, not for minor clarifications.
    """
    widget = pn.widgets.TextInput()
    confirm = pn.widgets.Button(name='Ok')
    ui.interface.send(pn.Column(topic_to_clarify, pn.Row(widget, confirm)), respond=False, user='Tool')
    while not confirm.clicks:
        await asyncio.sleep(0.5)
    return f'The user clarified that by {topic_to_clarify} they meant "{widget.value}"'

ui = lmai.ui.ExplorerUI(..., tools=[user_clarify]).servable()
```

Screenshot 2025-01-28 at 16 03 18

PR #992, opened 2025-01-28
Fix retry

ahuang11:
1. Mutate user messages rather than system for higher priority
2. Add "but as the expert, you may need to deviate from it" because it was strictly following the instructions from the plan
3. Bump temperature to 0.25 because it was still not changing its answers

image

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -844,9 +844,10 @@ class BaseViewAgent(LumenBaseAgent):
         if errors:
             errors = '\n'.join(errors)
             if self._last_output:
-                system += (
-                    f"\nNote, your last specification did not work as intended:\n```json\n{self._last_output}\n```\n\n\n"
-                    f"Your task is to expertly revise these errors:\n```\n{errors}\n```\n"
+                messages = mutate_user_message(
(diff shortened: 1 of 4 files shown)


lumen/ai/agents.py
```diff
@@ -566,18 +566,25 @@ async def _create_valid_sql(
...
-                f"If the error is `syntax error at or near \")\"`, double check you used "
+            else:
+                return
+            num_errors = len(errors)
+            errors = ('\n'.join(f"{i+1}. {error}" for i, error in enumerate(errors))).strip()
+            content = (
+                f"\n\nYou are a world-class SQL user. Identify why this query failed:\n```sql\n{last_query}\n```\n\n"
+                f"Your goal is to try a different query to address the question while avoiding these issues:\n```\n{errors}\n```\n\n"
+                f"Please build upon your previous thought: {chain_of_thought!r}, but note, a penalty of $100 will be incurred "
+                f"for every time the issue occurs, and thus far you have been penalized ${num_errors * 100}! "
```
philippjfr, line 582: Does this kind of stuff work? I've read conflicting things.

ahuang11: In my testing, on the second try, all I can conclude is that it changes its query strategy.

philippjfr (approved):

PR #993, opened 2025-01-28
Tweak planner prompt to not redo SQL

ahuang11:
image


ahuang11:
https://github.com/holoviz/lumen/pull/994

PR #994, opened 2025-01-28
Remove check validity

ahuang11:
With planner having the SQL context and schemas, I don't think this is actually needed anymore, and makes Lumen feel so much more faster! And internally less complex :)

1. I ask about some Nino tables
2. I change topics to Olympics
3. I subset Olympic hosts

image
image
image


philippjfr (approved):
I like it!

PR #997, opened 2025-01-29
Add validation and re-attempts to Coordinator

philippjfr:
Optional validation of results/answers and re-attempts. For now the default is not to validate.


ahuang11:
Already implemented ValidationAgent.

PR #1010, opened 2025-01-30
Refactor SQL agent

ahuang11:
The goal of this PR is to simplify SQLAgent and make it run faster by reducing the number of steps to output SQL.

Changes:
1. Merged `select_table`, `require_joins`, `find_joins` into `find_tables`.
2. Changed the separator because `<->` was really hard to look at, compared to `:::`
3. Reduced duplication in Planner (was showing double of the same schemas)
4. `make_tables_model` also highlights issues when joining

image


lumen/ai/agents.py
```diff
@@ -656,131 +598,70 @@ async def _create_valid_sql(
...
-        tables_to_source = {table: source}
-        if join_required is None:
-            join_required = await self._check_requires_joins(messages, table, schema, tables_schema_str)
-            if join_required is None:
-                # Bail if query was cancelled or errored out
-                return None
-            if join_required:
-                tables_to_source = await self.find_join_tables(messages)
+        tables_to_source, tables_schema_str = await gather_table_sources(sources, include_sep=True)
+        tables_to_source, comments = await self._find_tables(messages, tables_schema_str)
```
philippjfr, line 658: I'm finding this can return no tables at all and then errors below.

philippjfr:
I think this is a nice improvement in principle but I've also found issues and am not confident I can track them all down.

lumen/ai/agents.py
ahuang11: Maybe we can wrap `retry_llm_output`?

ahuang11:
Maybe test more and merge after announcement/release then?

ahuang11:
image

philippjfr (approved):

PR #1011, opened 2025-01-30
Remove error on missing API_KEY

ahuang11:
I think the UI will handle it gracefully now.


philippjfr:
I prefer early errors, launching into the UI and then finding out that you have to re-configure is more annoying than erroring out early. I did hide the traceback by default though: https://github.com/holoviz/lumen/pull/1014

PR #1018, opened 2025-01-30
Fix export issues

ahuang11:
image

### DIFF
--- a/lumen/ai/export.py
+++ b/lumen/ai/export.py
@@ -8,6 +8,7 @@ import nbformat
 
 from panel import Column
 from panel.chat import ChatMessage, ChatStep
+from panel.pane import SVG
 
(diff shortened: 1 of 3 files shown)


lumen/ai/export.py
```diff
@@ -67,7 +68,7 @@ def format_output(msg: ChatMessage):
 def render_cells(messages: list[ChatMessage]) -> tuple[Any, list[str]]:
     cells, extensions = [], []
     for msg in messages:
-        if msg.user == 'Help':
+        if msg.user == 'Help' or (isinstance(msg.avatar, SVG) and msg.user == " "):
```
philippjfr, line 71: Isn't checking `msg.user == " "` enough and then separately we should actually handle Image avatars?

philippjfr (approved):

PR #1019, opened 2025-01-30
Add AI gallery item

ahuang11:
Strangely, I can't seem to make it show up in the gallery
image


philippjfr:
Don't really see the point of a gallery item that doesn't lead anywhere. Will close and move the gallery under the spec.

PR #1022, opened 2025-01-31
Remove semi colon on all transforms

ahuang11:
I think at some point, I encountered some Lumen AI syntax error related to `;` or comments

```
WITH gold_medals AS (
    SELECT
        "athlete_full_name",
        COUNT(*) AS gold_count
    FROM read_csv('olympic_medals.csv')
    WHERE "medal_type" = 'GOLD'
      AND "slug_game" LIKE '%-2022'  -- Assuming Winter Olympics are identified by the year
    GROUP BY "athlete_full_name"
)
SELECT
    "athlete_full_name",
    gold_count
FROM gold_medals
ORDER BY gold_count DESC
LIMIT 1
```

Not sure if this will fix, but I think it should be applied to all transforms anyways.

Ideally, these transforms would be applied to the generated SQL for visibility, and not after under the hood in pipeline (okay applied)

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -628,16 +628,17 @@ class SQLAgent(LumenBaseAgent):
         else:
             source = next(iter(sources))
 
+        sql_transforms = [SQLLimit(limit=1_000_000)]
+        for transform in sql_transforms:
+            sql_query = transform.apply(sql_query)
+
         # check whether the SQL query is valid
         expr_slug = output.expr_slug
         try:
             sql_expr_source = source.create_sql_expr_source({expr_slug: sql_query})
             # Get validated query
             sql_query = sql_expr_source.tables[expr_slug]
-            sql_transforms = [SQLLimit(limit=1_000_000)]
-            pipeline = await get_pipeline(
-                source=sql_expr_source, table=expr_slug, sql_transforms=sql_transforms
-            )
+            pipeline = await get_pipeline(source=sql_expr_source, table=expr_slug)
         except Exception as e:
             report_error(e, step)
             raise e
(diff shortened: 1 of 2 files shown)


lumen/ai/agents.py
```diff
@@ -628,16 +628,17 @@ async def _create_valid_sql(
         else:
             source = next(iter(sources))
 
+        sql_transforms = [SQLLimit(limit=1_000_000)]
+        for transform in sql_transforms:
+            sql_query = transform.apply(sql_query)
```
philippjfr, line 633: Confused, why iterate over a loop when there is only a single item? Also why bake the transform into the query?

ahuang11: Was just keeping the original structure, in case there were more transforms. I'm fine with dropping the loop.

Applying it on the original query so that it shows up in the SQL code editor, in case there's a bug in the SQLTransform output.

philippjfr: Baking it into the query means it falls outside of the programmable interface, e.g. if we want to toggle the limit on and off. So I'm -1 on that part.

ahuang11: On the flip side, if the transform (limit) isn't visible to the user, the user might think it's an issue with the data (unaware there's a limit to 1 million rows; the full data is pretty tiny and I didn't notice it initially because I was looking at the query):

image

> if we want to toggle the limit on and off
I don't follow; can't it still be toggled on and off (e.g. `SELECT * FROM read_parquet('num.parquet') AS num_table LIMIT 1000000`, turns into `SELECT * FROM read_parquet('num.parquet') AS num_table`?

ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/1085

PR #1026, opened 2025-01-31
Tweak Analyst and Planner

ahuang11:
Minor changes to show the plan length, make Planner use joins for multiple queries, and make analyst agent suspicious of data more often because I think the value in AnalystAgent is someone who questions the results, providing unthought of insights, rather than regurgitating what the data table shows.

image

Who won the most gold medals during the winter olympics
image

### DIFF
--- a/lumen/ai/prompts/AnalystAgent/main.jinja2
+++ b/lumen/ai/prompts/AnalystAgent/main.jinja2
@@ -3,7 +3,7 @@
 {%- block instructions %}
 You are a world-class data analyst and subject expert. Your role is to analyze the data and present precise, actionable insights in a way that is clear and understandable to someone without significant technical knowledge, such as a CEO. Focus on distilling complex findings into straightforward, meaningful takeaways while avoiding technical jargon. Explain trends, relationships, and/or key details in the data with clarity, emphasizing their business relevance and impact, but make sure to leverage your broader perspective to call out any results that appear odd. Avoid general overviews; instead, highlight specific patterns, relationships between columns, and/or actionable insights.
 
-- Do no take the result at face value, be suspicious and  a devil's advocate, questioning the results, e.g. if the results show a 100% increase in sales, ask yourself if that is realistic.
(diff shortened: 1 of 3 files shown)


lumen/ai/prompts/AnalystAgent/main.jinja2
```diff
@@ -3,6 +3,7 @@
 {%- block instructions %}
 You are a world-class data analyst and subject expert. Your role is to analyze the data and present precise, actionable insights in a way that is clear and understandable to someone without significant technical knowledge, such as a CEO. Focus on distilling complex findings into straightforward, meaningful takeaways while avoiding technical jargon. Explain trends, relationships, and/or key details in the data with clarity, emphasizing their business relevance and impact, but make sure to leverage your broader perspective to call out any results that appear odd. Avoid general overviews; instead, highlight specific patterns, relationships between columns, and/or actionable insights.
 
+- Try not to always take the result at face value, be a devil's advocate by questioning the results, e.g. if the results show a 100% increase in sales, ask yourself if that is realistic, and critique the SQL that was used to generate the results.
```
philippjfr, line 6: Seems like this should be baked into the main description above rather than a separate bullet point. In fact it's kind of duplicative of this:

>but make sure to leverage your broader perspective to call out any results that appear odd

Think we can condense it a little too:

>  You are a world-class data analyst known for your analytical rigor and critical thinking. Your task is to analyze the data and deliver precise, actionable insights in clear, non-technical language.

> - Distill complexity into straightforward, meaningful takeaways.
> - Explain trends, relationships, and key details with a focus on business relevance and impact.
> - Identify and question anomalies—highlighting potential outliers or unexpected results.
> - Avoid generic summaries—instead, emphasize specific patterns, column interactions, and actionable insights.

> Your goal is to make complex findings easy to understand while maintaining depth and precision.

ahuang11: I added back a bit:
`- Do not always take results at face value; identify and question anomalies—highlighting potential outliers or unexpected results, addressing the SQL if necessary.`

image

philippjfr (approved):

PR #1035, opened 2025-02-03
Add chat loc

ahuang11:
image

### DIFF
--- a/lumen/ai/components.py
+++ b/lumen/ai/components.py
@@ -33,8 +33,8 @@ class SplitJS(JSComponent):
 
     left = Child()
     right = Child()
-    sizes = param.NumericTuple(default=[60, 40], length=2)
-    min_sizes = param.NumericTuple(default=[300, 200], length=2)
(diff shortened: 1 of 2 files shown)


lumen/ai/ui.py
```diff
@@ -347,6 +347,9 @@ class ExplorerUI(UI):
 
     title = param.String(default='Lumen Explorer', doc="Title of the app.")
 
+    chat_loc = param.Selector(default='left', objects=['left', 'right'], doc="""
```
philippjfr, line 350: Would use a more descriptive name here. Also PLEASE keep parameter definitions sorted unless there's a compelling reason not to.

lumen/ai/components.py
```diff
@@ -31,6 +33,8 @@ class SplitJS(JSComponent):
 
     left = Child()
     right = Child()
+    sizes = param.List(default=[60, 40])
```
philippjfr, line 36: Would use `NumericTuple` with length 2.

philippjfr (approved):

PR #1036, opened 2025-02-04
Add llama_url in cli

ahuang11:
Was playing around with local models and experienced how tedious it was to change the local model, so I added support for:

```
lumen-ai serve --llama-url https://huggingface.co/unsloth/Mistral-Small-24B-Instruct-2501-GGUF/blob/main/Mistral-Small-24B-Instruct-2501-Q4_K_M.gguf?chat_format=mistral-instruct&n_ctx=1028
```
 
URL's query_params are used as model_kwargs

### DIFF
--- a/doc/lumen_ai/how_to/llm/llama_cpp.md
+++ b/doc/lumen_ai/how_to/llm/llama_cpp.md
@@ -74,12 +74,12 @@ lumen-ai serve --provider llama --model-kwargs '{
 }'
 ```
 
-Providing these arguments via the CLI can be cumbersome. Instead, paste the quantized model file's `llama-url` and pass `model_kwargs` as query parameters. If `llama-url` is set, `provider` automatically defaults to `llama`, and will error if another `provider` is set.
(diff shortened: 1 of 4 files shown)


philippjfr (approved):
Would like to pick a better name for the `--llama-url` parameter because it has nothing at all to do with Llama. I already strongly dislike the fact that the `Llama` LLM provider isn't called `LlamaCpp`, but here the case is even stronger since LlamaCpp is only the execution engine and we could eventually switch to a completely different engine. Something like `--llm-model-url`?

ahuang11:
I was thinking `huggingface-file-url` originally since it only supports huggingface ones right now but for consistency and conciseness I ended with `llama-url`...

However, open to `llm-model-url` or `llm-file-url.

>  already strongly dislike the fact that the Llama LLM provider isn't called LlamaCpp

Yeah not sure why it's not LlamaCpp. Maybe we can change it?

philippjfr:
Sure, I'm in favor on both.

PR #1041, opened 2025-02-05
Move tabs to the right

ahuang11:
image

Without this PR, the tabs labels' is awkwardly placed in the center of the screen


philippjfr:
Perfect, was thinking the same!

philippjfr (approved):

PR #1047, opened 2025-02-06
Simplify source/table separator logic

philippjfr:
The logic for source table disambiguation was a little overcomplicated. This PR replaces the previous separator (`:::`) with `__@__` which isn't valid SQL syntax and even if it was has a tiny chance of collision.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -638,10 +638,12 @@ class SQLAgent(LumenBaseAgent):
                 f"{a_source}{SOURCE_TABLE_SEPARATOR}{a_table}" for a_source in sources.values()
                 for a_table in a_source.get_tables()
             ]
+            sep = SOURCE_TABLE_SEPARATOR
         else:
             tables = self._memory["sources"][0].get_tables()
+            sep = None
         system = await self._render_prompt(
-            "find_tables", messages, separator=SOURCE_TABLE_SEPARATOR, tables_schema_str=tables_schema_str
+            "find_tables", messages, separator=sep, tables_schema_str=tables_schema_str
         )
         tables_model = self._get_model("find_tables", tables=tables)
         model_spec = self.prompts["find_tables"].get("llm_spec", "default")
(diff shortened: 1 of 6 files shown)


ahuang11:
Also gather_table_sources.

```

async def gather_table_sources(sources: list[Source], include_provided: bool = True, include_sep: bool = False) -> tuple[dict[str, Source], str]:
    """
    Get a dictionary of tables to their respective sources
    and a markdown string of the tables and their schemas.

    Parameters
    ----------
    sources : list[Source]
        A list of sources to gather tables from.
    include_provided : bool
        Whether to include the provided source in the string; will always be included in the dictionary.
    include_sep : bool
        Whether to include the source separator in the string.
    """
    tables_to_source = {}
    tables_schema_str = ""
    for source in sources:
            tables_to_source[table] = source
            if source.name == PROVIDED_SOURCE_NAME and not include_provided:
                continue
            label = f"{SOURCE_TABLE_SEPARATOR}{source}{SOURCE_TABLE_SEPARATOR}{table}" if include_sep else table
            if isinstance(source, DuckDBSource) and source.ephemeral or "Provided" in source.name:
                sql = source.get_sql_expr(table)
                schema = await get_schema(source, table, include_enum=True, limit=3)
                tables_schema_str += f"- {label}\nSchema:\n```yaml\n{yaml.dump(schema)}```\nSQL:\n```sql\n{sql}\n```\n\n"
            else:
                tables_schema_str += f"- {label}\n\n"
    return tables_to_source, tables_schema_str.strip()

```

lumen/ai/agents.py
```diff
@@ -578,9 +578,9 @@ async def _create_valid_sql(
                     sql_query = sql_query.replace(a_table, renamed_table)
                 else:
                     renamed_table = a_table
+                # Remove source prefixes from table names
                 if SOURCE_TABLE_SEPARATOR in renamed_table:
-                    # Remove source prefixes from table names
-                    renamed_table = re.sub(rf".*?{SOURCE_TABLE_SEPARATOR}", "", renamed_table)
+                    _, renamed_table = renamed_table.split(SOURCE_TABLE_SEPARATOR)
```
ahuang11, line 583: For maximum stability, I think we should do maxsplit=1.

philippjfr: Sure.

PR #1049, opened 2025-02-06
Save some tokens by disabling URLs in errors

ahuang11:
Was trying to figure out a way to handle retries, and discovered this util
https://python.useinstructor.com/concepts/reask_validation/#optimizing-token-usage

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -16,6 +16,7 @@ import param
 
 from instructor.dsl.partial import Partial
 from instructor.patch import Mode, patch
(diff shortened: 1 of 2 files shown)


lumen/ai/llm.py
```diff
@@ -34,6 +35,8 @@ class Message(TypedDict):
 
 BASE_MODES = list(Mode)
 
+disable_pydantic_error_url()
```
philippjfr, line 38: Probably should go in `lumen.ai.config`?

PR #1050, opened 2025-02-06
Fix validation error

ahuang11:
System prompt had the table listed, but not the literal.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -650,7 +650,6 @@ class Planner(Coordinator):
             title="Obtaining additional context...",
             user="Assistant"
         ) as istep:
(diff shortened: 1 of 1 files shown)


lumen/ai/coordinator.py
```diff
@@ -651,6 +650,7 @@ async def _lookup_context(
             title="Obtaining additional context...",
             user="Assistant"
         ) as istep:
+            print(messages, system)
```
philippjfr, line 653:
```suggestion
```

philippjfr (commented):
Oh I get the point that it's confusing the model since it does see that those tables exist and might list them again.

philippjfr (approved):

PR #1053, opened 2025-02-06
move RuntimeError if no provider is given

amaloney:
Moving the `RuntimeError` closer to when the `provider` is defined from the `args` object. I found this error message made a lot more sense than the one found in the try-block when/if I tried to run `lumen-ai` on the command line with no provider set.

### DIFF
--- a/lumen/command/ai.py
+++ b/lumen/command/ai.py
@@ -96,6 +96,16 @@ class LumenAIServe(Serve):
     def invoke(self, args: argparse.Namespace) -> bool:
         """Override invoke to handle both sets of arguments"""
         provider = args.provider
+        if provider is None:
+            raise RuntimeError(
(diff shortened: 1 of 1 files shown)


ahuang11:
Thanks for pointing that out.

The reason why it's below is because of `detect_provider()`.

```python
        if llm_model_url and provider and provider != "llama":
            raise ValueError(
                f"Cannot specify both --llm-model-url and --provider {provider!r}. "
                f"Use --llm-model-url to load a model from HuggingFace."
            )
        elif llm_model_url:
            provider = "llama"
        elif not provider:
            provider = LLMConfig.detect_provider()
```

I think you can just replace this:
```
                f"Could not find LLM Provider {provider!r}, valid providers include: {list(LLM_PROVIDERS)}."
            )
```

ahuang11 (approved):

PR #1062, opened 2025-02-18
Add BigQuerySource

### LINKED ISSUE #1051: SQLAlchemy source for Lumen AI
Include `BigQuery` as a source for Lumen AI.

amaloney:
resolves #1051 

Adds the following source objects

- SQLAlchemy
- BigQuery

### DIFF
--- a/lumen/tests/transforms/test_sql.py
+++ b/lumen/tests/transforms/test_sql.py
@@ -294,7 +294,7 @@ def test_sql_sample_tablesample_percent():
 def test_sql_sample_tablesample_size():
     """Test size-based sampling with TABLESAMPLE (PostgreSQL, etc.)"""
     result = SQLSample.apply_to("SELECT * FROM TABLE", size=100, write="postgres")
-    expected = "SELECT * FROM (SELECT * FROM TABLE) AS subquery TABLESAMPLE (100)"
+    expected = "SELECT * FROM (SELECT * FROM TABLE) AS subquery ORDER BY RANDOM() LIMIT 100"
     assert result == expected
 
 


lumen/sources/sqlalchemy.py
```diff
@@ -0,0 +1,47 @@
...
+
+from sqlalchemy.engine.create import create_engine
+from sqlalchemy.engine.url import URL
+from sqlalchemy.sql.expression import text
+
+from lumen.sources.base import BaseSQLSource
+
+
+class SQLAlchemySource(BaseSQLSource):
+    driver = param.String(default=None, doc="SQL driver.")
```
ahuang11, line 11: I think it'd be wise to have a way to input URL, perhaps as a classmethod `from_url`

philippjfr: +1

philippjfr (commented):
Thanks for the PR, looks like a great start. I do have a few initial comments:

At the highest level I'm not quite seeing how this correctly implements the `BaseSQLSource` APIs, i.e. at minimum I would have expected an implementation of

- `execute`: This is meant for executing a SQL query and returning the result as a DataFrame, your `run_query` method seems to come close but you have to wrap it in `pd.DataFrame.from_records` I'm guessing.
- `get_tables`: This should return a list of valid tables.

Additionally I would expect a somewhat similar API to other sources, where you can define a `tables` parameter that accepts a list of tables (as to limit which tables you can access), or a dictionary mapping from table name alias to a SQL expression.

I appreciate the documentation is a little sparse, so please don't hesitate to reach out to myself or Andrew to clarify things about the API.

I think there's also a misunderstanding of the role of `get_schema`. The additional schema information you are getting from BigQuery is quite helpful but I'd consider that part metadata. The schema in Lumen specifically refers to a dictionary that contains the type of the column and it's min-max and unique values.

amaloney:
This is very helpful information. From the discussion I need to implement the following.

- [ ] add a `from_url` method to the `SQLAlchemy` class
- [ ] implement the methods required from the `BaseSQLSource` class (listed below). I definitely overlooked these requirements when focusing on the BigQuery class, which only required the immediate URL implementation.
  - [ ] `get_sql_expr`
  - [ ] `create_sql_expr_source`
  - [ ] `execute -> pd.DataFrame`
  - [ ] `get_tables -> list[str]` I don't see this in the base class, but I'll add it so others don't make the same mistake as myself
  - [ ] add a `tables = param.ClassSelector(...)`
- [ ] refactor the `get_schema` method to be `get_metadata` and follow the `get_schema` method more closely found in e.g. `DuckDB` or `IntakeBaseSource`

For documentation purposes I think I will also update the `Lumen AI -> How to Guides -> Custom Data Sources` to include a more in-depth discussion on the methods a dev needs to write in order to follow the API specs found in the base class.

- [ ] Update `Lumen AI -> How to Guides -> Custom Data Sources` for a better DX.
- [ ] Add docstrings to methods I personally missed, to ensure they are not overlooked by others in the future.

Any other items y'all can think of will be added to the above to-do lists.

ahuang11:
One other thing are tests (terribly lacking in the ai/ directory, but should be maintained for source/).

Not sure how hard it is to set up a mysql server, but this seems useful for testing locally (probably not CI?) https://docs.getwren.ai/oss/getting_started/sample_data/hr

philippjfr:
@amaloney if you have review comments please still leave them, but I have to build on top of this and we can tidy this up further before release.

amaloney:
Everything has been encapsulated nicely, and I cannot see any issues with how you refactored things, nor did I find any when I tested locally. Thank you for the help

PR #1063, opened 2025-02-18
Indent within step to surface VegaLite validation error

ahuang11:
Wasn't able to trigger this again, but I'm fairly confident this fixes the green 

image


philippjfr (approved):

philippjfr (approved):
Yep, thanks, the change makes sense.

PR #1072, opened 2025-02-21
Propagate tool context

ahuang11:
While testing tools, I noticed that the Planner saw the tool output:  

> *The user requested information about Kern County from Wikipedia. Since this data was already retrieved, I can summarize key details without another lookup.*  

However, the subsequent agent did not:  

> *I cannot look up external sources like Wikipedia directly...*  

To address this, I included the tool output in the user message.

image

The downside is that if a tool is meant only for planning, it wastes tokens downstream. Should tools have a `coordinator_only` flag? Open to feedback.

---

## Examples:

Before:
image

After:
image


amaloney (approved):
nice

philippjfr (approved):

PR #1074, opened 2025-02-25
Add Source.get_metadata

ahuang11:
Just an initial set up to explore how `get_metadata` could be implemented and used in Lumen AI.

Right now, it's only used in TableLookup, but I think it can be used more widely, like in ChatAgent / SQLAgent, or maybe just Planner?

I imagine it can be used similarly like `DocumentLookup`

I was using this for testing since these were the only tables that were not sensitive and had descriptions...

```python
from lumen.sources.snowflake import SnowflakeSource

source = SnowflakeSource(
    account=account
    authenticator="externalbrowser",
    database="SNOWFLAKE_SAMPLE_DATA",
    user=user
)

tables_to_get = {
    "APPLICABLE_ROLES": "APPLICABLE_ROLES",
    "COLUMNS": "COLUMNS",
    "DATABASES": "DATABASES"
}

all_metadata = source.get_metadata(tables_to_get)

all_metadata
```
image

### DIFF
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -394,13 +394,10 @@ class TableLookup(VectorLookupTool):
     async def _fetch_and_store_metadata(self, source, table_name: str):
         """Fetch metadata for a table and store it in the vector store."""
         async with self._semaphore:
-            if self.batched:
-                source_metadata = self._raw_metadata[source.name]
-                if isinstance(source_metadata, asyncio.Task):
-                    source_metadata = await source_metadata
-                    self._raw_metadata[source.name] = source_metadata
-            else:
-                source_metadata = await asyncio.to_thread(source.get_metadata, table_name, batched=False)
+            source_metadata = self._raw_metadata[source.name]
+            if isinstance(source_metadata, asyncio.Task):
+                source_metadata = await source_metadata
+                self._raw_metadata[source.name] = source_metadata
 
         table_name_key = next((
             key for key in (table_name.upper(), table_name.lower(), table_name)
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -126,7 +126,7 @@ class TableLookup(VectorLookupTool):
         """Fetch metadata for a table and store it in the vector store."""
         async with self._semaphore:
             try:
-                metadata = await asyncio.to_thread(source.get_table_metadata, table_name)
+                metadata = await asyncio.to_thread(source.get_metadata, table_name)
                 source_table = f'{source.name}{SOURCE_TABLE_SEPARATOR}{table_name}'
                 self._table_metadata[source_table] = metadata
 
@@ -136,10 +136,10 @@ class TableLookup(VectorLookupTool):
 
                 if columns := metadata.get('columns', {}):
                     enriched_text += "\nColumns:"
-                    for col_name, col_desc in columns.items():
+                    for col_name, col_info in columns.items():
                         col_text = f"\n- {col_name}"
-                        if col_desc:
-                            col_text += f": {col_desc}"
+                        if 'description' in col_info:
+                            col_text += f": {col_info['description']}"
                         enriched_text += col_text
 
                 if not self.vector_store.query(source_table, threshold=1):
@@ -156,13 +156,12 @@ class TableLookup(VectorLookupTool):
                 if not self.vector_store.query(source_table, threshold=1):
                     self.vector_store.add([{"text": source_table}])
 
-                if hasattr(source, 'get_table_metadata'):
-                    task = asyncio.create_task(
-                        self._fetch_and_store_metadata(source, table_name)
-                    )
-                    self._metadata_tasks.add(task)
-                    task.add_done_callback(lambda t: self._metadata_tasks.discard(t))
-                    # TODO: Add ability to update existing source table?
+                task = asyncio.create_task(
+                    self._fetch_and_store_metadata(source, table_name)
+                )
+                self._metadata_tasks.add(task)
+                task.add_done_callback(lambda t: self._metadata_tasks.discard(t))
+                # TODO: Add ability to update existing source table?
 
     async def respond(self, messages: list[Message], **kwargs: dict[str, Any]) -> str:
         closest_tables = []
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -330,7 +330,7 @@ class TableLookup(VectorLookupTool):
     requires = param.List(default=["sources"], readonly=True, doc="""
         List of context that this Tool requires to be run.""")
 
-    provides = param.List(default=["closest_tables"], readonly=True, doc="""
+    provides = param.List(default=["closest_tables", "table_similarities"], readonly=True, doc="""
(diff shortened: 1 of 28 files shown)


lumen/sources/snowflake.py
```diff
@@ -117,3 +117,87 @@ def get(self, table, **query):
...
+        columns_info = self.execute(column_query)
+
+        for _, row in columns_info.iterrows():
+            column_name = row['COLUMN_NAME']
+            description = row['COMMENT']
+            result["columns"][column_name] = description if description else ""
+
+        return result
+
+    def get_metadata(self, tables: dict[str, str] | None = None) -> dict[str, dict]:
```
philippjfr, line 177: I know it's somewhat overloaded but I'd follow the pattern we used for `.get` and `.get_schema` and have one method that accepts a `table` argument, which, when undefined, returns data for all tables.

Minor comments:
- Type declaration seems wrong and docstring is wrong format.

philippjfr (commented):
Looks like a great start to me 🚀

ahuang11:
I've been thinking about this approach and believe we should implement an iterative table column and description discovery process, likely within the _lookup_context function to gather necessary context.

The process would work as follows:

1. Use embeddings to perform cosine similarity search across all table names
2. While we currently implement step 1, it seems ineffective for poorly named tables. We should enhance this by adding weighted cosine similarity scores for column names (Though with potentially 1000+ tables, we should evaluate the computational cost of reading all column names)
3. From the top 10 results (weighted by combined table and column name relevance), present all tables and their column names to the LLM for selection
4. Have the LLM select approximately 3 tables, then request their complete schemas (including column values, enumerations, minimums, maximums) and get_metadata descriptions
5. If the LLM is satisfied with the context, proceed to the agent steps; otherwise, restart the process with awareness of previously examined tables

I should watch this for more inspiration: https://www.youtube.com/watch?v=n5EQL9FpCyU

lumen/ai/tools.py
```diff
@@ -104,45 +104,100 @@ class TableLookup(VectorLookupTool):
     provides = param.List(default=["closest_tables"], readonly=True, doc="""
         List of context values this Tool provides to current working memory.""")
 
+    include_metadata = param.Boolean(default=True, doc="""
```
ahuang11, line 107: I think the doc should be improved here. Does this include column descriptions too?

lumen/ai/tools.py
```diff
@@ -104,45 +104,100 @@ class TableLookup(VectorLookupTool):
...
+        Whether to include table descriptions in the embeddings and responses.""")
+
+    include_columns = param.Boolean(default=False, doc="""
+        Whether to include column descriptions in the embeddings.""")
```
ahuang11, line 111: Include column names and descriptions maybe?

lumen/ai/tools.py
```diff
@@ -104,45 +104,100 @@ class TableLookup(VectorLookupTool):
...
+        async with self._semaphore:
+            metadata = await asyncio.to_thread(source.get_metadata, table)
+        multi_source = len(self._memory.get("sources", [])) > 1
+        if multi_source:
+            enriched_text = table_name = f"{source.name}{SOURCE_TABLE_SEPARATOR}{table}"
+        else:
+            enriched_text = table_name = table
+        self._table_metadata[table_name] = metadata
+
+        vector_metadata = {"table": table, "source": source.name}
```
ahuang11, line 139: What is `table` here? Should it be table_name Maybe add a comment if this is intentional.

Also, semi confusing it's called table_name some places, and source_table elsewhere.

lumen/ai/tools.py
```diff
@@ -104,45 +104,100 @@ class TableLookup(VectorLookupTool):
...
             for table in source.get_tables():
-                source_table = f'{source.name}{SOURCE_TABLE_SEPARATOR}{table}'
-                if not self.vector_store.query(source_table, threshold=1):
-                    self.vector_store.add([{"text": source_table}])
+                metadata = {"table": table, "source": source.name}
+                if not self.include_metadata:
+                    if self.vector_store.filter_by(metadata):
+                        # TODO: Add ability to update existing source table?
+                        continue
+                    source_table = f"{source.name}{SOURCE_TABLE_SEPARATOR}{table}"
```
ahuang11, line 162: Maybe this should be renamed to table_name, or better yet, let's officially differentiate "table_name" as "table_slug" (so we know it's attached to source)

lumen/ai/tools.py
```diff
@@ -104,45 +104,100 @@ class TableLookup(VectorLookupTool):
...
+        """
+        query = messages[-1]["content"]
+        results = self.vector_store.query(query, top_k=self.n)
+        multi_source = len(self._memory.get("sources", []))
+        any_matches = any(result['similarity'] >= self.min_similarity for result in results)
+        closest_tables, descriptions = [], []
         for result in results:
-            if SOURCE_TABLE_SEPARATOR not in result["text"]:
-                closest_tables.append(result["text"])
+            if result['similarity'] < self.min_similarity and any_matches:
```
ahuang11, line 182: I have a hard time figuring out why any_matches is appended, rather than prepended.

philippjfr: Doesn't matter much, but sure I'll move it first.

ahuang11:
Hmm, so I tried getting metadata for each table individually, and it was really slow, 4 minutes for 121 tables. Then I realized, we have all the information to do it in one execute query, and that took only 9 seconds. 
Screenshot 2025-03-07 at 6 33 22 PM

I also added other metadata, for use with tool

Screenshot 2025-03-07 at 6 34 13 PM

ahuang11:
Please review https://github.com/holoviz/lumen/pull/1098 and https://github.com/holoviz/lumen/pull/1097 first.

lumen/sources/duckdb.py
```diff
@@ -319,36 +319,34 @@ def get(self, table, **query):
             df = Filter.apply_to(df, conditions=conditions)
         return df
 
-    def _get_table_metadata(self, table: str | list[str], batched: bool = False) -> dict[str, Any]:
+    def _get_table_metadata(self, table: str | list[str]) -> dict[str, Any]:
```
philippjfr, line 322: It's always a list now so don't have to handle the str case.

ahuang11:
I've added a way to bypass the LLM iterative search process to make things move a tad quicker, especially if the user provides an exact table name match. here's how it works:
1. The TableLookup tool now stores table similarity scores in memory as table_similarities (which makes me think we should probably have typed memory requirements). table_similarities is {table_slug: similarity_score}
2. If any table exceeds the Coordinator.table_similarity_threshold, we skip the LLM table selection process entirely (no more iterative process of examining 3 tables at a time and checking for matches). Instead, we directly provide ALL the closest tables found by TableLookup to the SQLAgent so it has sufficient context for potential joins.
3. From there, the SQLAgent's _find_tables method handles sorting the rest.

also, I'd like to propose a new convention, going forward, we should always store table slugs (i.e. SOURCE__@__TABLE ) in the memory object

philippjfr (approved):

Issue #1084, opened 2025-03-01
`get_schema` enum limitation due to 1000 rows fetched only


ahuang11:
Just marking this so I don't forget to investigate...

I was asking it:
`What are all the Category 5 hurricanes in the Atlantic since 2000?`

```
Schema:
```yaml
basin:
  enum:
  - SP
  - SI
  - WP
  type: str
```

The dataset should have more than just these three, and should have an ellipses. For enums I think we should include unique value count.

ahuang11:
Okay not a bug, just that the limit didn't capture the other basins `'BASIN': {'type': 'string', 'enum': [' ', 'SP', 'SI', 'NI']},`

PR #1085, opened 2025-03-01
Refactor SQLTransforms internals with sqlglot and add SQLSample

ahuang11:
Closes https://github.com/holoviz/lumen/issues/1084

Adds the shuffle to have a better randomly sample of the data so that enums are not limited to the first 1000, but to random 1000 rows, especially useful if the data is chronological / grouped specifically.

### DIFF
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -781,7 +781,6 @@ class BaseSQLSource(Source):
                 sql_expr = sql_transform.apply(sql_expr)
             data = self.execute(sql_expr)
             schemas[entry] = schema = get_dataframe_schema(data)['items']['properties']
-            print(schema)
             if limit:
                 continue
 
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -31,7 +31,7 @@ from ..filters.base import Filter
 from ..state import state
 from ..transforms.base import Filter as FilterTransform, Transform
 from ..transforms.sql import (
-    SQLCount, SQLDistinct, SQLLimit, SQLMinMax, SQLSample,
+    SQLCount, SQLDistinct, SQLLimit, SQLMinMax, SQLSample, SQLTransform,
 )
 from ..util import get_dataframe_schema, is_ref, merge_schemas
 from ..validation import ValidationError, match_suggestion_message
@@ -724,7 +724,7 @@ class BaseSQLSource(Source):
     # Declare this source supports SQL transforms
     _supports_sql = True
 
-    def _apply_transforms(self, source, sql_transforms):
+    def _apply_transforms(self, source: Source, sql_transforms: list[SQLTransform]) -> Source:
         if not sql_transforms:
             return source
         sql_expr = source._sql_expr
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -822,15 +822,20 @@ class BaseSQLSource(Source):
                     cast = str
                 else:
                     cast = lambda v: v
-                min_data = minmax_data[f'{col}_min'].iloc[0]
+
+                # some dialects, like snowflake output column names to UPPERCASE regardless of input case
+                min_col = f'{col}_min' if f'{col}_min' in minmax_data else f'{col}_MIN'
+                min_data = minmax_data[min_col].iloc[0]
+                max_col = f'{col}_max' if f'{col}_max' in minmax_data else f'{col}_MAX'
+                max_data = minmax_data[max_col].iloc[0]
                 schema[col]['inclusiveMinimum'] = min_data if pd.isna(min_data) else cast(min_data)
-                max_data = minmax_data[f'{col}_max'].iloc[0]
                 schema[col]['inclusiveMaximum'] = max_data if pd.isna(max_data) else cast(max_data)
 
             count_expr = SQLCount(read=self.dialect).apply(sql_expr)
             count_expr = ' '.join(count_expr.splitlines())
             count_data = self.execute(count_expr)
-            schema['__len__'] = cast(count_data['count'].iloc[0])
+            count_col = 'count' if 'count' in count_data else 'COUNT'
+            schema['__len__'] = cast(count_data[count_col].iloc[0])
 
         return schemas if table is None else schemas[table]
 
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -109,7 +109,7 @@ def cached_schema(method, locks=weakref.WeakKeyDictionary()):
         with main_lock:
             schema = self._get_schema_cache() or {}
         tables = self.get_tables() if table is None else [table]
-        if all(table in schema for table in tables) and limit is None and shuffle is False:
+        if all(table in schema for table in tables) and limit is None and not shuffle:
             return schema if table is None else schema[table]
         for missing in tables:
             if missing in schema:
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -100,7 +100,7 @@ def cached(method, locks=weakref.WeakKeyDictionary()):
 
 def cached_schema(method, locks=weakref.WeakKeyDictionary()):
     @wraps(method)
-    def wrapped(self, table: str | None = None, limit: int | None = None):
+    def wrapped(self, table: str | None = None, limit: int | None = None, shuffle: bool = False):
         if self in locks:
             main_lock = locks[self]['main']
         else:
@@ -109,7 +109,7 @@ def cached_schema(method, locks=weakref.WeakKeyDictionary()):
         with main_lock:
             schema = self._get_schema_cache() or {}
         tables = self.get_tables() if table is None else [table]
-        if all(table in schema for table in tables) and limit is None:
+        if all(table in schema for table in tables) and limit is None and shuffle is False:
             return schema if table is None else schema[table]
         for missing in tables:
             if missing in schema:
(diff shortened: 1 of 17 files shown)


ahuang11:
Hmmm something interesting is happening with caching or something:


```
# in ai/utils/get_schema.py
if field == "BASIN":
    print(field, spec)

BASIN {'type': 'string', 'enum': [' ', 'SP', 'SI', 'NI']}
BASIN {'type': 'str', 'enum': [' ', 'SP', 'SI', 'NI']}
BASIN {'type': 'str', 'enum': [' ', 'SP', 'SI', '...']}

```

```
# in lumen/sources/base.py
```
if col == "BASIN":
    print(distinct_expr, "EXPR")
    print(schema[col]['enum'], "ENUM")
```
---
SELECT DISTINCT     "BASIN" FROM ( SELECT   * FROM READ_PARQUET('ibtracs.last3years.parquet') ) EXPR
['SP', 'EP', ' ', 'WP', 'SI', 'NI', 'NA'] ENUM
```

ahuang11:
Okay, I found the subtle issue--there was a if limit: continue, so shuffle is still necessary.

Schema before
```
{'SID': {'type': 'string', 'enum': [' ', '2022008S13148', '2022008S17173', '2022020S13059', '2022025S11091', '2022030S14147', '2022034S16082', '2022038S19164', '2022042S12063', '2022047S15073', '2022054S13100', '2022055S13129', '2022056S12088', '2022058S17163', '2022059S10106', '2022062N06085', '2022065S16055', '2022072S11107', '2022076S10126']}, 'SEASON': {'type': 'string', 'enum': ['Year', '2022']}, 'NUMBER': {'type': 'string', 'enum': [' ', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18']}, 'BASIN': {'type': 'string', 'enum': [' ', 'SP', 'SI', 'NI']}, 'SUBBASIN': {'type': 'string', 'enum': [' ', 'EA', 'WA', 'MM', 'BB']}, 'NAME': {'type': 'string', 'enum': [' ', 'TIFFANY', 'CODY', 'ANA', 'BATSIRAI', 'UNNAMED', 'CLIFF', 'DOVI', 'DUMAKO', 'EMNATI', 'VERNON', 'ANIKA', 'EVA', 'GOMBE', 'BILLY', 'CHARLOTTE']}, 'ISO_TIME': {'type': 'string', 'enum': [' ', '2022-01-08 00:00:00', '2022-01-08 03:00:00', '2022-01-08 06:00:00', '2022-01-08 09:00:00', '
...
```

Schema after (notice better sampling of the basins)
```
{'basin': {'type': 'string', 'enum': ['SI', 'WP', 'NA', 'NI', 'EP', 'SP']}, 'name': {'type': 'string', 'enum': ['ANIKA', 'MALOU', 'HERMINE', 'MINDULLE', 'NOT_NAMED', 'IDALIA', 'TAMMY', 'KARL', 'LUPIT', 'DOKSURI', 'NEPARTAK', 'BEATRIZ',
```

lumen/sources/base.py
```diff
@@ -722,6 +722,14 @@ class BaseSQLSource(Source):
     # Declare this source supports SQL transforms
     _supports_sql = True
 
+    def _apply_transforms(self, source, sql_transforms):
```
amaloney, line 725: I'm assuming typing this would be difficult, and while nice to have, it is a "hidden" method so maybe it's not even necessary

amaloney (approved):
I'd be curious to see if knowing the value counts distribution, and then sampling from that will help random samplings have more information about the underlying distribution. Only a thought experiment, as this looks great as is

philippjfr (commented):
In principle this is a good idea, in fact I considered this when first implementing limits on the schema. In practice though, shuffling data is a potentially extremely expensive operation. The reason we added the limit was not merely to limit the number of categories that might be returned but rather to avoid executing queries across the entire table.

ahuang11:
Maybe targeted for specific DBs, or even just DuckDB

From ChatGPT:

### Cost of "Sample" Operations in Different SQL Databases

The cost of sampling operations depends on the database engine and the specific sampling method used. Here's a breakdown:

#### 1. **`TABLESAMPLE` (PostgreSQL, SQL Server)**
   ```sql
   SELECT * FROM my_table TABLESAMPLE (10 ROWS);
   ```
   **Cost:** ✅ Efficient (if supported)  
   - Uses physical data distribution, often via **block/page-level sampling**, making it **fast** for large tables.  
   - **Downside:** Not truly random at the row level—biased by storage layout.  

#### 2. **`ORDER BY RANDOM()` (PostgreSQL, SQLite, MySQL)**
   ```sql
   SELECT * FROM my_table ORDER BY RANDOM() LIMIT 10;
   ```
   **Cost:** ❌ Expensive  
   - **Why?** `RANDOM()` must be evaluated for every row, sorting the entire table, then picking 10 rows.  
   - **Scales poorly** as the table grows.  

#### 3. **`USING SAMPLE` (DuckDB)**
   ```sql
   SELECT * FROM my_table USING SAMPLE 10;
   ```
   **Cost:** ✅ Efficient  
   - Uses internal optimizations, often at the **block level**, avoiding a full-table scan.  

#### 4. **`ORDER BY RAND()` (MySQL)**
   ```sql
   SELECT * FROM my_table ORDER BY RAND() LIMIT 10;
   ```
   **Cost:** ❌ Very Expensive  
   - Similar to `ORDER BY RANDOM()`, `RAND()` is computed for every row, then sorted.  
   - **Alternative:**  
     ```sql
     SELECT * FROM my_table WHERE RAND() < 0.01 LIMIT 10;
     ```
     (Less accurate but avoids full sort.)  

### **Best Approach for Large Tables**
- **If `TABLESAMPLE` or `USING SAMPLE` is available:** ✅ Use it (efficient).  
- **If only `ORDER BY RANDOM()/RAND()` is available:** ⚠️ Consider an alternative:
  - **Precompute random IDs** and select from them.
  - Use a **hashed random sample**:  
    ```sql
    SELECT * FROM my_table WHERE MOD(HASH(id), 100) < 1 LIMIT 10;
    ```
    (Only works if IDs are evenly distributed.)

ahuang11:
Alternatively tricks like:
https://stackoverflow.com/questions/249301/simple-random-samples-from-a-mysql-sql-database

>  ran into this problem again with a bigger table, and ended up using a version of @ignorant's solution, with two tweaks:

Sample the rows to 2-5x my desired sample size, to cheaply ORDER BY RAND()
Save the result of RAND() to an indexed column on every insert/update. (If your data set isn't very update-heavy, you may need to find another way to keep this column fresh.)
To take a 1000-item sample of a table, I count the rows and sample the result down to, on average, 10,000 rows with the the frozen_rand column:

SELECT COUNT(*) FROM table; -- Use this to determine rand_low and rand_high

  SELECT *
    FROM table
   WHERE frozen_rand BETWEEN %(rand_low)s AND %(rand_high)s
ORDER BY RAND() LIMIT 1000

ahuang11:
:| 

image

ahuang11:
Was motivated to refactor the internals to sqlglot because the SQLLimit was brittle, e.g. somehow produced: `SELECT * FROM (SELECT * FROM SELECT * FROM "table") LIMIT 1;

I have to do a bit more testing, but I think I'd like to also refactor 
```

            import sqlglot
            sql_clean = sqlglot.transpile(sql_query, write=source.dialect, pretty=True)[0]
            if sql_query != sql_clean:
                step.stream(f'\n\nSQL was cleaned up and prettified:\n\n```sql\n{sql_clean}\n```')
                sql_query = sql_clean
```
And display the final transformed query as a step for visibility to the user.

lumen/sources/base.py
```diff
@@ -122,10 +122,10 @@ def wrapped(self, table: str | None = None, limit: int | None = None):
             with lock:
                 with main_lock:
                     new_schema = self._get_schema_cache() or {}
-                if missing in new_schema and limit is None:
+                if missing in new_schema and limit is None and shuffle is False:
```
philippjfr, line 125: Why not?
```suggestion
                if missing in new_schema and limit is None and not shuffle:
```

Same above.

lumen/tests/sources/test_duckdb.py
```diff
@@ -119,6 +119,9 @@ def test_duckdb_filter(duckdb_source, table_column_value_type, dask, expected_fi
     table, column, value, _ = table_column_value_type
     kwargs = {column: value}
     filtered = duckdb_source.get(table, __dask=dask, **kwargs)
+    print(table, column, value)
+    print(filtered)
+    print(expected_filtered_df)
```
philippjfr, line 124:
```suggestion
```

lumen/tests/transforms/test_sql.py
```diff
@@ -1,85 +1,306 @@
...
+    assert "A IS NULL" in result
+
 
 def test_sql_filter_scalar():
-    assert (
-        SQLFilter.apply_to('SELECT * FROM TABLE', conditions=[('A', 1)]) ==
-        """SELECT\n    *\nFROM ( SELECT * FROM TABLE )\nWHERE ( A = 1 )"""
-    )
+    result = SQLFilter.apply_to("SELECT * FROM TABLE", conditions=[("A", 1)])
+    assert "WHERE" in result
```
philippjfr, line 62: Why not validate the full SQL statements as before?

ahuang11: Initially was defaulting to pretty=True and it was a pain to get the spacing exactly right in tests, but now that pretty=False by default, I suppose I can re-implement.

lumen/transforms/sql.py
```diff
@@ -1,32 +1,76 @@
...
 import datetime as dt
-import textwrap
+import re
 
-from typing import Any, ClassVar
+from typing import ClassVar
 
-import numpy as np
 import param  # type: ignore
+import sqlglot
```
philippjfr, line 9: What pulls in sqlglot right now? Do we need to make it an explicit dependency of Lumen?

philippjfr (approved):
Just saw that my question about the sqlglot dependency was already addressed. Love this PR, thanks for initiating the work on it!

PR #1088, opened 2025-03-04
Add Snowflake conn_kwargs and private_key support

ahuang11:
Missing private_key, but snowflake has way too many ways to authenticate, so I think it's better to just add conn_kwargs

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -339,7 +339,8 @@ class TableListAgent(Agent):
             widths={'Table': '90%'},
             disabled=True,
             page_size=10,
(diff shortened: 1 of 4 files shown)


philippjfr:
I agree we should add the kwargs but since we're using the `private_key_file` let's add that and the `private_key_file_pwd`.

ahuang11:
Ok, added `private_key` and `private_key_password` where `private_key` supports both a path and bytes.

I took it from:
https://github.com/PrefectHQ/prefect/blob/main/src/integrations/prefect-snowflake/prefect_snowflake/credentials.py

philippjfr (approved):

PR #1097, opened 2025-03-07
Exclude tables

ahuang11:
Some tables are off limits and may raise an error if trying to access it. To address that, we may want to pre-compile a list of tables that are inaccessible.

Builds off https://github.com/holoviz/lumen/pull/1074 because that other one is getting big.

### DIFF
--- a/lumen/sources/base.py
+++ b/lumen/sources/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import fnmatch
 import hashlib
 import json
 import re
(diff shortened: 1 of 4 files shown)


lumen/sources/base.py
```diff
@@ -833,9 +832,64 @@ class BaseSQLSource(Source):
 
     load_schema = param.Boolean(default=True, doc="Whether to load the schema")
 
+    excluded_tables = param.List(default=[], doc="""
+        List of table names that should be excluded from the results.""")
```
philippjfr, line 836: Either declare as constant or automatically rebuild the regex on change. Also nitpick, remove the extra empty line.

philippjfr (commented):
The regex exclusion logic seems somewhat complex and it's unclear to me to what extent actual regex patterns are allowed. Would something simpler like `fnmatch` be simpler?

In any case document the patterns that are supported in the paramter docstring!

Also keep parameters alphabetically sorted unless there's a good reason not to.

philippjfr (approved):

PR #1098, opened 2025-03-11
Refactor table look up

ahuang11:
Builds upon #1074 

Trying to integrate TableLookup without redundant table schemas in the prompt--ended up with huge changes, so I separated this PR from #1074

I noticed there were duplicate tools context in the prompt if Planner also inherited from Actor, so I added another Agent layer in the template.

Some issues I need to investigate:
1. Using Snowflake; I don't understand why I get this with `make_context_model`
```
instructor.exceptions.InstructorRetryException: Error code: 422 - {'detail': 'Data validation error', 'errors': [{'message': 'Input should be a valid dictionary or instance of UserMessage', 'type': 'model_type', 'code_context': None, 'field': 'instruction'}]}
```
2. The vector store returns really small similarities for exact matches:
```python
import lumen.ai as lmai
from lumen.ai.vector_store import NumpyVectorStore

vs = NumpyVectorStore()
vs.add(
    items=[
        {
            "text": """ProvidedSource00000__@__windturbines.parquet
Columns:
- case_id:
- faa_ors:
- faa_asn:
- usgs_pr_id:
- eia_id:
- t_state:
- t_county:
- t_fips:
- p_name:
- p_year:
- p_tnum:
- p_cap:
- t_manu:
- t_model:
- t_cap:
- t_hh:
- t_rd:
- t_rsa:
- t_ttlh:
- retrofit:
- retrofit_year:
- t_conf_atr:
- t_conf_loc:
- t_img_date:
- t_img_srce:
- xlong:
- ylat:
- easting:
- northing:
""",
            "metadata": {
                "source": "ProvidedSource00000",
                "table_slug": "ProvidedSource00000__@__windturbines.parquet",
            },
        }
    ]
)
vs.query("p_tnum")
```


lumen/ai/agents.py
```diff
@@ -243,7 +242,7 @@ class ChatAgent(Agent):
         default={
             "main": {
                 "template": PROMPTS_DIR / "ChatAgent" / "main.jinja2",
-                "tools": [TableLookup, DocumentLookup]
+                "tools": [DocumentLookup]
```
ahuang11, line 245: The context is provided through coordinator.

lumen/ai/agents.py
```diff
@@ -636,7 +620,7 @@ def _update_spec(self, memory: _Memory, event: param.parameterized.Event):
         memory['sql'] = event.new
 
     @retry_llm_output()
-    async def _find_tables(self, messages: list[Message], tables_schema_str: str, errors: list | None = None) -> tuple[dict[str, BaseSQLSource], str]:
+    async def _find_tables(self, messages: list[Message], errors: list | None = None) -> tuple[dict[str, BaseSQLSource], str]:
```
ahuang11, line 623: Maybe in the near future, I'd migrate this to Coordinator and have do a loop through::

From the top 10 results in embeddings, present all tables and their column names to the LLM for selection
Have the LLM select approximately 3 tables, then request their complete schemas (including column values, enumerations, minimums, maximums) and get_metadata descriptions
If the LLM is satisfied with the context, proceed to the agent steps; otherwise, restart the process with awareness of previously examined tables

philippjfr:
> In the background it also begins fetching actual metadata to enrich it. Ideally, this is already cached to disk, or prepared up by the time the user accesses the app.

We should have a mechanism to pre-populate on launch somehow.

>This context (and every other tools context by the coordinator) would be passed down to every agent; I don't know if I like this, but I couldn't think of another solution at the moment.

Can you clarify, is it all tool context passed down to all agents or just table lookup specifically?

>to write a valid query, it should know enums + min/maxes, so the SQLAgent uses find_tables to THEORETICALLY... subset from the most relevant table in the tool context

I don't really understand what this means. Are we providing schemas or no?

>Since it has access to ALL the tables tables_model = self._get_model("find_tables", tables=tables), it sometimes finds a better choice than the "most relevant tables" (which means our embedding method is not the best yet)

What does all tables mean here? If it's all tables then don't we again balloon the token length here?

>I think in the near future, I'd like to extract find_tables from SQLAgent to Coordinator and introduce yet another while loop:

How much extra work is this because that sounds like the most sensible approach and after reading your points I'm not 100% sure how that differs from what's done now.

ahuang11:
> Can you clarify, is it all tool context passed down to all agents or just table lookup specifically?

All tool context 

> I don't really understand what this means. Are we providing schemas or no?

Basic schema is provided to Coordinator/Agents (e.g. name, description, column names+types+descriptions) while SQLAgent is the only one that has the full schema (min/max/enum).

> What does all tables mean here? If it's all tables then don't we again balloon the token length here?

That is correct unfortunately. :( 

> How much extra work is this because that sounds like the most sensible approach and after reading your points I'm not 100% sure how that differs from what's done now.

Right now, it's a one shot thing. But yes I agree this should be implemented asap since we balloon in tokens again.

ahuang11:
Okay I added `_iterative_table_selection` which basically performs an iterative table selection process to gather context.
1. From the top tables in TableLookup results, selects 3 at a time
2. Gets complete schemas for these tables
3. Repeats until the LLM is satisfied with the context

I also added refine_query in VectorLookupTool based on https://python.useinstructor.com/blog/2024/09/26/implementing-anthropics-contextual-retrieval-with-async-processing
> Anthropic uses Claude to generate context. They provide this prompt:

philippjfr:
You happy with it after some testing?

ahuang11:
Not 100% due to https://github.com/holoviz/lumen/pull/1103 (inconsistent usage of normalize_table), but because I tried to extract it to a new branch on main, it caused a bunch of merge conflicts, so I re-stashed popped everything and it caused a merge conflict nightmare :'( so I need to clean that up

image

ahuang11:
However, I think we can consolidate this into the add_get_metadata branch.

PR #1100, opened 2025-03-12
Add _llm_spec_key for granular LLM control

ahuang11:
Allows user to specify something like

```
OpenAI(
   models={"reasoning": ..., "chat": ..., "sql": ..., "base_view": ...}
)
```

So that the developers can specify very granular control per agent/tool/coordinator.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -566,7 +566,7 @@ class SQLAgent(LumenBaseAgent):
             has_errors=bool(errors),
         )
         with self.interface.add_step(title=title or "SQL query", steps_layout=self._steps_layout) as step:
-            model_spec = self.prompts["main"].get("llm_spec", "default")
+            model_spec = self.prompts["main"].get("llm_spec", self._llm_spec_key)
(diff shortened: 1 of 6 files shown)


lumen/ai/agents.py
```diff
@@ -127,7 +129,7 @@ def __panel__(self):
 
     async def _stream(self, messages: list[Message], system_prompt: str) -> Any:
         message = None
-        model_spec = self.prompts["main"].get("llm_spec", "default")
+        model_spec = self.prompts["main"].get("llm_spec", self._llm_spec_key)
```
ahuang11, line 132: It'll still use default if the key isn't found
https://github.com/holoviz/lumen/pull/1100/files#diff-dc864a2206353ab4b7498c9a9361f5c45af0c5fd84adbc59fe054515346e55d4L77-L83

philippjfr (approved):
Seems like the spec key could simply be agent class name minus `Agent` and then camel case -> snake case.

ahuang11:
So maybe a property that does that?

PR #1102, opened 2025-03-12
Improve to work with Sonnet

ahuang11:
I discovered that the data field actually is required, and Sonnet was following instructions too closely, thus kept crashing.

Also key error with SQLLimit if no limit was set.


philippjfr (commented):
The problem that led to adding the instruction was that it sometimes would literally make up data (or at least restate data it had seen) verbatim.

ahuang11:
Okay changed the messaging
image

philippjfr (approved):

Issue #1105, opened 2025-03-13
If CSV has special characters or spaces, breaks the sqlglot


ahuang11:
Try uploading data from https://ourworldindata.org/life-expectancy
[life-expectancy.csv](https://github.com/user-attachments/files/19235784/life-expectancy.csv)

PR #1106, opened 2025-03-13
Fix parsing sql

ahuang11:
A little disappointed that sqlglot couldn't automagically parse `SELECT * FROM read_csv('life-expectancy')` and column names like `Period life expectancy at birth - Sex: total - Age: 0`, but still okay...

Closes https://github.com/holoviz/lumen/issues/1105


amaloney (approved):
lgtm

PR #1110, opened 2025-03-17
Cast Snowflake's date columns into timestamp for readability

ahuang11:
I noticed panel-graphic-walker / graphic-walker not properly serializing `datetime.date` types for human readability:

image

To fix this, casting it to `pd.datetime64` works.
```python
import datetime
import numpy as np
import pandas as pd
import panel_gwalker
import panel as pn

pn.extension()

# timeseries dataframe
df = pd.DataFrame({
    'date': pd.to_datetime([datetime.date(2020, 1, 1) + datetime.timedelta(days=i) for i in range(3)]),
    'value': np.random.randn(3).cumsum()
})
panel_gwalker.GraphicWalker(df)
```
image

### DIFF
--- a/lumen/sources/snowflake.py
+++ b/lumen/sources/snowflake.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import decimal
 import re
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Why not fix this in panel-graphic-walker?

ahuang11:
I guess we never merged https://github.com/panel-extensions/panel-graphic-walker/pull/53 so I implemented it here, but I made changes in https://github.com/panel-extensions/panel-graphic-walker/pull/53 too

philippjfr:
Just tagged panel-graphic-walker 0.6.4, so I guess just update this PR with the new pin.

PR #1116, opened 2025-03-19
Prevent deadlock

ahuang11:
I was experiencing deadlocks with asyncio.gather with duckdb sources so I reverted it back to sequential. I think it has to do with calling asyncio.to_thread underneath.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -99,9 +99,6 @@ class Coordinator(Viewer, Actor):
         those tables will be automatically selected for the closest_tables
         and there will not be an iterative table selection process.""")
 
-    max_concurrent = param.Integer(default=3, doc="""
-        The maximum number of concurrent tasks to run.""")
-
     __abstract = True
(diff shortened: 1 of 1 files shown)


lumen/ai/coordinator.py
```diff
@@ -709,9 +704,7 @@ async def _iterative_table_selection(self, sources: dict) -> dict:
 
                 # For subsequent iterations, the LLM selects tables in the previous iteration
                 step.stream(f"\n\nGathering complete schema information for {len(selected_tables)} tables...")
-                schema_tasks = [self._lookup_table_schema(source_table, sources) for source_table in selected_tables]
-                async with self._semaphore:
-                    schema_results = await asyncio.gather(*schema_tasks)
+                schema_results = [await self._lookup_table_schema(source_table, sources) for source_table in selected_tables]
```
amaloney, line 707: this is clever, I quite like it

amaloney (approved):

PR #1118, opened 2025-03-20
Refactor retry output

ahuang11:
Adds hierarchy to retry_output jinja template to match all the others, and modify VegaLiteAgent's

Also, fix blank plots by not including "params" if latitude/longitude or point: true


amaloney (approved):
nice language updates

PR #1119, opened 2025-03-20
Do not normalize twice

ahuang11:
Was confused as to why the SQL statement with a limit wasn't applying (LIMIT 3, but 72357 rows)

image

Realized I was normalizing the table twice, e.g. the first one returned a valid SQL statement `table = self.tables[self.normalize_table(table)]`, but the second time, it extracted the table, then applied the default `sql_expr`.


ahuang11:
I incurred it :P

PR #1122, opened 2025-03-20
Add default favicon

ahuang11:
Not sure if this is the proper place to add it, and whether reading from URL is fine, or whether it should be added to each individual application?

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -297,6 +297,7 @@ class UI(Viewer):
             template = state.template
             if isinstance(template, FastListTemplate):
                 template.main_layout = None
+            template.favicon = "https://raw.githubusercontent.com/holoviz/lumen/refs/heads/main/doc/_static/favicon.ico"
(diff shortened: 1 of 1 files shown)


amaloney (approved):
favicons from urls are not the norm, but this fixes the issue sufficiently.

I'm not 100% sure why it isn't the norm, but I can guess that if the site is down where the favicon is pulled from, then you won't get it showing up

lumen/ai/ui.py
```diff
@@ -297,6 +297,7 @@ def _create_view(self, server: bool | None = None):
             template = state.template
             if isinstance(template, FastListTemplate):
                 template.main_layout = None
+            template.favicon = "https://raw.githubusercontent.com/holoviz/lumen/refs/heads/main/doc/_static/favicon.ico"
```
philippjfr, line 300:
```suggestion
            template.favicon = "https://lumen.holoviz.org/_static/favicon.ico"
```

PR #1127, opened 2025-03-20
Insert context

ahuang11:
I noticed with the `add_get_metadata` changes, if there are less than 5 tables, the ChatAgent no longer has any context about the datasets, so here, I create an artificial `TableLookup` tool context. Not 100% sure this is the best approach.

Also, renaming source_table -> table_slug for consistency.


lumen/ai/coordinator.py
```diff
@@ -467,30 +459,6 @@ def _serialize(self, obj: Any, exclude_passwords: bool = True) -> str:
             obj = obj.value
         return str(obj)
 
-    def _fuse_messages(self, messages: list[Message], max_user_messages: int = 2) -> list[Message]:
```
ahuang11, line 470: Moved to utils.py

lumen/ai/coordinator.py
```diff
@@ -632,227 +600,13 @@ class Planner(Coordinator):
 
     prompts = param.Dict(
         default={
-            "context": {
```
ahuang11, line 635: Let the planner sort it out because it'll likely invoke TableLookup if it needs context.

lumen/ai/tools.py
```diff
@@ -31,6 +36,9 @@ class Tool(Actor, ContextProvider):
         Whether to always use this tool, even if it is not explicitly
         required by the current context.""")
 
+    interface = param.ClassSelector(class_=ChatInterface, allow_None=True, doc="""
```
ahuang11, line 39: Need it to stream steps.

lumen/ai/models/details.js
```diff
@@ -0,0 +1,34 @@
+export function render({ model, view }) {
```
ahuang11, line 1: Since TableLookup output both text and schemas, the schema took up lots of space in the ChatStep, and it wasn't formatted well, so I created this Details component to hide it.


image

image

ahuang11:
Okay, I made huge changes, but significantly cleans up the codebase and makes it less complex i.e. coordinator/planner simply plans, and no longer handles iterative table selection; that's all in TableLookup, and SQLAgent depends on its output `tables_sql_schemas`, and cleans up the memory key space a bit, e.g. `closest_tables` is removed.

lumen/ai/actor.py
```diff
@@ -14,6 +15,9 @@
 
 class Actor(param.Parameterized):
 
+    interface = param.ClassSelector(class_=ChatInterface, doc="""
```
philippjfr, line 18: Keeping tools separate from an interface was a deliberate decision so we keep blurring the line between a Tool and an Agent.

ahuang11:
1. I noticed that Tool inherited from Actor, and Actor has the method _use_tools which I don't think is quite right (i.e. Tool using tools doesn't make sense), so I created a ToolUser mixin that the Coordinator and Agents use.
2. I unified the tools handling between Coordinator & Agent so that there is no more special __main__ key in self._tools--it's now part of the main prompt. I think the reason why that was introduced was because previously _use_tools applied all tools every time, depending on requires...
3. But now, we have a _gather_prompt_context that Actor has to inject memory into prompt template, and this is what Agent subclasses to additionally inject tool context into the prompt template
4. On the other hand, Planner will use its own _execute_graph_node to decide whether or not it needs to use tools/agents
5. Lastly, I completely separated Tool from requiring an interface by introducing a NullStep null context manager. However, Agent is heavily tangled, and probably best to address in a separate PR since this one is getting huge

lumen/ai/tools.py
```diff
@@ -534,23 +569,160 @@ async def respond(self, messages: list[Message], **kwargs: dict[str, Any]) -> st
...
+            available_slugs = [
+                table_slug for table_slug in closest_tables
+                if table_slug not in examined_slugs
+            ]
+            if not available_slugs:
+                log_debug("No more tables available to examine")
+                break
+
+            fast_track = False
+            # if iteration == 1:
```
philippjfr, line 647: What's the plan for this?

philippjfr (approved):

PR #1132, opened 2025-03-21
Add status badges

ahuang11:
Add loading status badges up in header

https://github.com/user-attachments/assets/b3069afc-09f3-4454-be42-0bca853d5de1

### DIFF
--- a/lumen/ai/components.py
+++ b/lumen/ai/components.py
@@ -24,10 +24,13 @@ CSS = """
 .left-panel-content {
     max-width: clamp(450px, 95vw, 1200px);
     margin: 0px auto;  /* Center the content */
-    background-color: var(--panel-background-color); /* Use theme variable for content background */
+    background-color: #ffffff; /* Brighter white background for better contrast */
     border-radius: 10px; /* Slight rounded corners */
     overflow-y: auto;   /* Enable scrolling if content is taller than the area */
     box-sizing: border-box; /* Include padding in size calculations */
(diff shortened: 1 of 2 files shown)


lumen/ai/ui.py
```diff
@@ -175,56 +176,45 @@ def __init__(
             sizing_mode='stretch_width'
         )
         self._main = Column(self._exports, self._coordinator, sizing_mode='stretch_both')
+        self._llm_status_badge = StatusBadge(name="Checking LLM")
+        self._vector_store_status_badge = StatusBadge(name="Preparing vector store")
```
amaloney, line 180: Nice separation between the two

lumen/ai/components.py
```diff
@@ -235,3 +239,157 @@ def _handle_msg(self, msg):
...
+    )
+
+    _base_stylesheet = """
+    :host {
+        position: relative;
+    }
+
+    /* Base badge styles */
+    .bk-btn {
+        border-radius: 16px !important;
```
amaloney, line 260: once you put `!important` on things, you've got to start putting them everywhere. it looks like this is happening

amaloney (approved):
I'd like to see the css in a separate file, but that's only a preference of mine

lumen/ai/components.py
ahuang11: Thanks for pointing that out. I found out I didn't target the right class. Styling in Panel/Bokeh is pretty tedious.

PR #1140, opened 2025-03-27
More fixes optimizations

ahuang11:


amaloney (changes requested):
I think the test errors came from the quotes

amaloney (approved):
thanks!

PR #1141, opened 2025-03-27
Add view definition to views

ahuang11:
Previously, I added a timeout for get_schema in TableLookup, but that simply:

- changed 5 minute waits with context to provide to the LLM 
- into 30 second waits with **no context** to provide to the LLM

Instead of doing that, I thought maybe I can just extract the view definition (view's derivative SQL) and share that to the LLM.

This is the first step; separate PR to integrate with TableLookup

Also fixes a bug to use uppercase names in Snowflake (was encountering KeyError without it)

### DIFF
--- a/lumen/sources/snowflake.py
+++ b/lumen/sources/snowflake.py
@@ -381,7 +381,11 @@ class SnowflakeSource(BaseSQLSource):
             }
             if table_type == "VIEW" and "INFORMATION_SCHEMA" not in table_slug:
                 try:
-                    result[table_slug]["view_definition"] = self.execute("SELECT GET_DDL('VIEW', %s)", ('.'.join(table_slug.split(".")[1:]),)).iloc[0, 0]
+                    # DATABASE.SCHEMA.TABLE -> SCHEMA.TABLE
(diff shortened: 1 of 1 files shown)


lumen/sources/snowflake.py
```diff
@@ -378,4 +379,13 @@ def subset_table_slugs(df: pd.DataFrame) -> pd.DataFrame:
...
             }
+            if table_type == "VIEW" and "INFORMATION_SCHEMA" not in table_slug:
+                try:
+                    # DATABASE.SCHEMA.TABLE -> SCHEMA.TABLE
+                    table_schema_slug = '.'.join(table_slug.split(".")[1:])
+                    result[table_slug]["view_definition"] = self.execute(
+                        "SELECT GET_DDL('VIEW', %s)", (table_schema_slug,)
+                    ).iloc[0, 0]
+                except Exception:
+                    result[table_slug]["view_definition"] = "N/A"
```
amaloney, line 390: this is an excellent catch, and the comment is quite helpful

amaloney (approved):

PR #1142, opened 2025-03-27
Refactor lookup

ahuang11:
Finally semi-satisfied with TableLookup.

1. Separates out TableLookup and IterativeTableLookup purpose. TableLookup is suitable for ChatAgent because it just uses vector store queries to quickly find all the relevant tables & columns + descriptions if available, while IterativeTableLookup is suitable for SQLAgent because it additionally makes database calls to look up schema, e.g. min/max of column values
2. Rather than try to look up schemas for tables that are actually views (taking 30 seconds or longer), uses the view definition (the SQL) as context.
3. Renames table_info to tables_sql_info for clarity to differentiate against tables_vector_info.

Next steps are:
ColumnLookup and Gated propagation for tool context

### DIFF
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -530,8 +530,9 @@ class TableLookup(VectorLookupTool):
 
         tables_vector_info = {}
         any_matches = any(result['similarity'] >= self.min_similarity for result in results)
+        same_table = len([result["metadata"]["table_name"] for result in results])
         for result in results:
-            if any_matches and result['similarity'] < self.min_similarity and len(results) > 1:
+            if any_matches and result['similarity'] < self.min_similarity and not same_table:
                 continue
             source_name = result['metadata']["source"]
             table_name = result['metadata']["table_name"]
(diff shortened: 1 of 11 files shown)


lumen/ai/tools.py
```diff
@@ -708,42 +714,60 @@ async def _select_tables(self, sources, messages, closest_tables, table_similari
...
-                step.stream("\nFetching detailed schema information for tables\n")
+                step.stream("\n\nFetching detailed schema information for tables\n")
                 for table_slug in selected_slugs:
                     table_name = table_slug.split(SOURCE_TABLE_SEPARATOR)[-1]
-                    table_similarity = table_similarities.get(table_slug, 0)
-                    step.stream(f"`{table_name}` (Similarity: {table_similarity:.2f})")
+                    vector_info = tables_vector_info[table_slug]
+                    similarity = vector_info["similarity"]
+                    caption = vector_info["caption"]
+                    step.stream(f"`{table_name}` (Similarity: {similarity:.2f})")
```
amaloney, line 748: not sure it matters, but everywhere else we are rounding to 3 digits

amaloney (approved):
I like the name split `table_info` -> `table_sql_info` and `table_vector_info`

PR #1143, opened 2025-03-28
Gated Context Propagation and Dataclasses

ahuang11:
- Dataclasses handle rendered context and data propagation
- Only check for tool relevance with YesNo if Tool.provided doesn't match Agent.required; else pass it
- Add column subsetting in TableLookup

TablesSQLMetaset > TableSQLMetadata
TablesSQLMetaset > TablesVectorMetaset > TableVectorMetadata > sel_tables_cols


philippjfr:
Seeing a bunch of breakpoints left all over. Almost ready here?

lumen/ai/agents.py
```diff
@@ -261,6 +261,7 @@ async def respond(
         step_title: str | None = None,
     ) -> Any:
         context = {"tool_context": await self._use_tools("main", messages)}
+        breakpoint()
```
```suggestion
```

PR #1146, opened 2025-04-03
Reimplement the UI on top of panel-material-ui

philippjfr:
This PR has multiple goals:

- Redesign the UI and UX to be cleaner and clearer
- Use `panel-material-ui` components for consistent look and feel
- Refactor the unmanageable `ExplorerUI` class with state on a bunch of different variables and instead move state onto `Exploration` objects.
- Factor out a `TableExplorer` class


ahuang11:
Based on my initial testing, I think the material UI shows too little information to the user; there's a lot of waiting (and wondering if it's working), which I don't think is ideal, especially since Lumen is still not very stable yet.

FWIW, when Gemini hid the CoT, [people complained about it](https://www.reddit.com/r/Bard/comments/1kr5yo4/new_update_ruined_gemini_25_cot_is_now_hidden/):
> Can't properly debug the thought process. For example, I can't see wrong assumptions by the LLM anymore
> I now have to wait for the entire thinking process to complete to obtain an answer (minutes). Even though the thought process often contains key insights within the first few seconds.

philippjfr:
> Based on my initial testing, I think the material UI version shows too little information to the user; there's a lot of waiting (and wondering if it's working, like here), which I don't think is ideal, especially since Lumen is still not very stable yet.

I'm quite confused by this, I cannot reproduce what you're seeing at all. All the context is shown when I test it.

philippjfr:
You were right initially, steps weren't reported after the first plan. Got a fix.

ahuang11:
11. Sometimes the input gets sent, but persists in the chat area
image

philippjfr:
I'm going to merge this. That doesn't mean this is now complete and polished, but I want us to test this and want to avoid further merge conflicts.

ahuang11:
Exciting!! Will test this soon!

PR #1148, opened 2025-04-03
Improve vega lite

ahuang11:
1. Used the proper retry_output path for VegaLite
2. Support for interactive, zoomable geographic maps
3. Tweak prompt so that Planner doesn't call SQLAgent again if data is sufficient

image


lumen/ai/agents.py
```diff
@@ -903,6 +916,44 @@ async def _update_spec(self, memory: _Memory, event: param.parameterized.Event):
...
+            vega_spec["params"] = []
+        vega_spec["params"].extend(VEGA_ZOOMABLE_MAP_ITEMS["params"])
+        if "projection" not in vega_spec:
+            vega_spec["projection"] = {"type": "mercator"}
+        vega_spec["projection"].update(VEGA_ZOOMABLE_MAP_ITEMS["projection"])
+
+        # add outlines
+        if "world-110m.json" in vega_spec_str and vega_spec["projection"]["type"] == "albersUsa":
+            vega_spec["projection"] = "mercator"  # cannot use albersUsa with world-110m
+        elif "world-110m.json" not in vega_spec_str and vega_spec["projection"]["type"] != "albersUsa":
```
philippjfr, line 935: Confused by these if/else clauses (actually most the `elif` part). Also it seems fairly presumptive to add a US layer automatically.

ahuang11: first if: using albersUsa with coastlines causes warping
image

second elif, if no coastline and the projection isn't albersUsa, add a coastline layer

third elif, if no us states but projection is albersUsa, add a states layer (assuming no other coastline layer, which would be caught in first if)

philippjfr (approved):

PR #1160, opened 2025-04-11
Add DbtslLookup and DbtslAgent

ahuang11:
This PR introduces a pre-planning step consisting of tools, which allows the Planner to gather a bit more context before deciding on what Actors (agents & tools) to select.

For each tool in the pre-planning step, there's a yes/no relevance check (admittedly I don't think it works super well).

The DbtslLookup tool simply would do a vector search to find the most relevant defined metrics embedded, and would provide a a dbtsl_metaset containing metrics' metadata:
```
    name: str
    similarity: float
    description: str | None = None
    dimensions: dict[str, list] = field(default_factory=dict)
    queryable_granularities: list[str] = field(default_factory=list)
```
Now, it includes a LLM to check if the metrics and dimensions are able to satisfy the query.

Then, the Planner starts planning, and *likely* would choose the DbtslAgent based on all the prompting (in the near future, I'd like to work on https://github.com/holoviz/lumen/issues/1161 rather than randomly injecting negative prompts like this all over)

This DbtslAgent includes a LLM step that looks at the most similar metrics and decides which metrics to use. It then populates DbtslQueryParams (like where and group_by but it sometimes does have trouble including dimensions referenced in where also in group_by and suffixing the time granularity). Once a `DbtslQueryParams` is generated, it uses those kwargs inside `dbtsl_client.compile_sql` and does what the SQLAgent does (saving `create_sql_expr_source`, `pipeline`, etc)

On follow up questions, there's another YesNo whether the user is asking a follow up question--if so, skip the pre-planning step.

### DIFF
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 
 from abc import abstractmethod
 from contextlib import nullcontext
@@ -347,3 +348,26 @@ class ContextProvider(param.Parameterized):
 
     async def requirements(self, messages: list[Message]) -> list[str]:
         return self.requires
+
+
+class DbtSLMixin(param.Parameterized):
+
+    auth_token = param.String(default=None, doc="""
+        The auth token for the dbt semantic layer client;
+        if not provided will fetched from DBT_AUTH_TOKEN env var.""")
+
+    environment_id = param.Integer(default=None, doc="""
+        The environment ID for the dbt semantic layer client.""")
+
+    host = param.String(default="semantic-layer.cloud.getdbt.com", doc="""
+        The host for the dbt semantic layer client.""")
+
+    def __init__(self, environment_id: int, **params):
+        from dbtsl.asyncio import AsyncSemanticLayerClient
+
+        super().__init__(environment_id=environment_id, **params)
+        self._dbtsl_client = AsyncSemanticLayerClient(
+            environment_id=self.environment_id,
+            auth_token=self.auth_token or os.environ.get("DBT_AUTH_TOKEN"),
+            host=self.host,
+        )
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -363,11 +363,14 @@ class DbtSLMixin(param.Parameterized):
         The host for the dbt semantic layer client.""")
 
     def __init__(self, environment_id: int, **params):
-        from dbtsl.asyncio import AsyncSemanticLayerClient
 
         super().__init__(environment_id=environment_id, **params)
-        self._dbtsl_client = AsyncSemanticLayerClient(
+
+    def _get_dbtsl_client(self):
+        from dbtsl.asyncio import AsyncSemanticLayerClient
+        dbtsl_client = AsyncSemanticLayerClient(
             environment_id=self.environment_id,
             auth_token=self.auth_token or os.environ.get("DBT_AUTH_TOKEN"),
             host=self.host,
         )
+        return dbtsl_client
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -350,7 +350,7 @@ class ContextProvider(param.Parameterized):
         return self.requires
 
 
-class DbtSLMixin(param.Parameterized):
+class DbtslMixin(param.Parameterized):
 
     auth_token = param.String(default=None, doc="""
         The auth token for the dbt semantic layer client;
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -1,5 +1,4 @@
 import datetime
-import os
 
 from abc import abstractmethod
 from contextlib import nullcontext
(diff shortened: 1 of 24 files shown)


philippjfr:
Looks like a great start but I'm not totally clear on how it's an alternative to the `TableLookup` search. Is it really true that there's a 1:1 mapping between tables and entries in the Dbt semantic layer? My assumption was that the semantic layer contains specific views that provide common and well defined queries for specific metrics that an organization is interested in. If my assumption holds it seems like it it should be used in addition to the TableLookup and also take higher precedence, i.e. if there is a good match on a metric in the semantic layer it would use that pre-defined query instead of asking the LLM to generate a new custom query. So I really need some more context on how Dbt is usually used.

ahuang11:
I implemented what we discussed, but there are a few issues to be addressed and things I need to test before merging. Feel free to start reviewing though!

lumen/ai/agents.py
```diff
@@ -723,6 +725,183 @@ async def respond(
...
+                    limit=query_params.get('limit'),
+                    order_by=query_params.get('order_by'),
+                    where=query_params.get('where')
+                )
+
+                # Log the compiled SQL for debugging
+                step.stream(f"\nCompiled SQL:\n```sql\n{sql_query}\n```", replace=False)
+
+            # Create a SQL source from the compiled query
+            source = self._memory["source"]
```
philippjfr, line 853: Not a totally safe assumption that the `source` is the same source that you can execute the Dbt generated queries against. Seems like we might have to make source a parameter on the agent in this case.

philippjfr:
Please update the top comment with the actual approach taken and also chime in with your experience using it so far.

lumen/ai/agents.py
```diff
@@ -723,6 +725,183 @@ async def respond(
...
+
+            # Create pipeline and get data
+            pipeline = await get_pipeline(
+                source=sql_expr_source, table=expr_slug, sql_transforms=sql_transforms
+            )
+
+            df = await get_data(pipeline)
+
+            # Update memory
+            self._memory["data"] = await describe_data(df)
```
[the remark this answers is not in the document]
ahuang11: SQLAgent has it:
```

        try:
            df = await get_data(pipeline)
        except Exception as e:
            if isinstance(source, DuckDBSource):
                source._connection.execute(f'DROP TABLE IF EXISTS "{expr_slug}"')
            report_error(e, step)
            raise e

        self._memory["data"] = await describe_data(df)
```
It's used in AnalystAgent:
```
Here is the current dataset:
{{ memory.data }}
```

ahuang11:
Updated; I think it works okay--much faster than IterativeTableLookup and it is able to fall back to that if necessary. However, I think there's an issue when I edit the yaml spec, it says pipeline doesn't include the table and I can't figure out why.

ahuang11:
Added a LLM to the DbtslLookup to assess whether it's able to answer the query before propagating context.

Also, added a follow up YesNo gate before pre-planning.

lumen/ai/actor.py
```diff
@@ -347,3 +348,29 @@ class ContextProvider(param.Parameterized):
 
     async def requirements(self, messages: list[Message]) -> list[str]:
         return self.requires
+
+
+class DbtslMixin(param.Parameterized):
```
philippjfr, line 353: Would move that elsewhere.

lumen/ai/coordinator.py
```diff
@@ -504,19 +525,38 @@ async def respond(self, messages: list[Message], **kwargs: dict[str, Any]) -> st
...
+
+            # If it's a follow-up and we already have the data, we can skip some tools execution
+            if is_follow_up:
+                log_debug("\033[92mDetected follow-up question, using existing context\033[0m")
+                with self.interface.add_step(title="Using existing data context...", user="Assistant") as step:
+                    step.stream("Detected that this is a follow-up question related to the previous dataset.")
+                    step.stream("\n\nUsing the existing data in memory to answer without re-executing data retrieval.")
+                    step.success_title = "Using existing data for follow-up question"
+
+            execution_graph = await self._compute_execution_graph(messages, agents, is_follow_up=is_follow_up)
```
philippjfr, line 541: We should have a stub on `Coordinator._compute_execution_graph` to define the API.

lumen/ai/coordinator.py
```diff
@@ -702,15 +741,115 @@ class Planner(Coordinator):
...
+            "follow_up": {
+                "template": PROMPTS_DIR / "Planner" / "follow_up.jinja2",
+                "response_model": YesNo,
+            },
         }
     )
 
+    def __init__(
+        self,
+        llm: Llm | None = None,
```
philippjfr, line 763: Instead of repeating all the arguments, let's disallow positional arguments on the base class.

lumen/ai/prompts/VegaLiteAgent/main.jinja2
```diff
@@ -77,9 +77,6 @@ spec:
 {{ super() }}
 
 {% if errors is defined and errors %}
-For extra context, here are the tables and columns available:
-{{ columns_context }}
```
philippjfr, line 81: Should be able to restore this after computing sql_metaset.

lumen/sources/snowflake.py
```diff
@@ -226,8 +228,36 @@ def create_sql_expr_source(self, tables: dict[str, str], **kwargs):
         params['conn'] = self._conn
         return SnowflakeSource(**params)
 
+    def _cast_to_supported_dtypes(self, df: pd.DataFrame, sample: int = 100) -> pd.DataFrame:
```
ahuang11, line 231: To temporarily support decimals instead of crashing.

lumen/ai/coordinator.py
```diff
@@ -101,10 +101,6 @@ class Coordinator(Viewer, VectorLookupToolUser):
             "main": {
                 "template": PROMPTS_DIR / "Coordinator" / "main.jinja2",
             },
-            "yesno_update": {
```
ahuang11, line 104: This was not used.

ahuang11:
Cleaned up based on your comments and other things I've found.

I also added an internal snowflake decimal handling temporarily so it doesn't crash for now.

PR #1167, opened 2025-04-15
Drop stop words

ahuang11:
This **significantly** improves results returned from vector stores' queries.

Took the stop words from:
https://github.com/stopwords-iso/stopwords-en/blob/master/stopwords-en.txt


amaloney (approved):
The list of stop words is nice for possible air gapped machines, but we may want to use a library in the future for a stop word list that is more expansive, and has other language support

PR #1168, opened 2025-04-16
Refactor error handling

ahuang11:
Previously, error guidance was injected haphazardly and also hardcoded. This allows users to update the error guidance as necessary.

Also, I noticed since we were mutating the user message, the error guidance was duplicated multiple times.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -32,8 +32,7 @@ from .tools import (
     IterativeTableLookup, TableLookup, Tool, ToolUser,
 )
 from .utils import (
-    fuse_messages, log_debug, mutate_user_message, retry_llm_output,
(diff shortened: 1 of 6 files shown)


lumen/ai/coordinator.py
```diff
@@ -364,21 +363,13 @@ async def _chat_invoke(self, contents: list | str, user: str, instance: ChatInte
         log_debug(f"New Message: \033[91m{contents!r}\033[0m", show_sep="above")
         await self.respond(contents)
 
-    @retry_llm_output()
```
ahuang11, line 367: Not entirely sure this is necessary, and it was a bit complex to include the system prompt here with planner so I just removed.

philippjfr (approved):
Nice, looks much better.

PR #1183, opened 2025-04-23
Various fixes

ahuang11:
- SourceAgent was missing provides sources
- Do not include unsatisfiable actors if their requirements are not in all of the actors' provides or memory
- A bit more planner tweaks
- cols_to_show: None is not iterable
- `query` was not using a filter
- duckdb wal can get corrupted if refreshed before complete

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -201,7 +201,7 @@ class SourceAgent(Agent):
 
     requires = param.List(default=[], readonly=True)
 
-    provides = param.List(default=["source", "document_sources"], readonly=True)
+    provides = param.List(default=["sources", "source", "document_sources"], readonly=True)
 
     on_init = param.Boolean(default=True)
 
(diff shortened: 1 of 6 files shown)


amaloney (approved):

ahuang11:
More fixes:

1. Prefixed vector_metaset and sql_metaset from dbtsl with `dbtsl_` because the Planner kept inheriting the old `sql_metaset` from dbtsl on completely separate topics, instead of using IterativeTableLookup

2. In planner, I now mention what the previous actors tried were so it doesn't continuously choose the same bad actors

3. I use gpt-4.1-mini as the default sql OpenAI model (vs gpt-4o-mini)

4. Prompt tweaks (will clean up in a `conditions` PR

5. Fix select columns (previously, it based it of all the available tables--now it's just the selected tables)

6. Do not be so hasty to finalize tables on the first iterative table selection

Issue #1161, opened 2025-04-11
Precedence and `use_with` and `incompatible_with` params in Agents


ahuang11:
Right now, we have these prompts:
```
Important Agent Rules:
- The SourceAgent is only useful for uploading new table or document sources; not for showing what's already available, which is the job for TableListAgent.
- The SQLAgent can generate and execute multiple queries in a single step with joins. DO NOT create two separate steps for generating the query and then executing it.
- The SQLAgent is a better candidate than TableListAgent if asked to show the table, and usually is followed by the AnalystAgent, which can help the user understand the results of the query.
- The SQLAgent usually needs help by *TableLookup if the current columns are insufficient.
- Only use SQLAgent when working with new data or different analysis requirements. If the user wants to visualize, discuss, or further analyze data that's already been retrieved, use the current SQL rather than executing the query again.
- If the user is asking for a plot of the data without any modifications, SQLAgent is unnecessary.
- When plotting, instruct SQLAgent to select more columns than requested to enable VegaLiteAgent to create faceted, multi-layout, or encoded visualizations.
- The ChatAgent must be used with TableLookup or other tools, but if the query is related to the data tables, please use AnalystAgent instead.
```

Which is semi-hard-coding. It'd be better if it's associated with the agents' params.

philippjfr:
Similarly we have `requires` but not the inverse which is to indicate that an actor shouldn't be used if a particularly piece of context is available. As an example in #1160 we are thinking of adding a Dbt tool that provides context about relevant metrics. If a user query can be answered by referring to a particular metric definition we would not want to call the SQLAgent, so we might want to declare on the `SQLAgent` that it should not be called if `dbt_metaset` is available.

ahuang11:
I think I might prefer a `conditions = param.List()` for maximal flexibility

PR #1187, opened 2025-04-24
Add situational conditions param

ahuang11:
Partially addresses https://github.com/holoviz/lumen/issues/1161

Cleans up the Planner prompt significantly and improves separation of concerns by managing the fuzzy conditions under each agent.

Then also tweaks other prompts

Next, I want to add hard dependencies: `exclusions` (keys that are incompatible) and `not_with` (actors that are incompatible)

### DIFF
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -11,8 +11,7 @@ import param
 from panel.chat import ChatFeed
 from pydantic import BaseModel
 
-from lumen.ai.config import PROMPTS_DIR
-
+from .config import PROMPTS_DIR
 from .llm import Llm, Message
 from .memory import _Memory, memory
 from .utils import log_debug, render_template, warn_on_unused_variables
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -335,15 +335,34 @@ class ContextProvider(param.Parameterized):
     requires and provides.
     """
 
+    conditions = param.List(default=[], doc="""
+        Specific criteria that determine when this actor should be invoked.
+        These conditions establish explicit relationships between actors in the system,
+        defining the circumstances under which this actor becomes relevant.
+        While 'purpose' describes what the actor does, conditions specify
+        the precise situations that warrant its use.""")
+
     provides = param.List(default=[], readonly=True, doc="""
         List of context values it provides to current working memory.""")
 
     purpose = param.String(default="", doc="""
-        Describes the purpose of this actor for consumption of
-        other actors that might invoke it.""")
+        A descriptive statement of this actor's functionality and capabilities.
+        Serves as a high-level explanation for other actors to understand
+        what this actor does and when it might be useful to invoke it.""")
 
     requires = param.List(default=[], readonly=True, doc="""
         List of context values it requires to be in memory.""")
 
     async def requirements(self, messages: list[Message]) -> list[str]:
         return self.requires
+
+    def __str__(self):
+        string = (
+            f"- {self.name[:-5]}\n"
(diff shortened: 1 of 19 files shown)


lumen/ai/agents.py
```diff
@@ -504,14 +521,30 @@ async def retry_invoke(event: param.parameterized.Event):
 
 class SQLAgent(LumenBaseAgent):
 
+    conditions = param.List(
+        default=[
+            "Preferred for data queries requiring table joins, filtering, or new calculations",
+            "Better choice than TableListAgent when asked to show table data",
+            "Works with IterativeTableLookup when additional columns are needed",
+            "Only use for new data analysis; for visualization of existing results, use current data",
+            "Defer to DbtslAgent when dbtsl_metaset is available",
```
philippjfr, line 530: Only comment on this PR is that I think the more generic agent shouldn't be listing conditions about a more specific agent, i.e. I'd invert this and have DbtAgent have a condition that says: "Prefer DbtslAgent if dbtsl_metaset" is available.

philippjfr (commented):
Overall this is a great improvement. Just left one comment that I think is important for this to be scalable, i.e. generic agents can't list every possible interaction with more specific agents.

ahuang11:
>  i.e. generic agents can't list every possible interaction with more specific agents.

I agree with this to a certain degree, but I think it can be very helpful for some agents to specify its relationship (i.e. SQLAgent with IterativeTableLookup), or else the Planner stumbles.

```
    conditions = param.List(default=[], doc="""
        Specific criteria that determine when this actor should be invoked.
        These conditions **establish explicit relationships between actors in the system**,
        defining the circumstances under which this actor becomes relevant.
        While 'purpose' describes what the actor does, conditions specify
        the precise situations that warrant its use.""")
```

Also, `not_with` now handles: `Defer to DbtslAgent when dbtsl_metaset is available` without mentioning it.

Other than that...
1. I've revamped most Actors' prompts
2. I've created a `_pre_plan` method in Coordinator. I realized that the exclusions must happen AFTER the planning tools are executed are outdated. For example, on the first run, `dbtsl_metaset` is generated in memory. Second run, if exclusions happen before the planning tools and DbtslLookup decides no relevant metrics, both SQLAgent and dbtsl_metaset is dropped (thus DbtslAgent), leaving no options to perform queries until the next message, but if exclusions happen after the planning tools, then the planning tools can inform what exclusions to make.
3. `is_follow_up` is now like the previous invalidation step that we removed, popping data/pipeline
4. I migrated some current memory context into `follow_up` prompt vs in the `planner` prompt

changed parts of them into conditions/exclusions/not_with and a few bugs like Actor not defining its main prompt.

philippjfr:
>I agree with this to a certain degree, but I think it can be very helpful for some agents to specify its relationship (i.e. SQLAgent with IterativeTableLookup), or else the Planner stumbles.

I agree, I'd think of those as a pair and very foundational.

lumen/ai/actor.py
```diff
@@ -11,6 +11,8 @@
 from panel.chat import ChatFeed
 from pydantic import BaseModel
 
+from lumen.ai.config import PROMPTS_DIR
```
philippjfr, line 14: Please be consistent in the imports. I know relative imports aren't everyone's cup of tea but as long as we're using them let's be consistent about it.

philippjfr (approved):
Looks good, thanks.

lumen/ai/actor.py
ahuang11: Turns out we were very inconsistent about it heh
https://github.com/holoviz/lumen/pull/1187/commits/2edc698cc8fe11fb85a1d6c5cb8559625c4a6f7a

PR #1189, opened 2025-04-24
Fix status badge

ahuang11:
The issue was that when table_lookup._ready is became False, it never accessed those clauses. If we move it outside, it works.
image


lumen/ai/ui.py
```diff
@@ -188,7 +188,6 @@ def __init__(
             sizing_mode='stretch_width'
         )
         self._main = Column(self._exports, self._coordinator, sizing_mode='stretch_both')
-        self._llm_status_badge = StatusBadge(name="LLM Pending")
```
ahuang11, line 191: This wasn't used anywhere

amaloney (approved):

PR #1199, opened 2025-04-29
Fix environment variable name for Anthropic API key in anthropic.md

parulnith:
I updated the environment variable name from `OPENAI_API_KEY` to `ANTHROPIC_API_KEY` in the setup instructions.
Since the section is about configuring Anthropic as the provider, using the correct key name avoids confusion and setup errors.


ahuang11 (approved):

ahuang11:
Thank you!

PR #1202, opened 2025-05-01
Local LLM refactor

amaloney:
- Refactors how key/values are displayed in the UI.
- Refactors using a `pop` from a dictionary to a `get` with a default value specified.


ahuang11 (approved):

ahuang11:
I assume changing pop -> get fixes the issues?

amaloney:
There is still a problem with the local LLM erroring out with an arcane code that I have yet to figure out that does not happen when using OpenAI, but this does fix some errors I was facing, so it's a start

ahuang11:
If it was related to the n_ctx, error, I had to move things around to get the n_ctx=0 to register.

Issue #1210, opened 2025-05-06
Store embeddings used in vector store's metadata


ahuang11:
Was stumped as to why results were empty when querying the vector store and the issue was that when you embed it with OpenAIEmbeddings, you must remember to use the same embeddings when re-opening it.

```
import lumen.ai as lmai
import panel as pn

INSTRUCTIONS = """
You are an expert scientist working in Python, with a specialty using Anndata and Scanpy.
Help the user with their questions, and if you don't know the answer, say so.
"""

vector_store = lmai.vector_store.DuckDBVectorStore(
    uri="scanpy.db",
    embeddings=lmai.embeddings.OpenAIEmbeddings(),
)
doc_lookup = lmai.tools.VectorLookupTool(vector_store=vector_store, min_similarity=0.0)
```

PR #1214, opened 2025-05-07
Ensure consistent embeddings are used

ahuang11:
Closes https://github.com/holoviz/lumen/issues/1210

If stored chunks was embedded with a certain embeddings model, but re-opened with another embeddings model, on query, it'll return no results.

### DIFF
--- a/lumen/ai/vector_store.py
+++ b/lumen/ai/vector_store.py
@@ -1,4 +1,5 @@
 import asyncio
+import importlib
 import io
 import json
 import os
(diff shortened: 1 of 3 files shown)


lumen/ai/vector_store.py
```diff
@@ -953,6 +1004,91 @@ def _setup_database(self, embedding_dim: int) -> None:
...
+            return  # No metadata table, can't check consistency
+
+        result = self.connection.execute(
+            "SELECT value FROM vector_store_metadata WHERE key = 'embeddings';"
+        ).fetchone()
+
+        if not result:
+            return
+
+        stored_config = json.loads(result[0])
```
amaloney, line 1030:
```suggestion
        stored_config = self._get_embeddings_config() or {"class": "", "params": {}}
```

amaloney (approved):
small change to remove duplicated code, otherwise approved

PR #1221, opened 2025-05-09
Implement add_directory method to vector store

ahuang11:
Adds a convenient method to add entire directories to the vector store.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,7 @@ filterwarnings = [
     "error",
     "ignore:datetime.datetime.utcfromtimestamp():DeprecationWarning:dateutil.tz.tz", # https://github.com/dateutil/dateutil/pull/1285
     "ignore:Couldn't find ffmpeg or avconv - defaulting to ffmpeg:RuntimeWarning",
+    "ignore:Unsupported Windows version",
 ]
 
 [tool.ruff]


amaloney:
What about using `faker` to generate the markdown data as a fixture, and then use `tempfile` to create a temp directory that the tests could read from?

ahuang11:
> What about using faker to generate the markdown data as a fixture, and then use tempfile to create a temp directory that the tests could read from?

Is there a benefit to using faker since that's adding a dependency? Theoretically, one could simply do something like "a" * 10000 and write it out, but I don't see the need to overcomplicate it.

amaloney (approved):

lumen/ai/vector_store.py
```diff
@@ -358,6 +360,76 @@ async def upsert(self, items: list[dict], situate: bool | None = None) -> list[i
...
+            # Check against exclusion patterns
+            if any(fnmatch(str(rel_path), exclude_pattern) for exclude_pattern in exclude_patterns):
+                continue
+
+            # Create file-specific metadata by extending base metadata
+            file_metadata = base_metadata.copy()
+            file_metadata["filename"] = str(file_path)
+
+            all_ids.extend(
+                await self.add_file(
```
droumis, line 424: Is it necessary to await each individual add_file call within the loop? Might be faster to use asyncio.gather on a collection

droumis (approved):
looks good. maybe make the file processing concurrent

Issue #1212, opened 2025-05-07
Integrate Model Context Protocol (MCP) Servers with Lumen


ahuang11:
The Model Context Protocol (MCP) ecosystem is growing fast, and so I think we should integrate it. MCP gives LLMs a standardized way to use external tools and data sources, which would let us tap into functionality that already exists rather than rebuilding everything ourselves.

For quick context of how MCP works, there's a server that defines a list of tools, resources, etc that the LLM can access:
```python
from fastmcp import FastMCP, Client

mcp = FastMCP("My MCP Server")

@mcp.tool()
def greet(name: str) -> str:
    return f"Hello, {name}!"

if __name__ == "__main__":
    mcp.run()
```

Then there is a client (Lumen):
```python
import asyncio
from fastmcp import Client

client = Client("my_server.py")

async def call_tool(name: str):
    async with client:
        print(await client.list_tools())
        result = await client.call_tool("greet", {"name": name})
        print(result)

asyncio.run(call_tool("Ford"))
```

Here I am not suggesting we expose Lumen as a MCP server (yet), just as the client / consumer of MCP servers.

For example, [here](https://python.useinstructor.com/blog/2025/03/27/understanding-model-context-protocol-mcp/?h=mcp), they access firecrawl with an api key. There are a bunch of existing servers https://github.com/punkpeye/awesome-mcp-servers that are available.

My suggestion is to follow our existing `DbtslLookup`/`DbtslAgent` pattern. We'd create an `MCPLookupTool` (placed in `planning_tools` in the `Planner`) that indexes multiple MCP servers using vector embeddings. The lookup tool would populate an `mcp_metaset` in memory with all the discovered tools and resources.
```python
class MCPLookupTool(VectorLookupTool):
    """
    Tool that indexes multiple MCP servers using vector embeddings
    and performs semantic search to find relevant tools and resources.
    """
    
    servers = param.List(class_=MCPServer, doc="""
        List of MCP servers to index.""")
    
    provides = param.List(default=["mcp_metaset"], readonly=True)
    
    async def _update_vector_store(self):
        # Index tools and resources from all servers
        for server in self.servers:
            client = await self._get_client(server)
            tools = await client.list_tools()
            # Create embeddings for each tool...

```

We'd pair this with an `MCPAgent` that consumes the `mcp_metaset` (like DbtslAgent with dbtsl_metaset`, picks the best tool for the job, and executes it with parameters extracted from the user's query.

```python
class MCPAgent(Agent):
[...]

PR #1224, opened 2025-05-09
Sketch MCP

ahuang11:
Starts on:
https://github.com/holoviz/lumen/issues/1212

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -39,8 +39,8 @@ from .controls import RetryControls, SourceControls
 from .llm import Llm, Message
 from .memory import _Memory
 from .models import (
-    DbtslQueryParams, PartialBaseModel, RetrySpec, Sql, VegaLiteSpec,
-    make_find_tables_model,
+    DbtslQueryParams, MCPToolExecution, PartialBaseModel, RetrySpec, Sql,
+    VegaLiteSpec, make_find_tables_model,
 )
 from .schemas import get_metaset
 from .services import DbtslMixin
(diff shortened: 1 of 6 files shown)


lumen/ai/prompts/Planner/main.jinja2
```diff
@@ -68,6 +111,11 @@ From query: "{{ memory['vector_metaset'].query }}"
 📋 Last table: `{{ memory['table'] }}`
 {%- endif %}
 
+You already have these provided in memory so you can use them directly in your plan:
+{%- for key in memory.keys() %}
+- `{{ key }}`
+{%- endfor %}
+
```
ahuang11:
```suggestion
```

philippjfr (commented):
I'm wondering if this isn't overcomplicating things quite a bit. My assumption was that we would simply add an `MCPTool` that you can configure with a specific `server_uri`. It fetches the server metadata and it uses that to populate the purpose. You then configure as many `MCPTool` instances as you need, so the planner can pick between them.

The current approach requires the agent to run the `MCPLookupTool` without knowing what MCP tools are even available, so really the Planner always has to run the lookup tool (which I guess would be okay), but I'm also not fully convinced similarity search will always actually surface all the relevant tools.

philippjfr:
Overall I'm okay with it, but I wonder if we shouldn't offer my suggested simpler approach too, hopefully without duplicating much code.

lumen/ai/agents.py
```diff
@@ -1377,6 +1378,179 @@ async def _extract_spec(self, spec: dict[str, Any]):
...
+        """Check if this agent should be used based on available MCP capabilities."""
+        metaset = memory.get("mcp_metaset")
+        return metaset is not None and (bool(metaset.tools) or bool(metaset.resources))
+
+    async def _get_client(self, server_url: str, transport_type: str = "auto"):
+        """Get or create a client for the specified server URL."""
+        try:
+            from fastmcp import Client
+        except ImportError:
+            raise ImportError("Please install the fastmcp package to use MCP agents")
```
```suggestion
        except ImportError as e:
            raise ImportError("Please install the fastmcp package to use MCP agents") from e
```

lumen/ai/tools.py
```diff
@@ -1522,6 +1522,213 @@ async def _fetch_dimension_values(self, client, metric_name, dim, num_cols, step
...
+        self._mcp_resources = {}
+        # Start indexing right away
+        state.execute(self._update_vector_store)
+
+    async def _get_client(self, server: MCPServer):
+        """Get or create a client for the specified server."""
+        try:
+            from fastmcp import Client
+        except ImportError:
+            raise ImportError("Please install the fastmcp package to use MCPLookupTool")
```
ahuang11:
```suggestion
        except ImportError as e:
            raise ImportError("Please install the fastmcp package to use MCPLookupTool") from e
```

ahuang11:
Wondering if we should refactor Planner into an IterativePlanner first before merging this.

PR #1230, opened 2025-05-12
Deduplicate work done

ahuang11:
A follow up on https://github.com/holoviz/lumen/pull/1227

I think that fixed another issue, but still encountering crashes on a deployment, and I noticed that we were doing the same thing twice due to TableLookup and IterativeTableLookup both using the same source:
```python

---
2025-05-12 21:19:19,821 Enriching 198 table metadata entries.
2025-05-12 21:19:20,425 Enriching 198 table metadata entries.
```

### DIFF
--- a/lumen/ai/tools.py
+++ b/lumen/ai/tools.py
@@ -228,6 +228,9 @@ class VectorLookupTool(Tool):
     Baseclass for tools that search a vector database for relevant
     chunks.
     """
+
+    # Class variable to track which sources have been embedded in which vector stores
+    _embedded_sources = {}
     enable_query_refinement = param.Boolean(default=True, doc="""
         Whether to enable query refinement for improving search results.""")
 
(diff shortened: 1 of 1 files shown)


lumen/ai/tools.py
```diff
@@ -752,7 +753,22 @@ async def _enrich_metadata(self, source, table_name: str):
     async def _update_vector_store(self, _, __, sources):
         await asyncio.sleep(0.5)  # allow main thread time to load UI first
         tasks = []
+        processed_sources = []
+
+        # Use class variable to track which sources are currently being processed
+        vector_store_id = id(self.vector_store)
```
amaloney, line 759: clever, but is there any way to ensure that there won't be collisions between threads

amaloney: answering my own question here... yes

amaloney (approved):

PR #1233, opened 2025-05-13
Add holoviews agent

ahuang11:
Was able to get it working in notebook, but not thru LLM due to some UnionType issue with instructor


ahuang11:
HoloViewsAgent now works in Lumen after changing | to Union[..., None] in the translator while that's being fixed in instructor.

Then, I found a bunch of edge cases in the translator where pet = param.ClassSelector(class_=(Dog, Cat)) wasn't properly translating to pet: Union[Dog, Cat], handling instantiate=False and proposed that we migrate translate.py to param https://github.com/holoviz/param/issues/828#issuecomment-2878253054

Just need to wrap up translating operations and fixing those warnings
image

PR #1237, opened 2025-05-16
Serializing HoloViews Operations

ahuang11:
Went down the rabbit hole of trying to serialize, deserialize, and now I'm confused.

I was able to make `param` -> `pydantic` -> `param` work successfully in https://github.com/holoviz/lumen/pull/1233, even with nested parameterized, e.g.

```python
class HoloViewsView(View):

    operations = param.List(item_type=[hv.operation.Operation])
```

However, I realized, Lumen works with text specs, and `hv.operation.Operation` does not serialize well into text... so in this PR, I was trying to figure it out and wrap `Operation` into some sort of Lumen component to make serialize/deserialize using to_spec/from_spec.

Initially I created a HoloViewsOperationView, but that didn't seem quite right, even though it outputs a plot, so I created a HoloViewsOperation(Transform), but also questionable.

Would like your advice @philippjfr

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -37,7 +37,7 @@ from ..filters.base import Filter, ParamFilter
 from ..panel import HtmlPdfDownloadButton
 from ..pipeline import Pipeline
 from ..state import state
-from ..transforms.base import Transform
+from ..transforms.base import HoloViewsOperation, Transform
 from ..transforms.sql import SQLTransform
 from ..util import (
     VARIABLE_RE, catch_and_notify, is_ref, resolve_module_reference,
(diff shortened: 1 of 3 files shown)


lumen/views/base.py
```diff
@@ -973,7 +967,7 @@ class hvPlotView(hvPlotBaseView, HoloViewsElementView):
     its simple API.
     """
 
-    operations = param.List(item_type=VALID_HV_OPERATIONS, doc="""
+    operations = param.List(item_type=HoloViewsOperation, doc="""
```
ahuang11, line 970: Is this weird that item_type is a Transform?

PR #1239, opened 2025-05-19
Import numpy

ahuang11:
Not sure if it'll help with anything, but harmless to try
https://duckdb.org/docs/stable/clients/python/known_issues#numpy-import-multithreading


amaloney (approved):
nice catch, and thank you for linking to the documentation in duckdb

PR #1240, opened 2025-05-19
Remove del method

ahuang11:
After merging https://github.com/holoviz/lumen/pull/1238 I've been experiencing `connection is already closed!` in Lumen. I suspect it's because `memory` is a WeakRefDictionary so it has been `del memory["source"]`, which closes the connection.


amaloney (approved):

amaloney:
Fascinating discovery, and good to know and look out for

PR #1241, opened 2025-05-21
Add custom source controls

ahuang11:
Allows users to upload custom data types by specifying `table_upload_callbacks`.

e.g.

```python

def upload_h5ad(file, table) -> int:
    """
    Uploads an h5ad file and returns an AnnDataSource.
    """
    adata = ad.read_h5ad(file)
    try:
        src = AnnDataSource(adata=adata)
        lmai.memory["sources"] = lmai.memory["sources"] + [src]
        lmai.memory["source"] = src
        return 1
    except Exception as e:
        print(f"Error uploading file: {e}")
        return 0

```


lumen/ai/agents.py
```diff
@@ -682,9 +683,14 @@ async def _find_tables(self, messages: list[Message], errors: list | None = None
             vector_metaset.selected_columns or
             vector_metaset.vector_metadata_map
         )
+        if len(selected_slugs) == 0:
+            source = next(iter(sources.values()))
+            selected_slugs = source.get_tables()[:10]
```
amaloney, line 688: is there a reason for selecting only the first ten tables

amaloney (approved):

droumis (approved):

PR #1263, opened 2025-06-10
Tweak prompts

ahuang11:
When I was asking "How do I use the sc.pp.highly_variable_genes function with Pearson residuals?" it was trying to load the dataset, which motivated to me to clarify the conditions. 

Albeit verbose and increases the token usage, I think we should aim to get things working well before optimizing it later with evals.

Also fixes a bug with AnalysisAgent (I think there are only two tabs, enumerating from 0)
                self._main.active = 1 if self.analysis.autorun else 0


lumen/ai/prompts/AnalystAgent/main.jinja2
```diff
@@ -1,7 +1,15 @@
 {% extends 'Agent/main.jinja2' %}
 
 {%- block instructions %}
-You are a world-class data AnalystAgent known for your analytical rigor and critical thinking.
+Act as the world's foremost trusted authority on the data. Your expertise
+lies in using topic-specific jargon and technical language to analyze and interpret complex datasets, providing
+Provide highly strategic, deeply analytical, and expert-level insights that only the top 0.1% of professionals
```
philippjfr: Still not sure how useful this kind of prompting really is these days. Seems like it's just as likely to push it to be overconfident in it's hallucinations as it is to improve the quality of the responses.

philippjfr (approved):

PR #1276, opened 2025-06-13
Validate final results

ahuang11:
Adds a validation step to the final output to check whether the results reflect what the user queried; if not, show a convenient "Rerun" button that posts the suggestions.

image

image


philippjfr:
Fix pre-commit please.

philippjfr (approved):

PR #1279, opened 2025-06-13
Optimize detect encoding for faster startups

ahuang11:
While loading a 19Mb CSV, the Lumen interface was taking a while to load. I thought it was because of the vector store, but discovered it was trying to detect file encoding, the bottleneck being:

```
        with file_obj.open("rb") as f:
            data = f.read()
```

Added sample_size and a few quick checks before using chardet.

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -375,8 +375,11 @@ def detect_file_encoding(file_obj: Path | str | io.BytesIO | io.StringIO, sample
     if isinstance(file_obj, (str, Path)):
         # File path
         file_path = Path(file_obj)
-        with file_path.open("rb") as f:
-            data = f.read(sample_size)
+        if not file_path.exists():
+            data = None
+        else:
+            with file_path.open("rb") as f:
+                data = f.read(sample_size)
     elif isinstance(file_obj, io.BytesIO):
         # BytesIO - preserve position
         pos = file_obj.tell()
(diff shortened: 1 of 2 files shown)


lumen/util.py
```diff
@@ -355,47 +355,69 @@ def slugify(value, allow_unicode=False) -> str:
...
+    if not data:
+        return "utf-8"
+
+    # Check BOM first (instant detection)
+    if data.startswith(b'\xef\xbb\xbf'):
+        return 'utf-8-sig'
+    elif data.startswith(b'\xff\xfe'):
+        return 'utf-16-le'
+    elif data.startswith(b'\xfe\xff'):
+        return 'utf-16-be'
```
amaloney: I think duckdb will error if it gets one of these return strings.

https://duckdb.org/docs/stable/data/csv/overview.html#parameters

lumen/util.py
```diff
@@ -355,47 +355,69 @@ def slugify(value, allow_unicode=False) -> str:
...
+    try:
+        data.decode('utf-8')
+        return 'utf-8'
+    except UnicodeDecodeError:
+        pass
+
+    # Use chardet if available, otherwise fallback
+    try:
+        import chardet
+        result = chardet.detect(data)
```
amaloney: I've encountered csv files that are utf-8 encoded at the beginning, and then later in the file contain latin-1 encoding. It sounds weird, but maybe it happened because of cat-ing two differently encoded csv files together.

ahuang11: Would adding a sample size suffice then? Right now, reading all of the CSV to memory isn't ideal.

philippjfr (approved):

PR #1289, opened 2025-06-17
Only situate if chunk char length is less than specified

ahuang11:
In preparation to embed vega-lite docs, trying to be more efficient about situating.

### DIFF
--- a/lumen/ai/vector_store.py
+++ b/lumen/ai/vector_store.py
@@ -82,11 +82,16 @@ class VectorStore(LLMUser):
         doc="List of metadata keys to exclude when creating the embeddings.",
     )
 
-    situate = param.Boolean(
+    situate = param.ClassSelector(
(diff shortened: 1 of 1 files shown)


lumen/ai/vector_store.py
```diff
@@ -82,11 +82,16 @@ class VectorStore(LLMUser):
...
         default=False,
         doc="""
-        Whether to insert a `llm_context` key in the metadata containing
-        contextual about the chunks.""",
+        Controls whether to add a `llm_context` key to the metadata
+        with contextual information about the chunks.
+        If set to True, all chunks will be situated.
+        If set to an integer, only chunks with a character
+        count less than or equal to this value will be situated.
+        If set to False, no chunks will be situated.""",
```
amaloney, line 94: I would like an example in the doc, because I do not fully understand what is being added

PR #1290, opened 2025-06-18
Add abstraction for batch jobs

ahuang11:
I've noticed that we've been embedding a lot more text in recent times (soon Vega-Lite docs), and the previous method of looping could be costly, so here I introduce batch processing for chunk situating. This could also be useful for batch generating multiple reports in the future.

This PR adds multi-provider batch support (OpenAI, Anthropic, Mistral) that processes chunks in parallel rather than sequentially.

Falls back to sequential processing if batch APIs aren't supported or batch_situate_threshold (default: 1000 chunks, set to 0 to disable).

### DIFF
--- a/lumen/ai/batch_providers.py
+++ b/lumen/ai/batch_providers.py
@@ -1,0 +1,346 @@
+"""
+Multi-provider batch processing implementations for LLM providers.
+"""
+
+import asyncio
+import json
+import time
+import uuid
+
+from abc import abstractmethod
+from typing import Any
+
+import param
+
+from .llm import (
+    AINavigator, AnthropicAI, AzureMistralAI, AzureOpenAI, Llm, MistralAI,
+    OpenAI,
+)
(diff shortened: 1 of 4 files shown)


lumen/ai/embeddings.py
```diff
@@ -91,8 +91,35 @@ def __init__(self, **params):
 
     async def embed(self, texts: list[str]) -> list[list[float]]:
         texts = [text.replace("\n", " ").strip() for text in texts]
-        response = await self.client.embeddings.create(input=texts, model=self.model)
-        return [r.embedding for r in response.data]
+
+        # Batch processing to handle token limits
+        # OpenAI has a 300,000 token limit per request
+        # We'll use a conservative batch size and estimate ~4 chars per token
+        max_chars_per_batch = 1_000_000  # Conservative estimate for ~250k tokens
```
philippjfr: Probably expose this as a parameter.

PR #1301, opened 2025-06-26
Migrate columns discovery to SQLAgent

ahuang11:
Previously:
1. **TableLookup/IterativeTableLookup** pre-selects columns using `_select_columns()` and stores them in `selected_columns`
2. **SQLAgent** then uses these pre-selected columns from the `sql_metaset` context

However, this can be error prone, because the table lookup can be decoupled, and if it selects the wrong columns, the SQLAgent will say "this table doesn't have the columns required" and give a bad answer.

So instead, since SQLAgent already has a discovery phase, I think column selection can be done inside. It eliminates the complex logic around `_should_select_columns`, `_should_refresh_columns`, etc, and also, it also makes the context cleaner by just showing table names and basic info.

I'm also tempted to take out IterativeTableLookup (because it's a complex mess), and have SQLAgent just call TableLookup, but baby steps...


lumen/ai/prompts/SQLAgent/main.jinja2
```diff
@@ -8,9 +8,15 @@ Write a SQL query for the user's data transformation request, focusing on intent
...
 - No inline comments in SQL code
 - NULL handling: Exclude NULLs unless specified otherwise
-- Performance: Use LIMIT (≥100k if unspecified), efficient ORDER BY, filter early
+- Performance: Use LIMIT 100000 if unspecified, efficient ORDER BY, filter early
 - Single SQL query: Provide a single, complete query that answers the user's request
 
+## Table Structure Exploration:
+- **Discovery/exploration queries**: Use `SELECT * LIMIT 1-5` to examine structure and sample values
+- **Final answer queries**: Use `LIMIT 100000` or appropriate limits for the user's needs
+- Use `SELECT DISTINCT "column" LIMIT 10` to explore available values in specific columns
```
amaloney: I think a limit of 10 is too low, and would suggest increasing it to 100

lumen/ai/prompts/SQLAgent/main.jinja2
```diff
@@ -27,15 +33,34 @@ Write a SQL query for the user's data transformation request, focusing on intent
...
 {% block examples %}
-Example:
+Example queries:
+
+**For discovery/exploration (small limits for understanding structure):**
+```sql
+-- Examine table structure and sample data
+SELECT * FROM "table_name" LIMIT 2;
+```
+
```
amaloney, line 38: I wonder if getting quantile information here would help both the LLM and human understand data patterns. Maybe even upper and lower bounds for non numeric columns like

```sql
-- top and bottom counts
SELECT * FROM (
    (
        SELECT TOP 10 <column>, COUNT(1) AS cnt
        FROM <table> GROUP BY <column> ORDER BY cnt DESC
    )
    UNION ALL
    (
        SELECT TOP 10 <column>, COUNT(1) AS cnt
        FROM <table> GROUP BY <column> ORDER BY cnt ASC
    )
)
```



```sql
-- percentage
SELECT * FROM (
(
  SELECT
      TOP 10 <column>
      , COUNT(1) AS cnt
      , cnt * 1.0 / (SELECT COUNT(DISTINCT <column>) FROM <table>) AS ave
  FROM <table>
  WHERE <column> IS NOT NULL
  GROUP BY <column>
  ORDER BY ave DESC
)
UNION ALL
(
  SELECT
      TOP 10 <column>
      , COUNT(1) AS cnt
      , cnt * 1.0 / (SELECT COUNT(DISTINCT <column>) FROM <table>) AS ave
  FROM <table>
  WHERE <column> IS NOT NULL
  GROUP BY <column>
  ORDER BY ave ASC
)
)
-- WHERE ave ...
```
```

amaloney (approved):
lgtm

there are a couple comments we can discuss for possible sql examples for the llm

lumen/ai/prompts/SQLAgent/main.jinja2
ahuang11: These should be targeted, e.g. `ILIKE %VALUE%`, and to prevent context overflow

lumen/ai/prompts/SQLAgent/main.jinja2
ahuang11: This is just to explore the column names.

PR #1305, opened 2025-06-28
Roadmap with iterative baby steps plans

ahuang11:
Motivation for this PR is that the original SQLAgent tried to one-shot everything based on the schema overview provided by the IterativeTableLookup.

However, as you may know, data can be huge, lots of rows, lots of columns, so we can't just capture everything in a single overview...

Instead, we should make agents work like humans... so I reflected on how I would approach analyzing the data, and realized, it's really just breaking it down into smaller queries, and using that info to build the final query.

So here, I introduce a roadmap, plus iterative planning with retries. Not 100% perfect, but much better than trying to 1-shot the results without enough context.

To save time, I think maybe SQLAgent can try one shotting the output, and then if it fails, it can go into planning mode, but maybe in a separate PR.

Also, ideally Lumen Planner would also recursively redesign the plan, but maybe for Lumen 1.0.

### DIFF
--- a/lumen/tests/transforms/test_sql.py
+++ b/lumen/tests/transforms/test_sql.py
@@ -595,7 +595,7 @@ class TestSQLRemoveSourceSeparator:
 
     def test_debug_complex_table_names(self):
         """Test the specific case that's failing with complex table names."""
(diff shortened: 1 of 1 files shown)


ahuang11:
Will fix merge conflicts tomorrow..

philippjfr:
Sadly the SQLAgent on main is also not working well for me and this PR just gave me 14 step plan for my usual "Calculate the total capacity?" query. What do you think, should I revert https://github.com/holoviz/lumen/pull/1301 for now?

ahuang11:
Okay, I think it's probabilistically entering discovery, which results in a long plan. If you'd like, I think we can add a toggle to either disable it, or limit the discovery steps to like 3 at most.

philippjfr:
Okay, I'm seemingly getting better results now and we can tweak more later. Will merge shortly.

philippjfr (approved):

PR #1306, opened 2025-06-30
Gallery workflow fixes

maximlt:
- Add a step to prepare the project before attempting, e.g. to deploy it to catch env solving errors before (having to connect to AE5 and fix stuff there is not fun)
- This workflow gets triggered when the build workflow completes, which also happens when the build workflow runs on the week-end, and leads to redeploying the dev gallery every Sunday. Let's avoid that and instead skip deploying on these types of runs.
- Bring this change that Simon made to Panel's gallery workflow https://github.com/holoviz/panel/pull/6779

### DIFF
--- a/.github/workflows/gallery.yaml
+++ b/.github/workflows/gallery.yaml
@@ -69,7 +69,7 @@ jobs:
           github.event.workflow_run.event != 'schedule' &&
           (
             (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'main') ||
-            (github.event_name == 'workflow_run' && !(contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
+            (github.event_name == 'workflow_run' && !(contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')) && contains(steps.vars.outputs.tag, 'v'))
           )
         run: |
           conda activate test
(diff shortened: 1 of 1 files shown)


.github/workflows/gallery.yaml
```diff
@@ -48,10 +48,17 @@ jobs:
...
           conda install -c ae5-admin anaconda-project ae5-tools
+      - name: Prepare environment to check it solves
+        run: |
+          anaconda-project prepare --directory examples/gallery
+          git clean -fxd examples/gallery
       - name: Upload and deploy dev gallery
         if: |
-          (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'dev') ||
-          (github.event_name == 'workflow_run' && (contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
+          github.event.workflow_run.event != 'schedule' &&
```
hoxbro: Does this work? Doesn't some workflow trigger this after the package has run?

.github/workflows/gallery.yaml
```diff
@@ -48,10 +48,17 @@ jobs:
         run: |
           conda activate test
           conda install -c ae5-admin anaconda-project ae5-tools
+      - name: Prepare environment to check it solves
+        run: |
+          anaconda-project prepare --directory examples/gallery
+          git clean -fxd examples/gallery
```
hoxbro, line 54: Maybe also `git reset --hard` (or similar) so we don't upload changed files by accident.

hoxbro (approved):

PR #1319, opened 2025-07-16
Factor out static Reasoning model

philippjfr:
There was no longer a reason to dynamically create the reasoning model.

### DIFF
--- a/lumen/ai/models.py
+++ b/lumen/ai/models.py
@@ -170,19 +170,22 @@ class LineChange(BaseModel):
         To remove a line set this to an empty string."""
     )
 
-class BaseStep(BaseModel):
+
+class RawStep(BaseModel):
     actor: str
     instruction: str
     title: str
 
-class BasePlan(BaseModel):
+
+class RawPlan(BaseModel):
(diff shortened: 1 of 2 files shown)


lumen/ai/models.py
```diff
@@ -170,35 +170,45 @@ class LineChange(BaseModel):
         To remove a line set this to an empty string."""
     )
 
+class BaseStep(BaseModel):
```
ahuang11: Should this be called PlanStep instead? Or what's the "Base" referencing (will there be subclasses for this)?

lumen/ai/models.py
```diff
@@ -170,35 +170,45 @@ class LineChange(BaseModel):
...
+    actor: str
+    instruction: str
+    title: str
+
+class BasePlan(BaseModel):
```
ahuang11: Same here, curious for the reason why it's prefixed Base

philippjfr: Good question, Base is maybe not accurate. Should probably just be StepType and PlanType since these are just type classes matching the dynamically created classes.

ahuang11: Or RawPlan?

PR #1321, opened 2025-07-18
Add SourceCatalog to pick and choose tables

ahuang11:
Adds a Table Catalog dialog icon to the top right:
image

This allows users to select the source tables (and in another PR, documents) they want.

Also, table/document list is grouped by source.

Lastly, unifies visible slugs in prompts and code.

Eventually, might combine the upload icon inside here and support a callback function to list remote catalogs.

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -323,8 +323,8 @@ class ListAgent(Agent):
 
     __abstract = True
 
-    def _get_items(self) -> list[str]:
-        """Return the list of items to display"""
+    def _get_items(self) -> dict[str, list[str]]:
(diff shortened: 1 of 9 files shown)


philippjfr:
The modal should have a title and subtitle explaining what this menu does.

ahuang11:
Okay, addressed that. and fixed other issues.

I don't really like the upload flow right now, but that'll be for a separate PR. Ideally, the chat area input can handle file uploads, and maybe get rid of SourceAgent.

https://github.com/user-attachments/assets/3e8f7d11-df22-44ee-93e4-2589b527d91e

droumis:
loving this addition

PR #1322, opened 2025-07-18
Reorg dialog and add source controls

ahuang11:
Builds upon https://github.com/holoviz/lumen/pull/1321 but wanted to create a new branch because the original was already getting big, and this was motivated by https://github.com/holoviz-topics/lumen-anndata/pull/20

Added source controls to the dialog and also cleaned up the controls significantly

### DIFF
--- a/lumen/ai/components.py
+++ b/lumen/ai/components.py
@@ -606,6 +606,7 @@ class SourceCatalog(Viewer):
             sources: Optional list of sources. If None, uses sources from memory.
         """
         sources = self.sources or memory.get('sources', [])
+
         # Create a lookup of existing cards by source
         existing_cards = {
             card.source: card for card in self._cards_column.objects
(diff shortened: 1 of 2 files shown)


lumen/ai/agents.py
```diff
@@ -183,26 +183,22 @@ class SourceAgent(Agent):
...
-        table alias, add or modify the `source` in memory, and return a bool
-        (True if the table was successfully uploaded).""")
-
     purpose = param.String(default="The SourceAgent allows a user to upload new datasets, tables, or documents.")
 
     requires = param.List(default=[], readonly=True)
 
     provides = param.List(default=["sources", "source", "document_sources"], readonly=True)
 
+    source_controls: ClassVar[SourceControls] = SourceControls
```
ahuang11, line 192: Thought it was really ugly to pass the source controls around as an instance var, so I made it a class var... I'd imagine most applications would want the same source controls?

PR #1343, opened 2025-07-28
LLM selection

ahuang11:
I wanted to add a "SQL Planning Toggle", but then adding another icon to the top right was confusing, and I think replacing the status chips with buttons makes sense to open different dialogs that can nest more options.

https://github.com/user-attachments/assets/48e8128c-c806-4619-a1a8-b2ab30d152dc

Maybe the report toggle can also be converted?


lumen/ai/llm_dialog.py
```diff
@@ -0,0 +1,433 @@
...
+This component provides a dialog for selecting LLM models and adjusting temperature.
+Similar to the SourceCatalog but for LLM management.
+"""
+
+import param
+
+from panel.layout import Column, HSpacer, Row
+from panel.pane import Markdown
+from panel.viewable import Viewer
+from panel.widgets import FloatSlider, Select
```
philippjfr: Why not a material UI widget?

lumen/ai/llm_dialog.py
```diff
@@ -0,0 +1,433 @@
...
+
+        # Watch for changes
+        self.model_select.param.watch(self._on_model_change, "value")
+
+    def _get_default_models(self):
+        """Get list of available models based on LLM provider type."""
+        llm_type = type(self.llm).__name__
+
+        # Default models based on LLM provider
+        model_options = {
```
philippjfr: Can we define these on the LLM classes themselves and then use param.concrete_descendants to detect the available LLMs?

philippjfr (approved):

PR #1353, opened 2025-07-31
Add toggle menu

ahuang11:
Consolidates multiple toggles into a single menu.

Also adds planning mode. Closes https://github.com/holoviz/lumen/issues/1334

image

https://github.com/user-attachments/assets/60908a88-3829-4bb4-87fc-ea5065ed2b59

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -794,8 +794,8 @@ class ExplorerUI(UI):
         ]
         self._report_toggle = None  # Keep for compatibility
 
-        # Override the settings click handler to include report mode logic
-        self._settings_menu.on_click = self._handle_explorer_settings_click
+        # Watch for toggle changes to handle report mode
+        self._settings_menu.param.watch(self._handle_explorer_settings_toggle, 'toggled')
 
         self._last_synced = self._home = Exploration(
             context=memory,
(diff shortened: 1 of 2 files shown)


lumen/ai/ui.py
```diff
@@ -319,12 +319,21 @@ def __init__(
...
+            color="primary",
+            margin=(0, 5, 0, 5),
+            align="center",
+            persistent=True,
+            on_click=self._handle_settings_click,
+            sx={
+                'borderRadius': '6px',  # Less rounded corners
+                'boxShadow': '0px 3px 5px -1px rgba(0,0,0,0.2)',  # Matching shadow
+                'width': '200px',  # Consistent width
+                'fontSize': '13px'  # Consistent font size
```
philippjfr: fontSize should be expressed in em so it scales if needed. Ideally we'd express boxShadow via the Mui shadow CSS variables and width should be set on the component itself.

philippjfr:
>Maybe just Toggles? I'm hesitant to create a separate toggle menu just for those. I guess Artifacts also don't fall under there.

I don't think they should be hidden away in a sub-menu. I think this is one of the core pieces we have to work out still. I'll iterate a bit here and merge but it's clear we need to revisit this piece.

philippjfr:
Played with this a bit, here's my thoughts:

- The toggles are not hooked up correctly, e.g. CoT toggle seems to trigger explorations.
- I'm not really sure what Artifacts is referring to (I guess the sidebar?!)
- The `Toggle` label is kind of meaningless, I'd probably just use variant='text' + CSS to make it appear as just gear icon.
- I still don't think Chat/Report, Explorations and Artifacts belong in here. A settings menu makes sense to toggle settings on/off not to access core functionality.

PR #1356, opened 2025-08-01
Support fully independent Azure llms

MarcSkovMadsen:
My azure llms are often running on separate endpoints and needs seperate api_versions.

This PR adds support for this use case.

### DIFF
--- a/lumen/tests/llm/test_llm.py
+++ b/lumen/tests/llm/test_llm.py
@@ -1,0 +1,58 @@
+"""Test suite for LLM implementations."""
+
+from lumen.ai.llm import AzureOpenAI
+
+
+async def test_azure_open_ai_get_model_kwargs():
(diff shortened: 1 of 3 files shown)


lumen/tests/llm/test_llm.py
```diff
@@ -0,0 +1,58 @@
+"""Test suite for LLM implementations."""
+
+from lumen.ai.llm import AzureOpenAI
```
ahuang11, line 3:
```suggestion
try:
    from lumen.ai.llm import AzureOpenAI # noqa
except ModuleNotFoundError:
    pytest.skip("lumen.ai could not be imported, skipping tests.", allow_module_level=True)
```

MarcSkovMadsen: Why is this needed?

philippjfr: Because we have tests for just lumen core.

PR #1362, opened 2025-08-07
Ensure mirrors don't erroneously get added into tables

ahuang11:
I noticed `mirrors` being erroneously added to the `tables`:
```
(Pdb) print(source.tables)
{'obs': 'SELECT * FROM obs', 'obsm_X_pca': 'SELECT * FROM obsm_X_pca', 'obsm_X_tsne': 'SELECT * FROM obsm_X_tsne', 'obsm_X_umap': 'SELECT * FROM obsm_X_umap', 'select_all_obs_from_AnnDataSource00863': 'SELECT * FROM "main"."obs" LIMIT 100000', 'var': 'SELECT * FROM var', 'varm_PCs': 'SELECT * FROM varm_PCs', 'mirrors': {}}  # <-- here
```

So when I was doing `source.create_sql_expr_source(source.tables)`, it was crashing (here, I manually set `mirrors: {}`, but in the code I was using `from_spec(to_spec())`
```
ParserException                           Traceback (most recent call last)
Cell In[7], line 4
      1 from lumen.sources.duckdb import DuckDBSource
      3 src = DuckDBSource(tables={"obs": "SELECT * FROM obs"})
----> 4 src.create_sql_expr_source(tables={"mirrors": {}})

File ~/repos/lumen/lumen/sources/duckdb.py:356, in create_sql_expr_source(self, tables, materialize, **kwargs)
    354 try:
    355     self._connection.execute(table_expr)
--> 356 except duckdb.CatalogException as e:
    357     original_e = e
    358     pattern = r"Table with name\s(\S+)"

ParserException: Parser Error: syntax error at or near "{"
```

And I was really confused how the code ran with a syntax error, then I realized it's actually the SQL...

I think the issue was that the `mirrors` was wrongly added to every recursive call, and so we should be adding it to the top level if I'm not mistaken--however, I'm not 100% sure if this is right.

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -191,20 +191,7 @@ class DuckDBSource(BaseSQLSource):
     def _recursive_resolve(
         cls, spec: dict[str, Any], source_type: type[Source]
     ) -> tuple[dict[str, Any], dict[str, Any]]:
-        mirrors = spec.pop('mirrors', {})
         spec, refs = super()._recursive_resolve(spec, source_type)
-        resolved_mirrors = {}
-        for table, mirror in mirrors.items():
-            if isinstance(mirror, tuple):
-                src_spec, src_table = mirror
-                source = cls.from_spec(src_spec)
-                resolved_mirrors[table] = (source, src_table)
-            elif mirror.get('type') == 'pipeline':
-                from ..pipeline import Pipeline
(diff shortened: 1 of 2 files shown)


ahuang11:
I think this got merged with https://github.com/holoviz/lumen/pull/1363

PR #1376, opened 2025-08-22
Switch OpenAI default to 4.1-mini

philippjfr:
The cost differences are not that large and 4.1-mini is both faster and better.


ahuang11:
It's technically slightly slower based on this latency vs intelligence graph here
https://openai.com/index/gpt-4-1/

The output token cost is significantly higher.
1.6 vs 0.6 so I would keep 4o mini for analyst.

https://platform.openai.com/docs/pricing

Also I just realized there's gpt-5-mini...

PR #1387, opened 2025-08-25
Add support for file upload via chat input

philippjfr:
Fixes handling of file upload via the chat input area. Previously the `on_submit` callback would be called long before the file uploads would complete. Instead we add an explicit listener for the `value_uploaded` parameter and then render a stripped down `SourceControls` and once fully uploaded we show the `TableSourceCard` corresponding to the newly added source.

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -234,7 +234,7 @@ class Coordinator(Viewer, VectorLookupToolUser):
                     for source in self._memory.get("sources", []) if source not in old_sources
                 ]
                 if len(source_cards) > 1:
-                    source_view = Accordion(*source_cards, sizing_mode="stretch_width")
+                    source_view = Accordion(*source_cards, sizing_mode="stretch_width", name="TableSourceCard")
                 else:
                     source_view = source_cards[0]
                 msg = Column(chat_input.value, source_view) if user_prompt else source_view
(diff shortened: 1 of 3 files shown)


lumen/ai/coordinator.py
```diff
@@ -210,33 +211,52 @@ def on_clear(instance, _):
...
+                respond=False
+            )
+
+            def display_uploaded(_):
+                source = self._memory["source"]
+                if old_source is not source:
+                    msg.object = Column(
+                        f"Successfully added {source.name}",
+                        TableSourceCard(source=source)
+                    )
```
ahuang11: Is it necessary that this is triple nested?

philippjfr: Could be rewritten, but as it is right now yes, it's a closure that accesses a variable from its enclosing scope.

ahuang11:
https://github.com/user-attachments/assets/5b31384b-9745-4ac6-a1e4-f388057eb459

It seems a tad awkward to have two additional clicks:
1. to confirm
2. to resend my original message

ahuang11:
The updates look awesome!

https://github.com/user-attachments/assets/3e93cacc-e87b-43f8-90df-c2c756f7898f

One minor, confusing issue is if I don't type anything to text area and send, I have to type something and press enter before it proceeds:

https://github.com/user-attachments/assets/fcffdff9-024b-4177-8599-52bd611270ca

philippjfr:
>One minor, confusing issue is if I don't type anything to text area and send, I have to type something and press enter before it proceeds:

Ah, indeed, should still move you into the full chat interface.

philippjfr:
>One minor, confusing issue is if I don't type anything to text area and send, I have to type something and press enter before it proceeds:

Actually, not sure if this is accurate, it would proceed without the second send, but I guess the question is whether you should be able to provide text while it's processing the upload.

ahuang11:
Okay using latest panel-material-ui makes the flow go directly.

One other minor thing is:
`No specific converter available for 'windturbines.parquet' of mime type ''.` can be confusing to the user, and looks like an error. We should instead show "Cannot display contents of ..."

image

ahuang11 (approved):

PR #1388, opened 2025-08-25
Better logfire integration

ahuang11:
I'm having trouble figuring out the ongoing bug, but this sets the stage for recording duration and token usage of each agent.

image

image

```python
import json
from io import StringIO
import pandas as pd
from logfire.query_client import AsyncLogfireQueryClient


pd.set_option("display.max_columns", None, "display.max_rows", None)


query = """
SELECT message, attributes
FROM records
LIMIT 10
"""

async with AsyncLogfireQueryClient(read_token=read_token) as client:
    df = pd.read_csv(StringIO(await client.query_csv(sql=query)))
df = df.join(df["attributes"].apply(json.loads).apply(pd.Series)).drop("attributes", axis=1)
```

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -54,8 +54,14 @@ class Llm(param.Parameterized):
         'default', 'reasoning' and 'sql'. Agents may pick which model to
         invoke for different reasons.""")
 
+    use_logfire = param.Boolean(default=True, doc="""
+        Whether to log LLM calls and responses to logfire.""")
+
     _ready = param.Boolean(default=False, doc="""
         Whether the LLM has been initialized and is ready to use.""")
+
+    # Whether the LLM supports logging to logfire
+    _supports_logfire = False
 
     # Whether the LLM supports streaming of any kind
     _supports_stream = True
@@ -70,6 +76,10 @@ class Llm(param.Parameterized):
             if isinstance(params["mode"], str):
                 params["mode"] = Mode[params["mode"].upper()]
         super().__init__(**params)
+        if self.use_logfire and not self._supports_logfire:
+            raise ValueError(
+                f"LLM {self.__class__.__name__} does not support logfire."
+            )
         if not self.model_kwargs.get("default"):
             raise ValueError(
                 f"Please specify a 'default' model in the model_kwargs "
@@ -167,7 +177,7 @@ class Llm(param.Parameterized):
         try:
             self._ready = False
             await self.invoke(
-                messages=[{'role': 'user', 'content': 'Are you there? YES | NO'}],
+                messages=[{'role': 'user', 'content': 'Init?'}],
                 model_spec="ui",
                 response_model=YesNo
             )
@@ -381,11 +391,13 @@ class OpenAI(Llm, OpenAIMixin):
         "sql": {"model": "gpt-4.1-mini"},
         "vega_lite": {"model": "gpt-4.1-mini"},
         "edit": {"model": "gpt-4.1-mini"},
+        "ui": {"model": "gpt-4.1-nano"},
     })
 
     select_models = param.List(default=[
-        "gpt-3.5-turbo",
         "gpt-4.1-mini",
+        "gpt-4.1-nano",
+        "gpt-5-mini",
         "gpt-4-turbo",
         "gpt-4o",
         "gpt-4o-mini"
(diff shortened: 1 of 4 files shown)


lumen/ai/llm.py
```diff
@@ -54,9 +54,15 @@ class Llm(param.Parameterized):
         'default', 'reasoning' and 'sql'. Agents may pick which model to
         invoke for different reasons.""")
 
+    use_logfire = param.Boolean(default=True, doc="""
```
ahuang11:
```suggestion
    use_logfire = param.Boolean(default=False, doc="""
```

ahuang11:
The LLM is set up correctly...
```python
from lumen.ai.llm import OpenAI
from pydantic import BaseModel, Field


class Translations(BaseModel):
    en: str = Field(..., description="English translation")
    es: str = Field(..., description="Spanish translation")
    fr: str = Field(..., description="French translation")
    ch: str = Field(..., description="Chinese translation")


llm = OpenAI(use_logfire=True)
async for response in llm.stream(messages=[{"role": "user", "content": "Hello!"}], response_model=Translations):
    print(response)
```

lumen/ai/utils.py
```diff
@@ -851,3 +853,14 @@ def class_name_to_llm_spec_key(class_name: str) -> str:
         i += 1
 
     return result
+
+
+def wrap_logfire(func: Callable, llm: Llm, span_name: str | None = None, extract_args: bool = True, **instrument_kwargs):
```
philippjfr: Not sure I understand why it's called `span_name`.

ahuang11: It is logfire's function arg naming
https://logfire.pydantic.dev/docs/reference/api/logfire/#logfire.Logfire.instrument

ahuang11:
I suppose we can implement it as a decorator. The reason I did not is because I wanted to target specific functions, like this line `task.respond`:
https://github.com/holoviz/lumen/pull/1388/files#diff-1540e8b3b2b09622911975e69245bf643bd629b9e9e43cdc86b674e534671c61R108

ahuang11:
Okay I used decorator instead; a bit more complex, but less messy I guess

lumen/ai/llm.py
```diff
@@ -167,9 +178,8 @@ async def initialize(self, log_level: str):
         try:
             self._ready = False
             await self.invoke(
-                messages=[{'role': 'user', 'content': 'Are you there? YES | NO'}],
+                messages=[{'role': 'user', 'content': 'Ready? "Y" or "N"'}],
```
ahuang11, line 182: Having an unnecessary response model used 70 tokens instead of just 15 tokens.

lumen/ai/llm.py
```diff
@@ -206,6 +216,21 @@ async def stream(
...
         """
+        if self.use_logfire:
+            output = await self.invoke(
+                messages,
+                system=system,
+                response_model=response_model,
+                model_spec=model_spec,
+                **kwargs,
+            )
+            if field is not None and hasattr(output, field):
```
ahuang11, line 228: Need to mimic streaming, without actually streaming because logfire doesn't count tokens if streaming

lumen/ai/ui.py
```diff
@@ -474,14 +474,6 @@ async def _initialize_new_llm(self):
             import traceback
             traceback.print_exc()
 
-    async def _setup_llm_and_watchers(self):
```
ahuang11, line 477: This method is equivalent to _initialize_new_llm AFAIK

ahuang11:
image

ahuang11:
Okay I added a script to parse logfire exports which creates this table

image

messages is what we input, while request_data is the modified input with instructor

image

In the next PR, I'll add a way to use `request_data` to do evals. To hold everything else constant, I think we'll need to just modify the system prompt from each request, or the tools if the response model changes.

Lastly, I discovered that the chat history was not serializing properly (included Panel object params...)

```
{'role': 'system',
   'content': "<Chat History>\nUser: What data is available?\n\nAssistant: The available datas are listed below. Click on the eye icon to show the data contents.Tabs(design=<class 'panel_material_ui...., sizing_mode='stretch_width', stylesheets=['[https://cdn.holoviz.org/...])\n](https://cdn.holoviz.org/...])/n)    [0] Tabulator(buttons={'show': '<i c...}, design=<class...
```

philippjfr (approved):

PR #1392, opened 2025-08-26
Should not depend on interface

ahuang11:
```python
f"User Request: {user_query['content']!r}\n\nComplete the next unchecked todo:\n{todos}" ~~~~~~~~~~^^^^^^^^^^^  TypeError: 'NoneType' object is not subscriptable
```

user_query was None because `messages == []` because Planner was depending on a message on an interface, but I bypassed that step with:
```python

planner = Planner(agents=[agent], llm=llm, interface=interface)
await planner.respond([{"role": "user", "content": "find first row in oni"}])
```

### DIFF
--- a/lumen/ai/coordinator.py
+++ b/lumen/ai/coordinator.py
@@ -649,7 +649,7 @@ class Coordinator(Viewer, VectorLookupToolUser):
 
             # TODO INVESTIGATE
             messages = fuse_messages(
-                self.interface.serialize(custom_serializer=self._serialize, limit=10),
+                self.interface.serialize(custom_serializer=self._serialize, limit=10) or messages,
                 max_user_messages=self.history
             )
 


lumen/ai/coordinator.py
```diff
@@ -649,7 +649,7 @@ async def respond(self, messages: list[Message], **kwargs: dict[str, Any]) -> st
 
             # TODO INVESTIGATE
             messages = fuse_messages(
-                self.interface.serialize(custom_serializer=self._serialize, limit=10),
+                self.interface.serialize(custom_serializer=self._serialize, limit=10) or messages,
```
philippjfr, line 652: Do you have to call fuse_messages at all if there's no messages to fuse?

ahuang11: You can input messages with length greater than 1 if I remember what fuse messages does correctly.

ahuang11: I technically can pass:
`planner.respond([{"role": "user", "content": "message1"}, {"role": "assistant", "content": "message2"}, {"role": "user", "content": "find first row in oni"}])` so I believe this is correct.

```
def fuse_messages(messages: list[dict], max_user_messages: int = 2) -> list[dict]:
    """
    Fuses the chat history into a single system message, followed by the last user message.
    This function is reusable across different components that need to combine
    multiple messages into a simplified format for LLM processing.

    Parameters
    ----------
    messages : list[dict]
        List of message dictionaries with 'role' and 'content' keys
    max_user_messages : int
        Maximum number of user messages to include in the history

    Returns
    -------
    list[dict]
        Processed messages with the chat history as a system message
    """
    user_indices = [i for i, msg in enumerate(messages) if msg['role'] == 'user']
    if user_indices:
        first_user_index = user_indices[0] if len(user_indices) <= max_user_messages else user_indices[-max_user_messages]
        last_user_index = user_indices[-1]
        last_user_message = messages[last_user_index]
    else:
        first_user_index = 0
        last_user_index = -1
    history_messages = messages[first_user_index:last_user_index]
    if not history_messages:
        return [last_user_message] if user_indices else []
    formatted_history = "\n\n".join(
        f"{msg['role'].capitalize()}: {msg['content']}" for msg in history_messages if msg['content'].strip()
    )
    system_prompt = {
        "role": "system",
        "content": f"<Chat History>\n{formatted_history}\n<\\Chat History>"
    }
    return [system_prompt] if last_user_index == -1 else [system_prompt, last_user_message]
```

PR #1403, opened 2025-09-08
Fix empty df  due to duckdb not being thread-safe

ahuang11:
Earlier, we saw that `df.empty is True` and we suspected it was due to SQLLimit, but even after disabling that, it was still failing. I've concluded that it's related to duckdb and race conditions or something with pytest starting multiple threads and fetch was not pointed to the right connection?

I've also attempted different variations of `self._connection.sql()` and `self._connection.query()` which result in:

```
Attempting to execute an unsuccessful or closed pending query result
```

From https://github.com/duckdb/duckdb/issues/13498#issuecomment-2540940925
```
duckdb.sql() is not thread safe!
To call .sql() in concurrent threads, create a new connection first and call con.sql() on it, like in the function good()
```

From https://duckdb.org/docs/stable/clients/python/dbapi.html
Connections are closed implicitly when they go out of scope or if they are explicitly closed using close(). Once the last connection to a database instance is closed, the database instance is closed as well.

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -401,6 +401,12 @@ class DuckDBSource(BaseSQLSource):
         rel = self._connection.execute(sql_expr)
         has_geom = any(d[0] == 'geometry' and d[1] == 'BINARY' for d in rel.description)
         df = rel.fetch_df(date_as_object=True)
+
+        if df.empty:
+            rel = self._connection.execute(sql_expr)
+            has_geom = any(d[0] == 'geometry' and d[1] == 'BINARY' for d in rel.description)
+            df = rel.fetch_df(date_as_object=True)
+
         if has_geom:
             import geopandas as gpd
             geom = self._connection.execute(
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -163,10 +163,6 @@ class DuckDBSource(BaseSQLSource):
 
         return self._local.connection
 
-    @property
-    def connection(self):
-        return self._connection
-
     def _setup_mirrors(self):
         """Setup mirrors in the shared database (called once)."""
         for table, mirror in self.mirrors.items():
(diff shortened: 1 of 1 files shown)


lumen/sources/duckdb.py
```diff
@@ -134,6 +136,35 @@ def __init__(self, **params):
...
+
+    @property
+    def _connection(self):
+        """Thread-safe connection property - each thread gets its own connection to shared DB."""
+        if not hasattr(self._local, 'connection') or self._local.connection is None:
+            # Use shared memory database if uri is :memory:
+            # This allows multiple connections to share the same database
+            uri = self.uri
+            if uri == ':memory:':
+                uri = f':memory:shared_db_{id(self)}'
```
philippjfr, line 151: I don't think this is appropriate. This means that two instances won't share the same DB, which is expected and required, e.g. when we call `create_sql_expr_source(...)`

lumen/sources/duckdb.py
```diff
@@ -134,6 +136,35 @@ def __init__(self, **params):
 
             self.tables = processed_tables
 
+        # Setup mirrors on first connection access
+        if self.mirrors:
+            _ = self._connection  # This will trigger connection creation and mirror setup
```
philippjfr, line 141: Why even assign it to a variable? I think it's also cleaner to move the initialization logic into a method and call that here (and in the `_connection` property).

lumen/sources/duckdb.py
```diff
@@ -450,6 +483,6 @@ def close(self):
         This method should be called when the source is no longer needed to prevent
         connection leaks and properly clean up server-side resources.
         """
-        if self._connection is not None:
-            self._connection.close()
-            self._connection = None
+        if hasattr(self._local, 'connection') and self._local.connection is not None:
```
philippjfr, line 486: Surely this should clean up the connection for all threads, not just the local thread.

philippjfr:
I don't think this approach addresses the issue and causes various other issues. If you read the docs you find out that DuckDB connections are, in fact, thread-safe, specifically read [this](https://duckdb.org/docs/stable/clients/python/dbapi.html):

> If you want to create a second connection to an existing database, you can use the cursor() method. This might be useful for example to allow parallel threads running queries independently. A single connection is thread-safe but is locked for the duration of the queries, effectively serializing database access in this case.

So clearly thread-safety shouldn't be at fault here at all. That said, it's perhaps still not a bad idea to allow creating new connections per-thread so we can parallelize reads properly. However that's independent from the issue you're trying to fix here.

ahuang11:
Okay I've tested the cursors and they work great!

PR #1406, opened 2025-09-08
Reinforce in SQL model to not apply limits unless specified

ahuang11:
In the system prompt, we specify:
```
5. No LIMIT clauses unless explicitly requested by the user - Pagination is automatically handled
```

However, I still see it LIMIT 10 probably because those instructions are up top--so added some prompt reinforcement in the model.


ahuang11:
I think LIMITs are due to https://github.com/holoviz/lumen/pull/1411

PR #1408, opened 2025-09-09
Fallback agent

ahuang11:
Supposedly, user asks "Visualize the data"
1. the previous agent (e.g. SQLAgent) provides an insufficient response (one column / value) answer
2. on the next agent (e.g. VegaLiteAgent), the `insufficient_context`* fields populates
3. that raises MissingContextError on validation
4. re-invoke fallback_agent.respond with mutated user messages containing the error feedback from the original agent like `"The dataset contains only a single row with one categorical value 'LLM' and no quantitative or temporal data to visualize."`
5. the new agent is happy with the fallback agent's response
6. user is happy

However, I cannot get to step 5 and 6 :'( because I can't get the memory injected properly into the fallback_agent...

Here VegaLiteAgent tries to call its fallback_agent (SQLAgent), but it encounters this
```python

  File "/Users/ahuang/repos/lumen/lumen/ai/prompts/SQLAgent/generate_roadmap.jinja2", line 30, in block 'context'
    {{ memory["sql_metaset"].compact_context }}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/jinja2/environment.py", line 490, in getattr
    return getattr(obj, attribute)
           ^^^^^^^^^^^^^^^^^^^^^^^
jinja2.exceptions.UndefinedError: 'lumen.ai.memory._Memory object' has no attribute 'sql_metaset'
Traceback (most recent call last):
  File "/Users/ahuang/repos/lumen/lumen/ai/report.py", line 149, in execute
    outputs += await self._run_task(i, task, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahuang/repos/lumen/lumen/ai/coordinator.py", line 103, in _run_task
    raise RuntimeError(f"{task.title} failed to provide declared context.")
```

I don't know how to get previous task's memory injected to the new task's fallback memory.


lumen/ai/report.py
```diff
@@ -107,6 +107,9 @@ async def _run_task(self, i: int, task: Self | Actor, **kwargs):
         memory = task.memory or self.memory
         messages = self.history
 
+        if getattr(task, "fallback_agent", None):
+            task.fallback_agent.memory = memory
```
ahuang11: My attempt at injecting memory, which fails

ahuang11:
> The current approach once again creates coupling between agents

Sure, but my main concern is:

> find the agent that provided that piece of context

this sounds like rerunning the whole planner again, which can be slow and costly. 

I'm also imagining a scenario where the plan is Planner > IterativeTableLookup > SQLAgent > ChatAgent, and the user requests a column that doesn't exist in the data tables. With the current implementation it, would "fallback" to ChatAgent, even though ChatAgent is planned after SQLAgent, or do you think it should ONLY rerun previously run agents?

lumen/ai/coordinator.py
```diff
@@ -90,13 +91,35 @@ async def _run_task(self, i: int, task: Self | Actor, **kwargs):
...
+                        traceback.print_exception(e)
+                        self.memory['__error__'] = str(e)
+                        raise e
+                elif isinstance(e, asyncio.CancelledError):
+                    step.failed_title = f"{task.title} agent was cancelled"
+                    raise e
+                else:
+                    traceback.print_exception(e)
+                    self.memory['__error__'] = str(e)
+                    raise e
```
philippjfr: Let's move this into a new method.

lumen/ai/coordinator.py
```diff
@@ -106,6 +129,70 @@ async def _run_task(self, i: int, task: Self | Actor, **kwargs):
...
+        """
+        Find the task that provided the pipeline or other relevant context.
+        Search backwards from the failed task.
+        """
+        pipeline_exists = 'pipeline' in self.memory
+        if not pipeline_exists:
+            return
+
+        for idx in range(failed_index - 1, -1, -1):
+            task = self.subtasks[idx]
```
philippjfr, line 142:
```suggestion
        for idx in reverse(enumerate(self.subtasks)):
            task = self.subtasks[idx]
```

lumen/ai/coordinator.py
```diff
@@ -106,6 +129,70 @@ async def _run_task(self, i: int, task: Self | Actor, **kwargs):
...
+    def _find_context_provider(self, failed_index: int) -> int | None:
+        """
+        Find the task that provided the pipeline or other relevant context.
+        Search backwards from the failed task.
+        """
+        pipeline_exists = 'pipeline' in self.memory
+        if not pipeline_exists:
+            return
+
+        for idx in range(failed_index - 1, -1, -1):
```
philippjfr:
```suggestion
        for idx in reversed(range(failed_index)):
```

PR #1409, opened 2025-09-09
Set nodefaults and add lower pins

maximlt:
This deployment keeps on failing, here's another attempt at fixing it:
- Set `nodefaults` to only pull packages from pyviz/label/dev and conda-forge
- Add lower pins to constrain the solver


philippjfr:
Thanks folks!

PR #1410, opened 2025-09-09
Use duckdb cursors

ahuang11:
Earlier, we saw that df.empty is True and we suspected it was due to SQLLimit, but even after disabling that, it was still failing. I've concluded that it's related to duckdb.

Supersedes https://github.com/holoviz/lumen/pull/1403 based on https://github.com/holoviz/lumen/pull/1403#issuecomment-3269554816


ahuang11:
Yes it addresses the problem and ready to merge if you're happy with it.

Maybe related:
https://github.com/duckdb/duckdb/discussions/12817#discussioncomment-10945893

philippjfr (approved):

Issue #1417, opened 2025-09-15
Regression: cannot change spec in editor once rendered


ahuang11:
MRVE:
```python
from lumen.ai.memory import memory
from lumen.sources.duckdb import DuckDBSource

source = memory["source"] = DuckDBSource(tables={"gpt_41_mini_csv": "gpt-4o-mini.csv"})
source.from_spec(source.create_sql_expr_source(
    tables={"test_get": 'SELECT "input_tokens", "output_tokens", "model_type", "api", "response_time" FROM "gpt_41_mini_csv"'}
).to_spec()).get("test_get")
```
Connection loses the table

Issue #1415, opened 2025-09-10
Download run does not work


ahuang11:
I tried downloading the chat, and because we standardize the naming somewhere:
```
CatalogException: Catalog Error: Table with name gpt_4o_mini_csv does not exist!
Did you mean "pg_proc"?

LINE 1: ...("input_tokens" + "output_tokens") AS "avg_total_tokens" FROM "gpt_4o_mini_csv" WHERE "model_type" = 'LLM') LIMIT 1
```

PR #1418, opened 2025-09-15
Fix duckdb from spec cannot find file-based tables

ahuang11:
Closes https://github.com/holoviz/lumen/issues/1417 and https://github.com/holoviz/lumen/issues/1415

I noticed this issue when I tried to change the spec:
image

Specifically, when I was trying to create https://github.com/holoviz/lumen/compare/main...iterative_vega_lite

I removed SQLOutput because there wasn't a way to store references to the previous tables with just showing the SQL statement, especially when serializing to notebook--also this way is more consistent with VegaLiteAgent

image


philippjfr:
Nvm, just didn't have the referenced CSV.

PR #1423, opened 2025-09-19
Sql discovery

ahuang11:
I noticed that the previous generate roadmap -> plan next step was consuming excessive tokens because it was not realizing that it was re-doing the same query over and over and over, and this burned through 60k tokens 😨

image

To limit the LLM from doing too much random things:
1. I decided to create a DiscoveryQueries model that only allows the model to do one or more of these queries: `list[SampleQuery | DistinctQuery]`
2. No more excessive retries--just one discovery sufficiency test in between.

Now, the SQLAgent flow is:
1. One shot--
2. If it fails or is empty, start the discovery phase
3. The LLM chooses to use the distinct or sample model and executes with those fields populated
4. DiscoverySufficiency is invoked to see if the discovery results are enough to answer the query--if not, redo it one more time with this context
5. Answer the question with the discovered info

Output example:
1.
`SELECT "athlete_full_name", COUNT(*) AS "gold_medal_count"\nFROM "olympic_medals_csv"\nWHERE "country_name" = \'China\'\n  AND "medal_type" = \'GOLD\'\n  AND "participant_type" = \'Athlete\'\nGROUP BY "athlete_full_name"\nORDER BY "gold_medal_count" DESC\nLIMIT 1' expr_slug='top_chinese_athlete_most_gold_medals_olympic_medals_csv`

2. empty df so start discovery

3.
```
LLM Response: reasoning='Check the distinct values of the country_name column in olympic_medals_csv to confirm the exact naming of China for filtering.' queries=[DistinctQuery(query="SELECT DISTINCT {column} FROM {table} WHERE {column} ILIKE '%{pattern}%' LIMIT 10 OFFSET {offset}", table='olympic_medals_csv', column='country_name', pattern='chin', offset```

4.
```
Discovery Results:

DistinctQuery: Distinct country_name: `['Chinese Taipei', "People's Republic of China", 'Hong Kong, China']`

Your task: Determine if these discoveries provide enough context to fix the original error, or if specific follow-ups are needed.

Complete the underlined todo:
- [x] Get the SQL metadata for olympic_athletes_csv and olympic_medals_csv tables to understand their schema.
- [ ] <u>Using the metadata, write and execute a SQL query to find the Chinese athlete who won the most gold medals by joining olympic_athletes_csv and olympic_medals_csv on athlete ID and filtering for Chinese nationality and gold medals.</u>
- [ ] Validate whether the executed plan fully answered the user's original query.

LLM Response: reasoning='The original query filtered on country_name = \'China\', but the distinct values for country_name in the dataset are [\'Chinese Taipei\', "People\'s Republic of China", \'Hong Kong, China\']. This indicates that the exact filter value \'China\' does not exist in the data. Instead, the correct value to filter for Chinese athletes is likely "People\'s Republic of China". This discovery provides a clear fix for the original query\'s filter condition to get results. Therefore, the discovery is sufficient to fix the original error by replacing \'China\' with "People\'s Republic of China" in the WHERE clause.' sufficient=True follow_up_needed=[]
```

5. ```
2025-09-19 16:42:27,119 LLM Response: query='SELECT "athlete_full_name", COUNT(*) AS "gold_medal_count"\nFROM "olympic_medals_csv"\nWHERE "country_name" IN (\'People\'\'s Republic of China\', \'Chinese Taipei\', \'Hong Kong, China\')\n  AND "medal_type" = \'GOLD\'\n  AND "participant_type" = \'Athlete\'\nGROUP BY "athlete_full_name"\nORDER BY "gold_medal_count" DESC\nLIMIT 1' expr_slug='top_chinese_athlete_most_gold_medals_olympic_medals'
```

### DIFF
--- a/lumen/ai/agents.py
+++ b/lumen/ai/agents.py
@@ -761,26 +761,63 @@ class SQLAgent(LumenBaseAgent):
             title=step_title
         )
 
-    async def _attempt_oneshot_sql(
+    async def _render_execute_query(
         self,
         messages: list[Message],
         source: Source,
-        step_title: str | None = None,
-    ) -> Pipeline | None:
-        """Attempt one-shot SQL generation without exploration context."""
-        with self._add_step(
-            title="Attempting one-shot SQL generation...",
-            steps_layout=self._steps_layout,
-            status="running"
-        ) as step:
-            # Generate SQL without exploration context
-            sql_queries = await self._generate_sql_queries(
-                messages, source.dialect, step_number=1,
-                is_final=True, context_entries=None
+        step_title: str,
+        success_message: str,
+        discovery_context: str | None = None,
+        raise_if_empty: bool = False,
+    ) -> Pipeline:
+        """
+        Helper method that generates, validates, and executes final SQL queries.
+
+        Args:
+            messages: List of user messages
+            source: Data source to execute against
+            step_title: Title for the execution step
+            success_message: Message to show on successful completion
+            discovery_context: Optional discovery context to include in prompt
+            raise_if_empty: Whether to raise error if query returns empty results
+
+        Returns:
+            Pipeline object from successful execution
+        """
+        with self._add_step(title=step_title, steps_layout=self._steps_layout) as step:
+            # Generate SQL using common prompt pattern
+            system_prompt = await self._render_prompt(
+                "main",
+                messages,
+                dialect=source.dialect,
+                step_number=1,
+                is_final_step=True,
+                current_step="",
+                sql_query_history={},
+                current_iteration=1,
+                sql_plan_context=None,
+                errors=None,
+                discovery_context=discovery_context,
             )
 
-            # Validate and execute the first query
-            expr_slug, sql_query = next(iter(sql_queries.items()))
+            # Generate SQL using existing model
+            model_spec = self.prompts["main"].get("llm_spec", self.llm_spec_key)
+            sql_response_model = self._get_model("main", is_final=True)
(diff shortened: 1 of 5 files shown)


lumen/ai/agents.py
```diff
@@ -828,270 +834,69 @@ async def _finalize_execution(
...
-
-    async def _generate_roadmap(self, messages: list[Message]) -> SQLRoadmap:
-        """Generate initial execution roadmap."""
-        with self._add_step(title="Generating execution roadmap", steps_layout=self._steps_layout) as step:
-            # Analyze query for characteristics
-            query_content = messages[0].get("content", "") if messages else ""
-            query_type = "complex" if any(word in query_content.lower() for word in ["compare", "vs", "between", "join"]) else "simple"
-            requires_joins = "join" in query_content.lower() or "vs" in query_content.lower()
-            has_temporal = any(word in query_content.lower() for word in ["date", "time", "year", "month", "season", "period"])
+        Args:
```
philippjfr: Wrong format

lumen/ai/agents.py
```diff
@@ -828,270 +834,69 @@ async def _finalize_execution(
         # Render output
         self._render_lumen(
             pipeline,
-            spec=result["sql"],
             messages=messages,
             title=step_title
```
philippjfr, line 858: Previously this would be an informative title, now it's hard-coded "Attempting one-shot SQL generation".

philippjfr:
I still do not understand why we switched from SQLOutput to LumenOutput here.

PR #1433, opened 2025-10-06
Add sqlglotrs dep

ahuang11:
It's much more efficient and it doesn't seem to be a large dep.


philippjfr:
Not opposed but is `sqlglot` really likely to be any kind of bottleneck?

PR #1434, opened 2025-10-06
Add xarray-sql support

ahuang11:
Just a proof of concept integrating https://github.com/alxmrs/xarray-sql 

```
import xarray as xr
ds = xr.tutorial.open_dataset('air_temperature')
ds.to_netcdf('air.nc')
```

`lumen-ai serve air.nc`

image
image
image

I wonder if this should be a lumen extension, like lumen-anndata. Part of me feels like it should be in core, so it works like hvplot.xarray, but at the same time, unsure about the limitations about the SQL & xarray.

I think if xarray-sql supported xarray-datatrees, then it's possible to rebuild AnndataSource on this since xarray is a generalized library supporting n-D dimension datasets.

### DIFF
--- a/lumen/ai/controls.py
+++ b/lumen/ai/controls.py
@@ -22,7 +22,13 @@ from ..sources.duckdb import DuckDBSource
 from ..util import detect_file_encoding
 from .memory import _Memory, memory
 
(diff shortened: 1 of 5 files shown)


droumis:
verrrry interesting.. It would be worthwhile to scope out how well it could support anndata

PR #1439, opened 2025-10-14
Enables parameterized queries using ? placeholders

ahuang11:
In SQL, it's good practice to use placeholders instead of string formatting to prevent SQL injections so here I added table_params to BaseSQLSource and params kwarg to create_sql_expr_source().

The use case is so users don't have to do this:
```python
from datetime import date
start_date = date(2024, 1, 1)
end_date = date(2024, 2, 1)
source.create_sql_expr_source(
    tables={
        'filtered': f"SELECT * FROM orders WHERE order_date >= '{start_date}' AND order_date < '{end_date}'"
    }
)
```
1. users would have to manually quote the inputs
2. risk of SQL injection

The following is easier, safer, and it handles type conversion automatically:

```python
source.create_sql_expr_source(
    tables={
        'filtered': 'SELECT * FROM orders WHERE order_date >= ? AND order_date < ?'
    },
    params={
        'filtered': [start_date, end_date]
    }
)
```

### DIFF
--- a/lumen/tests/sources/test_duckdb.py
+++ b/lumen/tests/sources/test_duckdb.py
@@ -730,3 +730,229 @@ def test_detour_roundtrip(sample_csv_files):
     assert read_df.iloc[[0]].equals(df.iloc[[0]])
     assert read_source.tables["limited_customers"] == 'SELECT * FROM customers LIMIT 1'
     assert "customers" in read_source.tables
+
+
+def test_table_params_basic(sample_csv_files):
+    """Test table_params with single and multiple parameters."""
+    files = sample_csv_files
+    original_cwd = os.getcwd()
+
+    try:
+        os.chdir(files['dir'])
+        source = DuckDBSource(
+            uri=':memory:',
+            tables={
+                'customers': 'customers.csv',
+                'orders': 'orders.csv',
+                'by_name': 'SELECT * FROM customers WHERE name = ?',
+                'by_customer_total': 'SELECT * FROM orders WHERE customer_id = ? AND total > ?'
+            },
+            table_params={
+                'by_name': ['Alice'],
+                'by_customer_total': [1, 200]
+            }
+        )
+
+        # Single parameter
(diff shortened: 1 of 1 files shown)


lumen/sources/bigquery.py
```diff
@@ -181,12 +181,33 @@ def get_sql_expr(self, table: str) -> str:
...
+        source: BigQuerySource
+        """
+        if params is None:
+            params = {}
+        source_params = dict(self.param.values(), **kwargs)
+        source_params.pop("name", None)
+        source_params["tables"] = tables
+        if params:
+            source_params["table_params"] = params
+        return type(self)(**source_params)
```
philippjfr, line 318: Does anything in BigQuerySource actually use the table_params?

lumen/ai/report.py
```diff
@@ -792,7 +796,10 @@ async def _execute(self, **kwargs):
             source = self.memory['source']
         if not self.table:
             raise ValueError("SQLQuery must declare a table name.")
-        source = source.create_sql_expr_source({self.table: self.sql_expr})
+
+        # Pass sql_params if provided
+        params = {self.table: self.sql_params} if self.sql_params else None
+        source = source.create_sql_expr_source({self.table: self.sql_expr}, params=params)
```
ahuang11:
```suggestion
    table_params = param.List(default=[], doc="""
        List of parameters to pass to the SQL expression.
        Parameters are used with placeholders (?) in the SQL expression.""")

table = param.String(doc="""
       The name of the table generated from the SQL expression.""")

@@ -792,7 +796,10 @@ async def _execute(self, **kwargs):
source = self.memory['source']
if not self.table:
raise ValueError("SQLQuery must declare a table name.")

        # Pass sql_params if provided
        params = {self.table: self.table_params} if self.table_params else None
        source = source.create_sql_expr_source({self.table: self.sql_expr}, params=params)
```

philippjfr (approved):

Issue #1440, opened 2025-10-14
Redesign Actor Context Handling & I/O Validation


philippjfr:
**Summary**
The current `Actor` architecture (covering `Agent` and `Tool` subclasses) relies on a shared `memory` object that is passed around and mutated by each actor. This approach has made context handling opaque, validation difficult, and debugging complex.

We propose a redesign that introduces **typed input/output models** and **explicit context passing** to improve traceability, composability, and correctness.

---

## Current Design

* Each `Actor` (Agent or Tool) declares what context keys it *requires* and *provides*.
* The system maintains a global `memory` (a dict-like object) that gets passed through and mutated as the plan executes.
* Each `Actor` reads its required keys from memory and writes new data under its provided keys.

**Problems**

* **Implicit data flow:** Hard to reason about what inputs and outputs exist at any point.
* **No type validation:** Inputs/outputs are loosely typed; bugs surface late.
* **Hidden side effects:** Memory mutations are implicit, making debugging and reproducibility difficult.
* **Limited introspection:** It’s unclear from code what structure an Actor expects or returns.
* **Difficult testing:** Unit tests require manual construction of partial memory dicts.

---

## Proposed Redesign

Each `Actor` defines:

```python
class MyInput(Context):
    # define required fields (allowing extras at runtime)
    user_query: str
    history: list[str] = []

class MyOutput(Context):
    answer: str
    metadata: dict[str, Any] = {}

class MyActor(Actor):
    # Configure which Context subclasses this actor expects/produces
    input  = param.ClassSelector(class_=Context, default=MyInput)
    output = param.ClassSelector(class_=Context, default=MyOutput)

    def respond(self, messages: list[Message], context: MyInput, **kwargs) -> tuple[list[Any], MyOutput]:
        # runtime: assert context ⊇ fields(MyInput)
        ...
        return outputs, context_out  # context_out ⊇ fields(MyOutput)
```

### Key points

- **No global memory**: context is passed in and returned.
- **Superset rule**: The context arg must include all fields of Input, but may include extra keys.
- **Outputs**: First element is list[Any] (arbitrary actor outputs).
- Second element is an Output context model instance, which may carry extra keys.
- **Validation helpers**: We’ll add runtime checks (e.g., validate_superset(Input, context)) and coercion (coerce_to_model(Output, mapping, allow_extra=True)).

---

### Expected Benefits

| Pain                 | Resolution                                               |
| -------------------- | -------------------------------------------------------- |
[...]

PR #1458, opened 2025-10-21
Define I/O interfaces of agents and remove global state

philippjfr:
Massive work in progress based on https://github.com/holoviz/lumen/issues/1440

- [x] Ensure SourceControls uploaded files still propagate to the global context object and visible_slugs is configured correctly 
- [x] Ensure that everything that was previously run via memory callbacks is now run when needed
- [x] Re-enable progressive rendering of outputs (right now they added altogether)
- [x] Decide whether to add runtime type validation
- [x] Compute execution graph and handle re-runs/edits by re-executing required graph edges.
- [x] Finalize design of context management
  - [x] Clarify how and when to merge output contexts when handing over from one TaskGroup to another
  - [x] Review how UI manages the context state when branching of new explorations
- [ ] Write up docs on I/O models


lumen/ai/controls.py
```diff
@@ -687,7 +684,7 @@ def add_medias(self):
                     n_docs += self._add_document(media_controls.file_obj, media_controls)
 
             if self.replace_controls:
-                src = self._memory.get("source")
+                src = self.output.get("source")
```
ahuang11: Is this supposed to be
```suggestion
                src = self.outputs.get("source")
```

ahuang11:
Excited for these changes!

Since messages and context are always passed together, would it make sense to include messages as part of the context object? That way, all the contextual information would be bundled into a single parameter, but the downside is more refactoring and maybe it's not worth the effort?

Trying to figure out if it's more beneficial or more of a headache to have it packaged together. The benefit I think is cleaner signatures and if you branch off a message, you can just pass context and it contains all the context (messages, inputs, outputs)...?

ahuang11:
```
/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py:2155: RuntimeWarning: coroutine 'SourceCatalog.sync' was never awaited
  m()
/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py:2155: RuntimeWarning: coroutine 'TableExplorer.sync' was never awaited
  m()
```

```

  File "/Users/ahuang/repos/lumen/lumen/ai/coordinator.py", line 1151, in _resolve_plan
    if "AnalystAgent" in agents and all(r in provided for r in agents["AnalystAgent"].requires):
```

Launched from 
`lumen-ai serve gpt*.csv --log-level debug`

It also couldn't find any datasets.

lumen/ai/prompts/ChatAgent/main.jinja2
```diff
@@ -30,10 +30,12 @@ Here's a summary of the dataset the user just asked about:
...
 No dataset has been loaded yet. Please ask the user to load a dataset or provide more context.
 {%- endif %}
-{%- if 'vector_metaset' not in memory and 'agent_tool_contexts' not in memory %}
+{%- if 'vector_metaset' not in memory and 'sql_metaset' not in memory and 'agent_tool_contexts' not in memory %}
 The user has not provided any data or documents to chat about. Please prompt the user
 to upload some data or documents.
 {%- elif 'vector_metaset' in memory %}
 {{ memory['vector_metaset'].table_context }}
+{%- elif 'sql_metaset' in memory %}
+{{ memory['sql_metaset'].vector_metaset.table_context }}
```
ahuang11: Wondering why there would be `sql_metaset` but not `vector_metaset`

lumen/ai/prompts/LumenBaseAgent/retry_output.jinja2
```diff
@@ -12,6 +12,7 @@ Critical Requirements:
 - Maintain the exact same language and syntax style as the original
 - If the feedback mentions an error, ensure the fix directly addresses that error
 - Ground your response in the specific dataset at hand; do not try to guess column names
+- You must declare if your edit is a "insert", "replace" or "delete" op
```
ahuang11:
```suggestion
- Declare if your edit is a "insert", "replace" or "delete" op
```

lumen/ai/prompts/SQLAgent/main.jinja2
```diff
@@ -21,39 +21,54 @@ Write a SQL query for the user's data transformation request, focusing on intent
...
+  - {{ tables | join("\n  - ") }}
+{%- endfor %}
+{%- endif %}
 
+### Data summary
 {{ memory["sql_metaset"].compact_context }}
 
 {%- if 'sql' in memory %}
-🗃️ Previous: `{{ memory['sql'][:100] }}...`
+Previous: `{{ memory['sql'][:100] }}...`
```
ahuang11: Maybe we should have if len > 100, append `...` and not always suffix

lumen/ai/prompts/SQLAgent/retry_output.jinja2
```diff
@@ -3,7 +3,7 @@
 {% block instructions %}
 {{ super() }}
 
-{% if discovery_context %}
+{% if discovery_context is defined %}
```
ahuang11: Not sure if needed
```suggestion
{% if discovery_context is defined and discovery_context %}
```

lumen/ai/agents.py
```diff
@@ -174,12 +174,21 @@ async def respond(
...
         """
-        system_prompt = await self._render_prompt("main", messages)
-        return await self._stream(messages, system_prompt)
+        system_prompt = await self._render_prompt("main", messages, context)
+        return [await self._stream(messages, system_prompt)], context
+
+
+class SourceOutputs(ContextModel):
+
+    source: Source
```
ahuang11, line 189: I might not fully understand; what happens if multiple sources are added?

lumen/ai/agents.py
```diff
@@ -255,21 +262,31 @@ class ChatAgent(Agent):
...
+            )).vector_metaset
+        system_prompt = await self._render_prompt("main", messages, context, **prompt_context)
+        return [await self._stream(messages, system_prompt)], {}
+
+
+class AnalystInputs(ContextModel):
+
+    data: NotRequired[Any]
+
+    source: Source
```
ahuang11, line 284: data vs source might be slightly confusing here; maybe a comment or docstring would help.

(I think data is the dataframe or summary of the data table)

lumen/ai/agents.py
```diff
@@ -491,159 +517,80 @@ class LumenBaseAgent(Agent):
...
-            message_kwargs = dict(value=out, user=self.user)
-            self.interface.stream(replace=True, max_width=self._max_width, **message_kwargs)
-        return out
+class SQLInputs(ContextModel):
+
+    data: NotRequired[Any]
+
+    source: Source
+
+    sources: Annotated[list[Source], ("accumulate", "source")]
```
ahuang11, line 584: TIL about annotated metadata.

It's not intuitive to me what accumulate vs source is. Would be nice to have a comment.

lumen/ai/agents.py
```diff
@@ -491,159 +517,80 @@ class LumenBaseAgent(Agent):
...
+    sql: NotRequired[str]
+
+    sql_metaset: SQLMetaset
+
+    visible_slugs: NotRequired[set[str]]
+
+
+class SQLOutputs(ContextModel):
+
+    data: Any
```
ahuang11, line 595: Maybe we can specify a custom type for this. If I'm not mistaken it's either a dataframe or a dict.

lumen/ai/agents.py
```diff
@@ -692,84 +639,37 @@ class SQLAgent(LumenBaseAgent):
...
-                sql_queries[query_obj.expr_slug.strip()] = query_obj.query.strip()
-
-        return sql_queries
+    input_schema = SQLInputs
+    output_schema = SQLOutputs
 
     async def _validate_sql(
-        self, sql_query: str, expr_slug: str, dialect: str,
-        source, messages: list[Message], step, max_retries: int = 2,
+        self, context: TContext,
```
```suggestion
        self,
        context: TContext,
```

lumen/ai/agents.py
ahuang11: Also, is it intended that sources here is a list of source? I see elsewhere it's dict[Source]

lumen/ai/agents.py
```diff
@@ -783,24 +683,28 @@ async def _validate_sql(
...
+                    discovery_context=discovery_context
                 )
-                sql_query = clean_sql(retry_result, dialect)
+                sql_query = clean_sql(retry_result, source.dialect)
         return sql_query
 
     async def _execute_query(
-        self, source, expr_slug: str, sql_query: str,
-        is_final: bool, should_materialize: bool, step
+        self, source, context: TContext, expr_slug: str, sql_query: str, tables: list[str],
```
```suggestion
        self, source: BaseSQLSource, context: TContext, expr_slug: str, sql_query: str, tables: list[str],
```

lumen/ai/agents.py
```diff
@@ -826,55 +730,71 @@ async def _finalize_execution(
...
+        self, sources: dict[tuple[str, str], BaseSQLSource], tables: list[tuple[str, str]]
+    ) -> tuple[BaseSQLSource, list[str]]:
+        if len(set(m.source for m in tables)) == 1:
+            m = tables[0]
+            return sources[(m.source, m.table)], [m.table]
+        mirrors = {}
+        for m in tables:
+            if not any(m.table.rstrip(")").rstrip("'").rstrip('"').endswith(ext)
+                       for ext in [".csv", ".parquet", ".parq", ".json", ".xlsx"]):
+                renamed_table = m.table.replace(".", "_")
```
ahuang11, line 777: If I vaguely remember correctly, we have these extensions littered across the codebase (e.g. coordinator.py or ui.py); we should have a constant for the recognized ones.

I feel like we should have a better way to handle this, i.e. util or have source include a method to do this.

lumen/ai/agents.py
```diff
@@ -909,33 +832,35 @@ async def _render_execute_query(
...
 
-            # Extract SQL query
+            if len(sources) == 1:
+                source = next(iter(sources.values()))
+                tables = [next(table for _, table in sources)]
+            else:
+                source, tables = self._merge_sources(sources, output.tables)
             sql_query = output.query.strip()
-            expr_slug = output.expr_slug.strip()
+            expr_slug = output.table_slug.strip()
```
ahuang11, line 857: Just wanted to verify this is correct.

Also, expr_slug I suppose is confusing.

lumen/ai/agents.py
```diff
@@ -909,33 +832,35 @@ async def _render_execute_query(
...
             pipeline, sql_expr_source, summary = await self._execute_query(
-                source, expr_slug, validated_sql, is_final=True,
-                should_materialize=True, step=step
+                source, context, expr_slug, validated_sql, tables=tables,
+                is_final=True, should_materialize=True, step=step
             )
 
-            # Finalize execution
             results = {
                 expr_slug: {
```
ahuang11: Do we need to rename to table_slug here?

lumen/ai/agents.py
```diff
@@ -1082,28 +1018,60 @@ async def _explore_tables(
...
+            sources=sources,
             step_title="Generating final answer with discoveries",
             success_message="Final answer generated with discovery context",
             discovery_context=discovery_context,
             raise_if_empty=False,
             output_title=step_title
         )
 
+
+    async def revise(
```
ahuang11, line 1045: Do we need `@retry_llm_output()` if the parent has it?

lumen/ai/agents.py
```diff
@@ -1528,7 +1523,7 @@ class VegaLiteAgent(BaseViewAgent):
 
     _output_type = VegaLiteOutput
 
-    _retry_target_keys = ["spec"]
+    _retry_target_keys = []
```
ahuang11: Do we need this anymore? I think the functionality was removed?

lumen/ai/agents.py
```diff
@@ -1843,50 +1841,58 @@ async def _get_doc_examples(self, user_query: str) -> list:
...
         messages: list[Message],
-        memory: _Memory,
-        original_output: str,
+        context: TContext,
+        view: LumenOutput | None = None,
+        spec: str | None = None,
         language: str | None = None,
-        **context
+        errors: list[str] | None = None,
+        **kwargs
```
ahuang11, line 1882: That seems to be a lot of args for a public method. Hopefully we can optimize the signature.

lumen/ai/config.py
```diff
@@ -20,6 +20,27 @@
...
+
+FORMAT_LABELS = {
+    "csv": "CSV (.csv)",
+    "xlsx": "Excel (.xlsx)",
+    "sql": "SQL (.sql)",
+    "jpeg": "JPEG (.jpeg)",
+    "png": "PNG (.png)",
+    "svg": "SVG (.svg)",
+    "pdf": "PDF (.pdf)",
+}
```
ahuang11, line 42: Not sure whwat this is for yet, but do we need parquet / json

lumen/ai/coordinator.py
```diff
@@ -266,172 +305,73 @@ def __init__(
...
             if isinstance(agent, AnalysisAgent):
                 analyses = "\n".join(
-                    f"- `{analysis.__name__}`: {(analysis.__doc__ or '').strip()}"
+                    f"- `{analysis.__name__}`: {dedent(analysis.__doc__ or '').strip()}"
                     for analysis in agent.analyses if analysis._callable_by_llm
                 )
                 agent.purpose = f"Available analyses include:\n\n{analyses}\nSelect this agent to perform one of these analyses."
-                agent.interface = interface
                 self._analyses.extend(agent.analyses)
+            # must use the same interface or else nothing shows
```
ahuang11, line 348: Do we still need this comment?

lumen/ai/models.py
```diff
@@ -51,18 +53,23 @@ class ThinkingYesNo(BaseModel):
...
+        table: Literal[tuple(set(table for _, table in sources))]
+    return LiteralSourceTable
+
+
 class SampleQuery(PartialBaseModel):
     """See actual data content - reveals format issues and patterns."""
 
-    query: SkipJsonSchema[str] = Field(default="SELECT * FROM {table} LIMIT 5")
-    table: str = Field(description="Table to sample")
+    query: SkipJsonSchema[str] = Field(default="SELECT * FROM {slug[table]} LIMIT 5")
```
ahuang11, line 66: I assume `table` is defined in the format kwargs?

lumen/ai/schemas.py
ahuang11, line 248: `prev` is missing docstring.

lumen/ai/agents.py
philippjfr: There's nowhere in the context that sources is a dict.

lumen/ai/prompts/SQLAgent/main.jinja2
philippjfr:
```suggestion
Previous: `{{ memory['sql']`
```

ahuang11:
Ah wasn't aware about 3.10 not supported: https://github.com/holoviz/lumen/pull/1482

Also there seems to be a bug when clicking the quick start buttons where it's stuck on the landing page.

https://github.com/user-attachments/assets/4d676069-d596-48ca-8f31-312ff84d1b43

Typing a question and entering manually works.

lumen/ai/agents.py
philippjfr: Good catch, fixed.

lumen/ai/agents.py
philippjfr: No, fixed.

lumen/ai/agents.py
philippjfr: Correct, thanks!

lumen/ai/agents.py
philippjfr: feedback, messages and context are the only required ones, so doesn't seem so bad. Also, it's mainly public so the Plan can call it, doesn't necessarily mean users are likely to use it.

lumen/ai/config.py
philippjfr: If we decide to expose those then yes, but for now I don't offer that for downloads.

lumen/ai/schemas.py
philippjfr: Think you fixed this.

Issue #1461, opened 2025-10-22
Incorporate multimodality


ahuang11:
Whether it's user's input or for vega lite agent feedback cycle.

```python
import matplotlib.pyplot as plt
import numpy as np

import instructor
from instructor.processing.multimodal import Image
import openai
import base64
from io import BytesIO
from pydantic import BaseModel

fig = plt.figure()
ax = fig.add_subplot(111)
ax.plot(np.random.rand(10))
plt.savefig("plot.png")


class ImageDescription(BaseModel):
    suggested_improvements: list[str]


with open("plot.png", "rb") as f:
    image_data = base64.b64encode(f.read()).decode("utf-8")
    # include header b64
    image_data = "data:image/png;base64," + image_data

client = instructor.from_provider("openai/gpt-4.1-mini")
response = client.chat.completions.create(
    response_model=ImageDescription,
    messages=[
        {
            "role": "user",
            "content": [
                "How do I improve plot so that it tells a story?",
                Image.from_base64(image_data),
            ],
        }
    ],
)

print(response)
```

<img width="2542" height="454" alt="Image" src="https://github.com/user-attachments/assets/932e65c8-62a2-43b5-a308-457ef154c783" />

PR #1462, opened 2025-10-23
Add `convert` to VegaLiteView

ahuang11:
A step towards https://github.com/holoviz/lumen/issues/1461 so the LLMs can see what they plotted and adjust the plot based on the image.

image

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1273,7 +1273,7 @@ class VegaLiteView(View):
     def get_panel(self) -> pn.pane.Vega:
         return pn.pane.Vega(**self._normalize_params(self._get_params()))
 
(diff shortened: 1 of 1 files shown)


lumen/tests/views/test_base.py
```diff
@@ -324,3 +324,425 @@ def test_view_list_param_function_roundtrip(set_root):
...
+        'table': 'test',
+        'spec': vegalite_spec,
+        'width': 600,
+        'height': 400
+    }
+    
+    view = View.from_spec(view_spec, source)
+    
+    # Test PNG conversion
+    png_data = view.convert('png')
```
philippjfr, line 391: Don't quite see the point in having these tests here, you're really just testing Panel functionality with this.

philippjfr: My bad, thought you were using Panel's ability to convert Vega to PNG/SVG. Ignore me.

PR #1466, opened 2025-10-25
Migrate to Material Docs

ahuang11:
Edit: for the latest changes, see https://holoviz-dev.github.io/lumen/
image

-----

Pretty happy with how it looks now:

https://github.com/user-attachments/assets/3fc0cc43-f947-4f53-9864-5918be704ed5

I've only migrated the Lumen AI docs, not Lumen base. In fact, I wonder if we should simplify `pip install lumen[ai]` to just `pip install lumen` because the main product of Lumen now in my opinion is the AI aspect?

So I'm semi-torn on whether to migrate the base Lumen specs, or will that be confusing? Maybe bury it somewhere and add a disclaimer...

- [ ] add recent video demo
- [x] update pixi.toml
- [x] update dev-dependencies somewhere?
- [x] migrate lumen base docs
- [x] break down agents / tools (seems a tad long)
- [x] add API reference
- [x] delete old doc/
- [ ] maybe a comparison to other tools page like wren ai
- [x] migrate memory references -> Philipp's upcoming context refactor


ahuang11:
@MarcSkovMadsen would like to hear your feedback as well!

PR #1470, opened 2025-10-28
Multimodal Support

ahuang11:
Partially addresses https://github.com/holoviz/lumen/issues/1461

In a separate PR, need to integrate with ChatAreainput

```python
from lumen.ai.agents import AnalystAgent
from lumen.ai.llm import OpenAI
from instructor.processing.multimodal import Image

import panel as pn

pn.extension()

openai = OpenAI()

url = "https://templates.mediamodifier.com/645124ff36ed2f5227cbf871/supermarket-receipt-template.jpg"

await AnalystAgent(llm=openai).respond(
    [
        {
            "role": "user",
            "content": [
                "Analyze this",
                Image.from_url(url),
            ],
        }
    ]
)
```

image

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import asyncio
+import base64
 import os
 
 from functools import partial
+from pathlib import Path
 from types import SimpleNamespace
 from typing import Any, Literal, TypedDict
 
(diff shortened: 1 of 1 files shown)


lumen/ai/llm.py
```diff
@@ -25,6 +26,13 @@ class Message(TypedDict):
     content: str
     name: str | None
 
+
+class ImageResponse(BaseModel):
```
philippjfr, line 32: Not sure I follow this part.

lumen/ai/llm.py
```diff
@@ -122,6 +130,17 @@ def _add_system_message(
             messages = [{"role": "system", "content": system}] + messages
         return messages, input_kwargs
 
+    def _check_for_image(self, messages: list[Message]) -> bool:
```
philippjfr: Would be nice to support `panel.pane.image.ImageBase` types here.

lumen/ai/llm.py
ahuang11: We don't always `patch`, since instructor doesn't natively support `str` type, or at least costs more tokens than usual
https://github.com/holoviz/lumen/blob/abe0e3bfee8078fee055f22d7b7bfb035bc09832/lumen/ai/llm.py#L507-L508

Without `patch`, the `from instructor.processing.multimodal import Image` doesn't work with OpenAI native calls, so this is a way to force the code to patch. Otherwise, I would need to implement separate paths for each provider LLM.

lumen/ai/llm.py
ahuang11: okay this works now
```python
from lumen.ai.agents import AnalystAgent
from lumen.ai.llm import OpenAI
import panel as pn

pn.extension()

openai = OpenAI()

url = "https://templates.mediamodifier.com/645124ff36ed2f5227cbf871/supermarket-receipt-template.jpg"
jpg = pn.pane.JPG(url)

await AnalystAgent(llm=openai).respond(
    [
        {
            "role": "user",
            "content": [
                "Analyze this",
                jpg
            ],
        }
    ]
)
```

PR #1488, opened 2025-12-02
Add table lookup as preplan

ahuang11:
Builds on https://github.com/holoviz/lumen/pull/1458 and does https://github.com/holoviz/lumen/pull/1487 more robustly.

The issue is that if there are a lot of tables, and the table list is truncated, and the planner doesn't see the table, it simply assumed there is no data related to <topic>, and calls on ChatAgent to pass this message. However, that is false because there is data related to <topic>--it's just not shown.

This PR addresses this by putting TableLookup as the default of planner_tools, so that it looks up vector store and provides better recon than just listing the first 5 slugs in visible slug.

```python
## Current Data Context
Data sources (sorted by relevance):
- data_test_olympics_csv (similarity: 0.38)
- data_olympic_athletes_csv (similarity: 0.34)
- data_olympic_results_csv (similarity: 0.21)
- data_olympic_hosts_csv (similarity: 0.21)
- data_gpt_4o_mini_csv (similarity: 0.20)
- (showing top 5 of 10 tables by relevance)
```

Separately, fixes a bug with interface not being passed in ui.py

### DIFF
--- a/lumen/tests/ai/test_coordinator.py
+++ b/lumen/tests/ai/test_coordinator.py
@@ -14,7 +14,8 @@ from panel_material_ui import Card, ChatMessage, Typography
 from lumen.ai.agents import AnalystAgent, ChatAgent, SQLAgent
 from lumen.ai.coordinator import Plan, Planner
 from lumen.ai.models import (
-    Reasoning, ReplaceLine, RetrySpec, SqlQuery, make_plan_model,
+    Reasoning, ReplaceLine, RetrySpec, SqlQuery, ThinkingYesNo,
+    make_plan_model,
 )
 from lumen.ai.report import ActorTask
 from lumen.ai.schemas import get_metaset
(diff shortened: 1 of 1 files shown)


lumen/ai/coordinator.py
```diff
@@ -661,10 +663,23 @@ class Planner(Coordinator):
...
+        # Now initialize planner_tools with access to self.vector_store
+        if planner_tools_input:
+            # Pass vector_store explicitly to ensure sharing
+            init_params = {**params, 'vector_store': self.vector_store}
+            self.planner_tools = self._initialize_tools_for_prompt(planner_tools_input, **init_params)
+
+            # Prepare planner_tools
+            context = params.get('context', self.context)
+            for tool in self.planner_tools:
+                state.execute(partial(tool.prepare, context))
```
philippjfr: Why would self.context differ from params['context']?

```suggestion
            for tool in self.planner_tools:
                state.execute(partial(tool.prepare, self.context))
```

PR #1495, opened 2025-12-04
metaset schemas

ahuang11:
After consolidating sql_metaset and vector_metaset, I realized metaset satisfies SQLAgent's requirements, and thus it loses the enums & min/max.

Here, I inject the schemas for the top 5 tables in _gather_prompt_context.

Previous flow was:
On init, store embeddings -> Planner -> IterativeTableLookup (gets schemas) -> SQLAgent (consumes schemas)
Now, the flow is:
On init, store embeddings -> Planner -> TableLookup (pre-plan) -> SQLAgent (JIT gather relevant table schemas, or plan next step if one shot fails)

I think maybe we can remove IterativeTableLookup...


lumen/ai/agents.py
```diff
@@ -656,6 +656,18 @@ class SQLAgent(LumenBaseAgent):
...
     output_schema = SQLOutputs
 
+    async def _gather_prompt_context(self, prompt_name: str, messages: list, context: TContext, **kwargs):
+        """Pre-compute metaset context for template rendering."""
+        prompt_context = await super()._gather_prompt_context(prompt_name, messages, context, **kwargs)
+
+        # Ensure schemas are loaded for top tables before rendering
+        if "metaset" in context:
+            metaset = context["metaset"]
+            top_tables = metaset.get_top_tables(n=5)
```
philippjfr, line 666: Hmm, why 5? Seems potentially extremely expensive.

PR #1496, opened 2025-12-04
Remove iterative table lookup

ahuang11:
Builds on https://github.com/holoviz/lumen/pull/1495

Previous flow was:
On init, store embeddings -> Planner -> IterativeTableLookup (gets schemas) -> SQLAgent (consumes schemas)
Now, the flow is:
On init, store embeddings -> Planner -> TableLookup (pre-plan) -> SQLAgent (JIT gather relevant table schemas, or plan next step if one shot fails)

Since SQLAgent has plan next step, I'm not sure if IterativeTableLookup is needed.

### DIFF
--- a/lumen/ai/models.py
+++ b/lumen/ai/models.py
@@ -331,50 +331,6 @@ def make_agent_model(agent_names: list[str], primary: bool = False):
     )
 
 
-def make_iterative_selection_model(table_slugs):
-    """
-    Creates a model for table selection in the coordinator.
-    This is separate from the SQLAgent's table selection model to focus on context gathering
-    rather than query execution.
-    """
-    table_model = create_model(
-        "IterativeTableSelection",
-        chain_of_thought=(str, FieldInfo(
-            description="""
-            Consider which tables would provide the most relevant context for understanding the user query,
-            and be specific about what columns are most relevant to the query. Please be concise.
-            """
-        )),
-        is_done=(bool, FieldInfo(
-            description="""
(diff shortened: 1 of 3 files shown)


philippjfr:
I'm all for the simplification but I am a little concerned that we're optimizing for local usecases with both of these PRs. We did add complexity for a reason originally. Once you connect to databases querying metadata for a table can become expensive so iterative discovery makes sense. Can you at outline the tradeoffs?

ahuang11:
> Once you connect to databases querying metadata for a table can become expensive so iterative discovery makes sense. Can you outline the tradeoffs?

I think IterativeToolLookup queries metadata too using `get_schema` right?

At the moment, since TableLookup from preplan already provides metaset, the Planner **never** calls IterativeTableLookup, especially since there's no iterative planning at the moment. We would have to introduce back sql_metaset or something to make it trigger.

Currently SQLAgent falls back to `make_discovery_model` to discover tables. My dream would be to make it work like Claude Code, while Claude Code uses grep / find to find relevant files in the codebase, SQLAgent would use various `SampleQuery` and `DistinctQuery` and potentially a `TableQuery` to find the relevant tables.

Admittedly, the other PR could better handle `n=5` (or use a LLM to decide how much to see or not see).

Currently, there are discovery layers:
1. Vector store embeddings (TableLookup, cheapest)
2. Schemas (IterativeTableLookup / SQLAgent for high level info for a few relevant tables, ??$)
3. Discovery queries (SQLAgent, ??$)

I can also test with a test database with many tables to see how it works out...

PR #1499, opened 2025-12-05
Fix uploading

ahuang11:
Unifies tool instances across `tools` and `planner_tools`, so that it shares the same vector store and state. 

Also moves the upload icon to the top right because I think it shouldn't be hidden.

In conjunction with https://github.com/holoviz/lumen/pull/1498, uploading datasets finally work.

image


philippjfr:
Not fully sold on the upload button up there but that's a minor detail. Will merge.

Issue #1526, opened 2025-12-15
Migrate IterativeTableLookup with Integrated Discovery Pipeline for SQLAgent


ahuang11:
<h2>Problem</h2>
<p>Table discovery is currently <strong>divorced from data verification</strong>:</p>
<ul>
<li><code>TableLookup</code> finds tables by metadata only (no schema)</li>
<li><code>IterativeTableLookup</code> examines schemas but <strong>never looks at actual data</strong></li>
<li><code>SQLAgent</code> only verifies data content <strong>after SQL fails</strong></li>
</ul>
<p>This leads to:</p>
<ul>
<li>❌ SQL errors from assuming schema = data reality</li>
<li>❌ Slow iteration: search → wrong SQL → error → fix → retry</li>
<li>❌ Poor handling of large databases (100+ tables)</li>
</ul>
<h2>Solution</h2>
<p><strong>Integrate discovery into SQLAgent</strong> following the Claude Code pattern:</p>
<pre><code>grep "pattern" → find candidates
read file1     → verify it's relevant
read file2     → verify it's relevant
               → narrow to 2-3 files that help
</code></pre>
<h2>Example: Discovery Loop in Action</h2>
<p><strong>Query:</strong> "Show top 5 customers by revenue in China"</p>
<h3>Iteration 1</h3>
<pre><code>📊 Found 20 candidates from vector search

🔍 LLM selects batch:
   • customers (similarity: 0.89)
   • orders (similarity: 0.85)  
   • revenue_by_country (similarity: 0.78)

⚡ Running discoveries in parallel:
   1. Sample customers → ✓ See structure, has customer_name
   2. Distinct customers.country LIKE '%chin%' → ✓ Found "China"
   3. Sample revenue_by_country → ✓ Pre-aggregated by country!
   4. Wildcard info_schema for "revenue" → Found 2 similar tables

✅ LLM: "Sufficient! revenue_by_country already has what we need"
   Selected: [revenue_by_country, customers]
   Confidence: HIGH
</code></pre>
<h3>Result</h3>
<pre><code class="language-sql">SELECT customer_name, total_revenue
FROM revenue_by_country
WHERE country = 'China'
ORDER BY total_revenue DESC
LIMIT 5
</code></pre>
<p>✨ <strong>First try success</strong> - no errors, no retries</p>
<hr>
<h2>Architecture Changes</h2>
<h3>Before (3 separate tools)</h3>
<pre><code>TableLookup → metadata only
    ↓
[...]

PR #1527, opened 2025-12-16
Refactor metadata ingestion

ahuang11:
Edit:

Like https://github.com/holoviz/lumen/issues/1526

> Table discovery is currently divorced from data verification

I don't think tables and documents metadata should be separated; thus I renamed to MetadataLookup and put document_sources into Metaset which to me feels like a better abstraction. Looks like this now...

- [x] Case 1: metadata associated with table

https://github.com/user-attachments/assets/44160557-97cc-4748-ab77-2114c7f7d74c

- [x] Case 2: docs disabled

https://github.com/user-attachments/assets/e958cf1e-71f7-47de-b707-1e71c4717af5

- [x] Case 3: only docs...

https://github.com/user-attachments/assets/4e385de5-c5ea-42f2-981a-e68590324ef2

- [x] Case 4: no docs, no tables (failing because I didn't handle disable tables)

image

- [x] Case 5: upload documents from chat area input and query at the same time

https://github.com/user-attachments/assets/8047781d-39ec-4939-90b2-f12e335a3c5b

Upload Docs → Metaset Flow

1. **Upload** (`BaseSourceControls._add_metadata_file`)
```python
# Store in catalog + sync to vector store
source_catalog._available_metadata.append({"filename": "readme.md", "content": "..."})
await source_catalog._sync_metadata_to_vector_store("readme.md")
```

2. **Auto-Association** (`SourceCatalog.sync` → `_auto_associate_unassociated_metadata`)
```python
# When new docs detected, associate with ALL tables
source.metadata[table]["docs"] = ["readme.md"]
context["visible_slugs"].add(table_slug)  # Mark table as visible
```

3. **Query Time** (`TableLookup._query_documents`)
```python
# Filter docs based on:
# - Skip if in disabled_docs (global disable)
# - Include if NOT table-associated (global doc)
# - Include if table-associated AND table in relevant_tables
docs = [DocumentChunk(...) for matching docs]
```

4. **Build Metaset** (`TableLookup._gather_info`)
```python
metaset = Metaset(
    catalog={table_slugs...},
    docs=docs  # Already filtered
)
```

5. **LLM Context** (Templates)
```jinja2
{{ memory["metaset"].compact_context(5) }}
# Renders docs from metaset.docs (already filtered)
```

**Key Points:**
- Docs stored in vector store (for semantic search)
- Table associations tracked in `source.metadata[table]["docs"]`
- Global disable tracked in `context["disabled_docs"]`
- Filtering happens in `TableLookup._query_documents()` based on both

Notes:

1. DocumentListAgent needs to be hooked up I think (DONE)
2. Unchecking all of the tables' doc filenames should also uncheck global doc (DONE)
3. Move global docs up before tables (and perhaps better name it) (DONE)
4. Handling document filenames clashing? (DONE)
5. Maybe tempted to either get rid of TableListAgent and DocumentListAgent or merge as SourceListAgent

Separate PR:
- MetadataLookup (vector search is cheap)--should have an auto-approve (no need for YesNo ToolRelevant for MetadataLookup)
- Remove IterativeTableLookup next
- NumpyEmbeddings vector search doesn't do so well; have to word it carefully or else it won't pick up on the relevant docs. (maybe)


lumen/ai/agents/base.py
```diff
@@ -41,7 +41,7 @@ class Agent(Viewer, ToolUser, ContextProvider):
         The name of the user that will be respond to the user query.""")
 
     # Panel extensions this agent requires to be loaded
-    _extensions = ()
+    _extensions = ('filedropper',)
```
ahuang11: Not sure if this is the proper place to put it, probably not.

philippjfr: Just add it in `UI._create_view` where this is used.

lumen/ai/controls.py
```diff
@@ -51,143 +54,555 @@ class DownloadConfig:
...
-        self._file_input.param.watch(self._generate_media_controls, "value")
-        self._upload_tabs = Tabs(sizing_mode="stretch_width", closable=True)
-        files_to_process = self._upload_tabs.param["objects"].rx.len() > 0
-        self._upload_tabs.visible = files_to_process
+        self._file_input.param.watch(self._on_file_upload, "value")
+
+        self._layout = MuiColumn(
+            self._file_input,
+            self._upload_cards,
+            Row(self._add_button),
```
philippjfr: Niggle, but why the row?

lumen/ai/ui.py
```diff
@@ -200,6 +194,13 @@ def __init__(
         self._configure_session()
         self._render_page()
 
+        # Wire up vector stores to components after they are created
+        # Use document_vector_store if available, otherwise fall back to main vector_store
+        doc_store = self.document_vector_store or self._coordinator.vector_store
+        self._source_catalog.vector_store = doc_store
+        self._upload_controls.source_catalog.vector_store = doc_store
+        self._download_controls.source_catalog.vector_store = doc_store
```
philippjfr: Please move these to the appropriate places in the methods below.

lumen/ai/ui.py
```diff
@@ -444,27 +445,99 @@ def _handle_file_drop(self, event):
...
+
+        # Send message and respond if there's a query
+        self._transition_to_chat()
+        if user_prompt:
+            msg = Column(user_prompt, source_view) if source_view else user_prompt
+            with hold():
+                self.interface.send(msg, respond=True)
+        elif source_view:
+            with hold():
+                self.interface.send(source_view, respond=False)
```
philippjfr: Why hold for a single action?

lumen/ai/ui.py
```diff
@@ -642,6 +633,13 @@ async def _do_sync():
         self._upload_controls = UploadControls(context=self.context, source_catalog=self._source_catalog)
         self._download_controls = DownloadControls(context=self.context, source_catalog=self._source_catalog)
 
+        # Wire up vector stores to components after they are created
+        # Use document_vector_store if available, otherwise fall back to main vector_store
+        doc_store = self.document_vector_store or self._coordinator.vector_store
+        self._source_catalog.vector_store = doc_store
+        self._upload_controls.source_catalog.vector_store = doc_store
+        self._download_controls.source_catalog.vector_store = doc_store
```
philippjfr: The vector store can't be passed straight to the constructor?

PR #1532, opened 2025-12-29
Split buttons

ahuang11:
I think clickable things should be in blue?

image


lumen/ai/config.py
```diff
@@ -59,6 +59,8 @@ class MissingContextError(Exception):
 
 THIS_DIR = Path(__file__).parent
 PROMPTS_DIR = THIS_DIR / "prompts"
+STYLESHEETS_DIR = THIS_DIR / "stylesheets"
+SPLITJS_STYLES_PATH = STYLESHEETS_DIR / "splitjs.css"
```
philippjfr: Just read it in from the file here. Panel will simply load it from the file each time otherwise.

PR #1535, opened 2025-12-30
Fix vs test

ahuang11:
Setting max_concurrent=1 seems to have fixed


philippjfr:
Doesn't this indicate that concurrency is broken?

Issue #1545, opened 2026-01-02
Simplify YAML Download


MarcSkovMadsen:
It seems I can download YAML in too places. This is overload and confusing - I would be in doubt whether its the same thing that happens.

I would recommend removing the lower/ second download button.

<img width="986" height="674" alt="Image" src="https://github.com/user-attachments/assets/c557a1f3-f6f5-4ebb-9a91-13a29ae57d80" />

## 2. Simplify Export

Simplify text by changing "Export as ..." to "Export".

<img width="540" height="221" alt="Image" src="https://github.com/user-attachments/assets/3a47b388-edbc-4d68-a659-8aacfa17a6c3" />

## 3. Give downloaded file meaningful name

"Create Spec" does not mean anything and could be coming from anywhere. Give it a name you can relate to and find later. Like lumen-vega-spec.yaml or lumen-vega-spec-20250201103700.yaml or penguins-vega-spec.yaml.

<img width="228" height="58" alt="Image" src="https://github.com/user-attachments/assets/99748b45-43f2-4cca-89dc-3810b2228d98" />

PR #1553, opened 2026-01-02
simplify-yaml-download

MarcSkovMadsen:
Closing #1545 

## Export as simplified

image

## Dropdown names simplified and icons improved

image

## Downloaded file name improved

image

### DIFF
--- a/lumen/ai/config.py
+++ b/lumen/ai/config.py
@@ -22,9 +22,9 @@ ChatMessage.default_avatars.update({
 
 FORMAT_ICONS = {
     "csv":   "table_chart",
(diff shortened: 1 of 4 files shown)


ahuang11:
I've made changes:

1. The width/height of output was not set, resulting in a poorly formatted plot. 

Before:
image

After (disregard the inline labels; I'll tweak the vega lite prompt after)
image


2. I've simplified the exploration UI, namely, removed the footer and merged it into the header. I noticed that there was multiple download icons and it was confusing.

Before:
image

After:
image

3. Tweaked based on Copilot's suggestions.

Issue #1557, opened 2026-01-02
Settle on one term (Analysis, Exploration or Session)


MarcSkovMadsen:
Through out the documentation and Explorer UI the words Analysis, Exploration and Session are used interchangeably. Please settle on one word and use this consistently. Will make it much easier for users.

I would recommend Session over Analysis over Exploration. Session is already known from ChatGPT. Exploration would be the least known and hardest to teach users.

Its really hard to understand what "Home" is. But "New Session" everyone would understand.

philippjfr:
I don't think either Chats or Threads conveys what we want it convey, personally. While I think it at least strongly and intuitively conveys that they are contextual "sessions" it risks framing Lumen as “just another chat tool” and does not adequately convey that these are chats with associated artifacts and data state. The important distinction is that an Exploration isn't created until a data artifact is associated with it.

Exploration has the opposite problem I think, it kind of conveys the idea that you're exploring something (the data) but doesn't convey the idea of a contextual session and so requires explanation.

The only thing I can think of that benefits from an existing intuitive understanding among users and conveys the fact that it encompasses artifacts would be "Workspace".

Issue #1556, opened 2026-01-02
Settings shown not the same as documented


MarcSkovMadsen:
Related to #1541.

The settings are all disabled in the menu

<img width="386" height="241" alt="Image" src="https://github.com/user-attachments/assets/b2c8c2ad-3576-41f5-befe-30ab55580514" />

But according to the documentation they are not

https://holoviz-dev.github.io/lumen/getting_started/navigating_the_ui/#control-how-analysis-runs.

<img width="861" height="392" alt="Image" src="https://github.com/user-attachments/assets/aac54498-ad4d-4dc9-b479-cd3d066dcbbc" />

PR #1559, opened 2026-01-02
Improve navigation docs

MarcSkovMadsen:
An attempt to understand the ExplorerUI and document this to the benefit of other users.


MarcSkovMadsen:
Needs a lot more work. Much more is needed then the time I have right now.

docs/getting_started/navigating_the_ui.md
```diff
@@ -43,23 +57,11 @@ The chat interface provides several built-in actions:
...
-**Validation Step** — When enabled, the AI double-checks results for correctness. This catches data errors early. Enabled by default.
-
-Simply click the toggle next to each option to turn it on or off.
-
 ### Navigate the sidebar
 
 The left sidebar provides quick access to all major features:
 
-**Home** — Your starting point. This is where all analyses begin. Switch back here anytime.
+**Home** — Your starting point. This is where all sessions begin. Switch back here anytime.
```
MarcSkovMadsen, line 64: We need to settle on a term. See #1557 .

docs/getting_started/navigating_the_ui.md
```diff
@@ -71,6 +73,20 @@ The left sidebar provides quick access to all major features:
 
 **Settings** — Toggle Chain of Thought, SQL Planning, and Validation Step options.
 
+### Control how analysis runs
+
+The settings menu (⚙️ icon in the left sidebar) lets you toggle three analysis features:
```
MarcSkovMadsen, line 78: Would need feedback on #1556 . Don't know if the original documentation was wrong. Or if the UI has a bug.

ahuang11: It was enabled initially, but now disabled.

docs/getting_started/navigating_the_ui.md
ahuang11: Not in favor of sessions because when I think of sessions, I think about each browser tab visiting Lumen.

Maybe Thread? Collection? Trace?

PR #1563, opened 2026-01-06
Show llm error on init

ahuang11:
Previously, we remove the status indicator of LLM. Here, I re-implement it by disabling chat area input on error and add alert.

image

image


philippjfr:
I'm happy with the Alert but I do not think we should be surfacing the full error like that.

ahuang11:
Okay added a friendly message

https://github.com/user-attachments/assets/25b82557-11b2-48aa-abaa-bc81a2776e64

PR #1589, opened 2026-01-12
missing file error

ahuang11:
```
(lumen) ➜  lumen git:(missing_file) ✗ lumen-ai serve missing
ERROR: ERROR: Table file not found: missing
```

Also makes the source code yellow
image

### DIFF
--- a/lumen/command/ai.py
+++ b/lumen/command/ai.py
@@ -199,12 +199,22 @@ class AIHandler(CodeHandler):
     ) -> None:
         # Validate that table files exist
         for table_path in tables:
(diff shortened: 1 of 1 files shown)


lumen/command/ai.py
```diff
@@ -196,6 +197,15 @@ def __init__(
         logfire_tags: list[str] | None = None,
         **kwargs,
     ) -> None:
+        # Validate that table files exist
+        for table_path in tables:
+            if table_path != "no_data":  # Skip validation for special "no_data" case
```
philippjfr: I don't know if it's AI or not, but an if statement like this is perfectly readable. No need to comment.

PR #1590, opened 2026-01-12
Migrate export

ahuang11:
Moved export to menu

image

https://github.com/user-attachments/assets/fcb4f43d-2d17-453c-ad93-e73a46cd021a

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -1767,21 +1767,7 @@ class ExplorerUI(UI):
         if exploration.plan is None:
             return
 
-        download = getattr(exploration, '_export_download', None)
-        if download is None:
-            def make_notebook():
-                nb = export_notebook(exploration.plan.views, preamble=self.notebook_preamble)
-                return StringIO(nb)
-
-            download = exploration._export_download = FileDownload(
-                callback=make_notebook,
-                filename=f"{exploration.plan.title.replace(' ', '_')}.ipynb",
-                visible=False
-            )
-            # Attach to output so it's in the DOM
-            self._output.append(download)
-            await asyncio.sleep(0.1)  # Wait for DOM attachment
-
+        download = exploration._export_download
         download.transfer()
 
     def _destroy(self, session_context):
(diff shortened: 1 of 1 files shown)


lumen/ai/ui.py
```diff
@@ -1778,6 +1761,31 @@ async def _delete_exploration(self, item):
...
+                return StringIO(nb)
+
+            download = exploration._export_download = FileDownload(
+                callback=make_notebook,
+                filename=f"{exploration.plan.title.replace(' ', '_')}.ipynb",
+                visible=False
+            )
+
+        # Attach to output so it's in the DOM
+        self._output.append(download)
```
philippjfr: Why remove and re-add. Seems prone to race conditions, particular with the added waits. Just add it once and make it invisible.

ahuang11: Was worried about a memory leak.

lumen/ai/ui.py
```diff
@@ -1778,6 +1761,29 @@ async def _delete_exploration(self, item):
...
+                nb = export_notebook(exploration.plan.views, preamble=self.notebook_preamble)
+                return StringIO(nb)
+
+            download = exploration._export_download = FileDownload(
+                callback=make_notebook,
+                filename=f"{exploration.plan.title.replace(' ', '_')}.ipynb",
+                visible=False
+            )
+            # Attach to output so it's in the DOM
+            self._output.append(download)
```
philippjfr: I'm still confused, why add it each time. Just add it once.

lumen/ai/ui.py
philippjfr: The point was to add it **once**, of course you'll get a leak if you add it each time.

lumen/ai/ui.py
```diff
@@ -1120,6 +1108,13 @@ async def _do_sync():
         self._explorations.param.watch(self._update_conversation, 'active')
         self._explorations.param.watch(self._sync_active, 'value')
         self._explorations.on_action('remove', self._delete_exploration)
+        self._explorations.on_action('export_notebook', self._export_exploration)
+
+        # Create hidden download button for exporting notebook
+        # Later appended to return so it's in the DOM
```
```suggestion
```

philippjfr (approved):

PR #1610, opened 2026-01-16
Fix duckdb detect

ahuang11:
Previously was not properly detecting duckdb db files, and then if tables was unset, it crashed

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -149,10 +149,6 @@ class DuckDBSource(BaseSQLSource):
                 processed_tables[table_alias] = sql_expr
 
             self.tables = processed_tables
-        elif self.tables is None:
-            self.tables = {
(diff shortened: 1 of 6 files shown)


lumen/sources/duckdb.py
```diff
@@ -424,7 +424,7 @@ def create_sql_expr_source(
                 self.sql_expr.format(table=f'"{table}"'),
                 self.sql_expr.format(table=table),
             )
-            if table in self.tables:
+            if table in (self.tables or {}):
```
ahuang11, line 427:
```python
  File "/Users/ahuang/repos/lumen/lumen/sources/duckdb.py", line 427, in create_sql_expr_source
    if table in self.tables or {}:
       ^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

philippjfr (approved):

PR #1453, opened 2025-10-17
First pass at getting layouts

ahuang11:
First pass at implementing https://github.com/holoviz/lumen/issues/1447

https://github.com/user-attachments/assets/e8ce37ff-fa92-4cd7-8c47-d546eda00fb0

The Three-Step Approach:
Break layout generation into three sequential steps similar to Planner -> Agents:
Planning (1 LLM call): Create a layout plan with natural language instructions for each subplot
Parallel Generation (N LLM calls): Generate each subplot independently in parallel
Assembly (Python): Programmatically combine specs into hconcat/vconcat structure

Note, these are the valid properties outside:
```yaml
spec:
  $schema: https://vega.github.io/schema/vega-lite/v5.json
  
  # SHARED by all charts
  data:
    name: token_use_and_response_time_from_gpt_4o_mini_csv
  
  # Optional: overall title for entire visualization
  title:
    text: GPT-4o Mini Token Usage Analysis
  
  # Optional: shared transforms applied before splitting
  transform:
    - filter: "datum.response_time > 0"
  
  # The concatenation container
  hconcat: [...]
```

Within each subplot:
```yaml
hconcat:
- # FIRST CHART
  width: 300                    # This chart's width
  height: container             # This chart's height
  title: Input Tokens Distribution  # This chart's title
  mark: bar                     # This chart's mark type
  encoding:                     # This chart's encodings
    x: {...}
    y: {...}
  transform:                    # Optional: transforms just for this chart
    - filter: "datum.input_tokens < 10000"

- # SECOND CHART  
  width: 300                    # Different width possible
  height: container
  title: Output Tokens Distribution
  mark: point                   # Different mark type
  encoding:
    x: {...}
    y: {...}
```

## Visual Representation
```
┌─────────────────────────────────────────────────────┐
│ spec (top-level)                                    │
│                                                     │
│  ┌─────────────────────────────────────────────┐  │
│  │ data: (SHARED by all)                       │  │
│  └─────────────────────────────────────────────┘  │
│                                                     │
│  ┌─────────────────────────────────────────────┐  │
│  │ hconcat: [                                  │  │
│  │   ┌──────────────────────────┐              │  │
│  │   │ Chart 1                  │              │  │
│  │   │ - width                  │              │  │
│  │   │ - height                 │              │  │
│  │   │ - title                  │              │  │
│  │   │ - mark                   │              │  │
│  │   │ - encoding               │              │  │
│  │   └──────────────────────────┘              │  │
[...]

PR #1616, opened 2026-01-21
Add Altair code generation mode to VegaLiteAgent

ahuang11:
Builds on https://github.com/holoviz/lumen/pull/1615

This PR introduces a generation_mode parameter for VegaLiteAgent with three options:

1. vega-lite (current behavior) - generates Vega-Lite specs directly
2. altair - generates Altair Python code, validates with AST, executes safely, and outputs vega-lite spec in the code editor area
3. altair_llm_validated - same as above but adds LLM-based security review

I was inspired by Marc's impressive holoviz-mcp dashboard demos, which uses code generation, and also frustrated because VegaLiteAgent has been the bottleneck of Lumen, which should specialize in plot generation, but right now is the slowest to run and error-prone.

I wanted to explore code generation as an alternative to direct spec generation, and it does well, 1/2 of the time to generate (altair top 7.81s, vega-lite spec bottom 16.3s), due to how succinct altair is.

image

The main drivers:

1. Token efficiency - Vega-Lite spec generation is slow and token-heavy (in fact, I suspect this is what drove Marc to create the holoviz-mcp display tool instead of using Lumen)

2. We should flow with local models. Previously smaller models struggle with Vega-Lite's verbose JSON/YAML syntax but handle Python reasonably well. Below is what Qwen3-8b was able to generate:

image

4. We can easily generate layouts now without a separate mechanism attempted in https://github.com/holoviz/lumen/pull/1453 

This **does use** exec, but some mitigating factors:

1. The LLM generates this code, and has its own safeguards against malicious output
AST validation restricts imports to altair only and blocks dangerous operations (file I/O, network, eval, etc.)
2. Database sources are read-only
4. This is not the default - users must explicitly configure generation_mode="altair" and accept the associated risk
5. Users edit a vega-lite spec in the code-editor area, not the altair code
6. The optional altair_llm_validated mode adds a second LLM pass specifically for security review, though this adds latency.


philippjfr:
Concretely I might suggest the following:

- A parameter on the ExplorerUI that allows toggling code execution to one of three states `False`, `True` and `'prompt'` and allow that to flow down to the agents. This defaults to False. If it's enabled the UI renders a settings toggle to switch between prompt or True states (with a big warning). 
- Very explicitly documenting this and discussing the fact that code execution in its current form is not safe.
- Begin work exploring what it would take to run a sandboxed Python alongside Lumen.

philippjfr:
I guess the real problem with sandboxing is how we'd get the data into the sandbox.

philippjfr (approved):

PR #1624, opened 2026-01-22
Polish report

### LINKED ISSUE #1622: Using a report isn't intuitive
From a dev POV, I think it'd be nice for actions to automatically wrap with Report, or some kind of UI, if they're shown alone, e.g. ProcessDataAction().show()

```python
import pandas as pd
import panel as pn
from lumen.ai.report import Action, Report, Section
from lumen.ai.views import LumenOutput
from lumen.views.base import VegaLiteView

pn.extension("vega")

class ProcessDataAction(Action):
    async def _execute(self, context, **kwargs):
        source = DuckDBSource(tables={"data": "data/windturbines.parquet"})
        avg_source = source.create_sql_expr_source(tables={"avg_data": "SELECT t_state, AVG(p_cap) as avg_p_cap FROM data GROUP BY t_state"})
        pipeline = Pipeline(source=avg_source, table="avg_data")

        view = VegaLiteView(pipeline=pipeline, spec={
            "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
            "mark": "bar",
            "encoding": {
                "x": {"field": "t_state", "type": "nominal"},
                "y": {"field": "avg_p_cap", "type": "quantitative"}
            },
        })
        output = LumenOutput(component=view, title="Data Overview")

        df = pipeline.data
        summary = f"Loaded {len(df)} rows, {len(df.columns)} columns"
        return [df, output], {"pipeline": pipeline, "summary": summary}


report = Report(Section(ProcessDataAction(), title="Average Power Capacity by State"))
report.show()
```

I also am confused as to how to prepopulate the report. I thought report.execute() would work, but no.

From a user POV, on reaching this page, it's confusing as to what I need to do; lots of icons, and whitespace.

Image

We should hide all icons except play if there are no outputs.

Other ideas?

ahuang11:
Closes https://github.com/holoviz/lumen/issues/1622

Adds a background to report for more of a report feel rather than blank white page, hide icons if on init so users can only see play, and finally, make the play button animate

image

https://github.com/user-attachments/assets/2eb92592-8e9f-46d1-ac9f-aa1843b7f7be


zensical.toml
```diff
@@ -23,7 +23,8 @@ nav = [
     {"Tutorials" = [
       {"Overview" = "examples/tutorials/index.md"},
       {"Weather Data AI Explorer" = "examples/tutorials/weather_data_ai_explorer.md"},
-      {"Penguins Dashboard Spec" = "examples/tutorials/penguins_dashboard_spec.md"}
+      {"SaaS Company Report Dashboard" = "examples/tutorials/saas_company_report_dashboard.md"},
+      {"Penguins Dashboard Spec" = "examples/tutorials/penguins_dashboard_spec.md"},
```
philippjfr: Let's just get rid of this for now (the link also doesn't seem to work):

```suggestion
```

philippjfr (approved):

PR #1636, opened 2026-01-26
Handle bad files

ahuang11:
Show an error message if upload failed

image


lumen/ai/controls.py
```diff
@@ -390,6 +420,152 @@ def _add_table(
...
+        -------
+        pd.DataFrame
+            The parsed DataFrame.
+
+        Raises
+        ------
+        ValueError
+            If the JSON cannot be converted to a DataFrame.
+        """
+        import json
```
philippjfr: Why inline a stdlib import?

PR #1638, opened 2026-01-27
Add controls docs

ahuang11:
image
image


philippjfr (commented):
Really nice, but why remove the reports tutorial? I'll re-add since you didn't mention wanting to remove it.

PR #1653, opened 2026-01-27
Fix analyses from being unable to re-render if unserializable

ahuang11:
image
image


lumen/ai/editors.py
```diff
@@ -236,7 +236,7 @@ async def render_context(self):
         else:
             return {}
 
-    @param.depends('spec')
+    @param.depends('spec', 'component')
```
philippjfr: This worries me as it could double the re-render events.

philippjfr (approved):

PR #1686, opened 2026-02-04
Implement LLM tool calls including FunctionTool/MCPTool

philippjfr:
Adds first-class tool-call support to the LLM interface, including `FunctionTool` conversion, execution, and streaming follow-ups so models can call tools and receive results by tool_call_id.

### Summary

- Implement `MCPTool`
- Extend `invoke`/`stream` to accept `FunctionTool` and `MCPTool` instances alongside raw tool specs.
- Translate `FunctionTool` into OpenAI-compatible tool definitions and execute tool calls.
- Return tool outputs as `role="tool"` messages keyed by `tool_call_id`, then re-invoke the model so it can use results.
- Add streaming support that accumulates incremental `tool_calls`, executes tools after stream completion, and continues streaming the follow-up.

### Implementation details
- Added helper utilities in `lumen/ai/llm.py` to normalize tools, parse/format tool calls, and handle streaming tool-call deltas.
- Updated message typing to include tool-role messages and tool call IDs.
- Integrated tool-call execution into both non-streaming and streaming flows.

### Example

```python
import lumen.ai as lmai

llm = lmai.llm.OpenAI()

@lmai.tools.define_tool
def search(query):
    return query + "!!!"

await llm.invoke([{"role": "user", "content": "Call the search tool and restate the result verbatim"}], tools=[search])
```

```
ChatCompletion(id='chatcmpl-D5ZPBml2xXv10plPj9WgFjY19k9aH', choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content='The latest news is: "latest news!!!"', refusal=None, role='assistant', annotations=[], audio=None, function_call=None, tool_calls=None))], created=1770220109, model='gpt-4.1-mini-2025-04-14', object='chat.completion', service_tier='default', system_fingerprint='fp_82dbabdb2c', usage=CompletionUsage(completion_tokens=11, prompt_tokens=79, total_tokens=90, completion_tokens_details=CompletionTokensDetails(accepted_prediction_tokens=0, audio_tokens=0, reasoning_tokens=0, rejected_prediction_tokens=0), prompt_tokens_details=PromptTokensDetails(audio_tokens=0, cached_tokens=0)))
```


lumen/ai/tools/base.py
```diff
@@ -137,17 +139,21 @@ async def respond(
...
-                messages,
-                system=prompt,
-                model_spec=model_spec,
-                response_model=self._model,
-                allow_partial=False,
-                max_retries=3,
-            )
+        tool_args = kwargs.pop("tool_args", None)
+        if tool_args is None:
+            kwargs = {}
```
ahuang11: Is it intentional we overwrite kwargs?

lumen/ai/llm.py
```diff
@@ -1275,13 +1737,33 @@ async def run_client(self, model_spec: str | dict, messages: list[Message], **kw
...
+                    parameters = dict(func.get("parameters", {}))
+                    # Remove keys that are not valid in Google's schema
+                    parameters.pop("title", None)
+                    declarations.append(FunctionDeclaration(
+                        name=func["name"],
+                        description=func.get("description", ""),
+                        parameters=parameters,
+                    ))
+            if declarations:
+                google_tools = [GoogleTool(function_declarations=declarations)]
```
ahuang11: Should we have an abstract method for this?

lumen/ai/tools/base.py
philippjfr: No, good catch.

philippjfr: Actually nvm, it kind of is but I'll likely revert these changes because the tool won't be invoked this way.

lumen/ai/llm.py
```diff
@@ -345,6 +377,203 @@ def _get_content(cls, response) -> str:
...
+        return name, args or {}, call_id
+
+    async def _run_tool_calls(
+        self,
+        tool_instances: dict[str, FunctionTool | MCPTool],
+        tool_calls: list[Any],
+        tool_contexts: dict[str, Any],
+        messages: list[Message],
+    ) -> list[Message]:
+        from .tools import FunctionTool, MCPTool
```
ahuang11, line 552: Are these imports intentionally in the function due to circular imports? If so, maybe a comment?

lumen/ai/tools/base.py
philippjfr: Simplified.

lumen/ai/llm.py
philippjfr: Right.

lumen/ai/llm.py
philippjfr: Yes.

PR #1702, opened 2026-02-19
Add export selector to Report

ahuang11:
Builds upon https://github.com/holoviz/lumen/pull/1701 to allow subsetting which subsections to download.

https://github.com/user-attachments/assets/f3293a46-40f2-492b-8ef8-8bb505062d04

The output:

image

In use:

https://github.com/user-attachments/assets/79e47c6d-b785-4930-8491-9ef56dbad82d

image


lumen/ai/report.py
```diff
@@ -720,13 +796,23 @@ def _init_view(self):
             "", variant="body2", margin=(10, 10),
             sx={"color": "text.secondary"}
         )
+        self._export_selector = MultiChoice(
+            name="Include in process & export",
+            options=[],
+            value=[],
+            sizing_mode="stretch_width",
```
philippjfr, line 803: Can we see what it looks like with this?

```suggestion
            sizing_mode="stretch_width",
            margin=(5, 0),
            variant="text",
```

ahuang11:
```
ValueError: Selector parameter 'MultiSelect.variant' does not accept 'text'; valid options include: '[filled, outlined, standard]'
```

philippjfr (approved):

lumen/ai/report.py
```diff
@@ -282,10 +291,9 @@ def __init__(self, *tasks, **params):
         self._populate_view()
 
     def _init_views(self):
-        views = self.param._header.rx()
+        self.views = self.param._header.rx()
```
philippjfr, line 294: Why? Seems like this will trigger a bunch of events (needlessly).

lumen/ai/report.py
```diff
@@ -596,7 +640,11 @@ def reset(self, start: int = 0):
         """
         Resets the view, removing generated outputs.
         """
+        self._prepared = False
         self.status = "idle"
+        self._view[:] = []
+        self.views = []
```
philippjfr, line 646: Don't think this is correct, it unlinks the `rx`.

lumen/ai/report.py
```diff
@@ -561,6 +598,12 @@ def invalidate(
         if invalidated:
             self.status = "idle"
             self._current = max(min(invalidated), 0)
+            # Keep parent views in sync with child invalidation state since
+            # execute() assigns a concrete list to self.views.
+            views = list(self._header)
+            for task in self:
+                views += task.views
+            self.views = views
```
philippjfr, line 606: Again, confused here, views should be managed entirely by the rx expression.

philippjfr (commented):
Quite worried about some of the changes here. A big part of making things here not be insane was to manage the views of a `TaskGroup` via rx expressions. This eliminated a ton of logic around ensuring that things are reset and synced appropriately. Based on what I've seen here this undoes a lot of that.

I haven't followed the exact details but I'm not sure why that logic had to change at all.

PR #1710, opened 2026-03-01
Deduplicate upload source references for multi-file uploads

SanketMeghale:
## Description
This PR adds regression coverage for multi-file upload source de-duplication behavior.

The underlying source registration logic is already present on `main` via #1711. This PR is limited to a focused test that verifies multiple uploaded data files still result in a single logical source reference containing both tables.

## Changes
- Added regression test covering multi-file upload and asserting a single source reference with both expected tables.

## Files changed
- `lumen/tests/ai/test_controls/test_upload.py`

## Testing
- `python -m pytest lumen/tests/ai/test_controls/test_upload.py -q`


SanketMeghale:
@ahuang11 You’re right. The implementation described in the PR body had already landed separately in #1711, so this branch now only contains the regression test. I’ve updated the PR description to reflect that this PR is test-only and is just preserving coverage for the multi-file upload source de-duplication behavior already present on main. If this is already sufficiently covered by #1711, I’m also happy to close this PR as superseded to avoid duplicate review.

ahuang11 (approved):

PR #1719, opened 2026-03-07
Fix per-table lock reuse in cached() decorator

ghostiee-11:
## Summary

- Fixed a bug in the `cached()` decorator where the per-table lock lookup checked `if table in locks` instead of `if table in locks[self]`
- Since `locks` is a `WeakKeyDictionary` keyed by Source instances, a string `table` key can never match, so the condition was always `False`
- This caused a new `RLock` to be created on every `Source.get()` call, defeating per-table serialization for concurrent access
- The sibling function `cached_schema()` already does this correctly (line 130)

## Bug Demonstration

```python
>>> import weakref
>>> locks = weakref.WeakKeyDictionary()
>>> locks[source] = {"main": RLock(), "test_table": RLock()}
>>> "test_table" in locks       # Bug: always False (strings cannot be WKD keys)
False
>>> "test_table" in locks[source]  # Fix: correctly checks per-source dict
True
```

## Change

```diff
- if table in locks:
+ if table in locks[self]:
      lock = locks[self][table]
```

## Test plan

- [x] Added regression test `test_cached_reuses_per_table_lock` that verifies lock identity across calls
- [x] All 74 existing source tests pass
- [x] Verified independently that string keys always return `False` from `WeakKeyDictionary.__contains__`

## AI Disclosure

    Tools: Claude — used for planning the debugging approach. The bug identification, fix, and test were written and verified by me.

### DIFF
--- a/lumen/tests/sources/test_base.py
+++ b/lumen/tests/sources/test_base.py
@@ -1,6 +1,8 @@
 import asyncio
 import datetime as dt
 import os
+import threading
+import weakref
 
 from pathlib import Path
 
@@ -9,7 +11,7 @@ import pytest
 
 from hvplot.tests.util import makeMixedDataFrame
 
-from lumen.sources.base import Source
+from lumen.sources.base import Source, cached
 from lumen.state import state
 from lumen.transforms.sql import SQLLimit
 
@@ -199,11 +201,6 @@ def test_cached_reuses_per_table_lock(source):
     Since a string key can never exist in a WeakKeyDictionary, this caused
     a new RLock to be created on every call, defeating per-table locking.
     """
-    import threading
-    import weakref
-
-    from lumen.sources.base import cached
-
     locks = weakref.WeakKeyDictionary()
 
     def mock_get(self, table, **query):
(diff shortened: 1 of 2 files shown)


lumen/tests/sources/test_base.py
```diff
@@ -190,6 +190,46 @@ def test_extension_of_comlicated_url(source):
...
+    Regression test: the per-table lock lookup in cached() previously
+    checked `if table in locks` (the WeakKeyDictionary keyed by Source
+    instances) instead of `if table in locks[self]` (the per-source dict).
+    Since a string key can never exist in a WeakKeyDictionary, this caused
+    a new RLock to be created on every call, defeating per-table locking.
+    """
+    import threading
+    import weakref
+
+    from lumen.sources.base import cached
```
ahuang11: Can we move this up?

ahuang11 (commented):
Thanks for finding this issue! If you can address the one minor comment + the comment from GH Copilot, I'm happy to merge!

ghostiee-11:
Addressed both comments:

- Moved `threading`, `weakref`, and `cached` imports to the module level (top of file)
- Replaced the brittle `'RLock' in type().__name__` assertion with `isinstance(lock_first, type(threading.RLock()))`

Thanks for the review!

ahuang11 (approved):

ahuang11:
Thanks for updating it!

PR #1720, opened 2026-03-08
Fix: use `max()` for `merged_max` in `merge_schemas` to produce correct upper bound

SuMayaBee:
## Description

Fixes a bug in `merge_schemas()` in `lumen/util.py` where `min()` was incorrectly used to compute the merged upper bound (`inclusiveMaximum`).

When merging two schemas — for example one with range `[0, 100]` and another with `[0, 200]` — the merged maximum should be `200` (the broader union), but the code was returning `100` instead.

**Before:**
```python
# lumen/util.py, line 249
merged_max = min(schema['inclusiveMaximum'], old_schema['inclusiveMaximum'])
```

**After:**
```python
merged_max = max(schema['inclusiveMaximum'], old_schema['inclusiveMaximum'])
```

This caused filters and range sliders to clip to a smaller range than the data actually contains, silently hiding valid data from users.

## How Has This Been Tested?

The fix can be verified with the following snippet:

```python
from lumen.util import merge_schemas

old = {'type': 'integer', 'inclusiveMinimum': 0, 'inclusiveMaximum': 100}
new = {'type': 'integer', 'inclusiveMinimum': 0, 'inclusiveMaximum': 200}

result = merge_schemas(new, old)
assert result['inclusiveMaximum'] == 200  # was incorrectly 100 before fix
assert result['inclusiveMinimum'] == 0
```

## AI Disclosure

- [ ] This PR contains AI-generated content.
  - [ ] I have tested all AI-generated content in my PR.
  - [ ] I take responsibility for all AI-generated content in my PR.
    Tools: No AI tools are used. 

## Checklist

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -246,7 +246,7 @@ def merge_schemas(schema, old_schema):
         return dict(old_schema, enum=merged_enum)
     elif 'inclusiveMinimum' in schema and 'inclusiveMinimum' in old_schema:
         merged_min = min(schema['inclusiveMinimum'], old_schema['inclusiveMinimum'])
-        merged_max = min(schema['inclusiveMaximum'], old_schema['inclusiveMaximum'])
+        merged_max = max(schema['inclusiveMaximum'], old_schema['inclusiveMaximum'])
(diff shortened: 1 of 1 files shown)


ahuang11 (approved):
Great find!

PR #1721, opened 2026-03-08
fix: fix garbled `NullStep.stream` debug output

SuMayaBee:
## Description

Fixes a bug in `NullStep.stream()` in `lumen/ai/actor.py` where the f-string `f"[{text}"` had an **unclosed opening bracket**, producing garbled log output like `[Running analysis...` every time the fallback step was used.

**Before:**
```python
def stream(self, text, **kwargs):
    log_debug(f"[{text}")   # unclosed bracket → "[Running analysis..."
```

**After:**
```python
def stream(self, text, **kwargs):
    log_debug(f"[NullStep] {text}")   # clear, readable → "[NullStep] Running analysis..."
```

`NullStep` is used as the fallback whenever `self.interface is None` (e.g. in unit tests and minimal runs), so this garbled prefix appears in every debug log line emitted by those code paths.

## How Has This Been Tested?

The fix can be verified by instantiating a `NullStep` and calling `stream`:

```python
from lumen.ai.actor import NullStep

step = NullStep()
step.stream("Running analysis...")
# Before: logs "[Running analysis..."
# After:  logs "[NullStep] Running analysis..."
```

## AI Disclosure

- [ ] This PR contains AI-generated content.
  - [ ] I have tested all AI-generated content in my PR.
  - [ ] I take responsibility for all AI-generated content in my PR.
    Tools: No AI tools are used.

## Checklist

### DIFF
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -27,7 +27,7 @@ class NullStep:
         self.status = None
 
     def stream(self, text, **kwargs):
-        log_debug(f"[{text}")
+        log_debug(f"[NullStep] {text}")
 
 
 class LLMUser(param.Parameterized):


ahuang11 (approved):
Thanks!

PR #1722, opened 2026-03-08
refactor: introduce BaseOpenAIEmbeddings and normalize embed() whitespace handling

SuMayaBee:
## Description

`AzureOpenAIEmbeddings.embed()` was not stripping leading/trailing whitespace from input texts, unlike `OpenAIEmbeddings.embed()` which calls `.strip()`. This inconsistency means switching between providers can produce subtly different embeddings for the same input text, degrading similarity search in the vector store.

Following the reviewer's suggestion, instead of just patching the one-liner, a `BaseOpenAIEmbeddings` class was introduced that both `OpenAIEmbeddings` and `AzureOpenAIEmbeddings` now inherit from. The shared `embed()` method (with normalization) lives there — so there's a single source of truth and no duplication.

**Before:**
```python
# OpenAIEmbeddings
texts = [text.replace("\n", " ").strip() for text in texts]

# AzureOpenAIEmbeddings (missing .strip())
texts = [text.replace("\n", " ") for text in texts]
```

**After:**
```python
# BaseOpenAIEmbeddings (shared by both)
texts = [text.replace("\n", " ").strip() for text in texts]
```

## How Has This Been Tested?

```python
from lumen.ai.embeddings import AzureOpenAIEmbeddings, OpenAIEmbeddings

text = "  hello world  "

# Both now produce the same normalized input: "hello world"
assert text.replace("\n", " ").strip() == "hello world"
```

## AI Disclosure

    Tools: Github copilot

## Checklist

### DIFF
--- a/lumen/ai/embeddings.py
+++ b/lumen/ai/embeddings.py
@@ -124,8 +124,7 @@ class AzureOpenAIEmbeddings(Embeddings, AzureOpenAIMixin):
         self.client = self._instantiate_client(async_client=True)
 
     async def embed(self, texts: list[str]) -> list[list[float]]:
-        # Same normalization as OpenAIEmbeddings.embed()
-        texts = [text.replace("\n", " ").strip() for text in texts]
+        texts = [text.replace("\n", " ").strip() for text in texts] # Same normalization as OpenAIEmbeddings.embed()
         response = await self.client.embeddings.create(input=texts, model=self.model)
         return [r.embedding for r in response.data]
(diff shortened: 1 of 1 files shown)


ahuang11 (approved):
Thanks. Conflicted between `BaseOpenAIEmbeddings` vs `OpenAIEmbeddingsMixin`

What do you think is more suitable?

Separately, can you update the PR title to reflect changes?

PR #1733, opened 2026-03-10
Fix multimodal content handling for non-vision LLM providers

ghostiee-11:
## Description

When using a non-vision LLM provider (e.g. Groq, local LLMs via OpenAI-compatible endpoints), follow-up queries after a visualization fail with:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'messages[N].content must be a string', 'type': 'invalid_request_error'}}
```

**Root cause**: `_check_for_image` in `Llm` replaces the entire list content `[str, Image]` with just the serialized Image object, discarding the text parts. When the provider doesn't support multimodal content, it rejects the non-string `content` field.

**Impact**:
- With `instructor` retries: errors are masked but waste API tokens and add latency on every follow-up after a visualization
- With `litellm` router: causes a full crash — "Internal execution error during planning stage"
- Text context paired with the image is silently lost

### Before (server logs show repeated 400 errors on follow-up):
```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'messages[2].content must be a string', ...}}
openai.BadRequestError: Error code: 400 - {'error': {'message': 'messages[4].content must be a string', ...}}
```

### After (zero errors, follow-up works cleanly):
Server logs show no `content must be a string` errors. Follow-up queries after visualization complete without retries.

## Changes

Three targeted fixes in `lumen/ai/llm.py`:

1. **`_check_for_image`**: Now preserves all items in mixed `[str, Image]` lists instead of replacing the entire list with just the last image.

2. **New `_strip_images` static method**: Removes image content from messages while keeping text. Used as a fallback when a provider rejects multimodal content.

3. **`run_client` retry logic**: Catches `"content must be a string"` errors, strips images from messages, and retries — so non-vision providers gracefully degrade instead of crashing.

## How Has This Been Tested?

- 13 new unit tests added covering `_check_for_image` and `_strip_images` (5 + 8 tests)
- All 15 tests in `test_llm.py` pass
- Live tested with Groq (`llama-3.3-70b-versatile`) via OpenAI-compatible endpoint:
  - Loaded penguins dataset → created bar chart → asked follow-up "change colors"
  - Before fix: server logs show `messages[N].content must be a string` errors
  - After fix: zero errors in server logs, follow-up completes cleanly

## AI Disclosure

I identified this bug during manual QA testing of Lumen AI with Groq as the LLM provider. I traced the root cause through server logs to `_check_for_image`, designed the fix approach, and used AI tooling to help scaffold the implementation and tests. All changes were reviewed and verified by me.

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -278,11 +278,41 @@ class Llm(param.Parameterized):
                 contains_image = True
 
             elif isinstance(content, list):
+                new_content = []
                 for item in content:
                     if isinstance(item, (Image, pn.pane.image.ImageBase)):
-                        messages[i]["content"] = self._serialize_image_pane(item)
+                        new_content.append(self._serialize_image_pane(item))
                         contains_image = True
+                    else:
+                        new_content.append(item)
+                messages[i]["content"] = new_content
         return messages, contains_image
+
+    @staticmethod
+    def _strip_images(messages: list[Message]) -> list[Message]:
+        """
+        Remove image content from messages, keeping only text.
+
+        Used as a fallback when a provider does not support multimodal
+        content (e.g. non-vision models or providers that require
+        ``content`` to be a plain string).
+        """
+        stripped = []
+        for msg in messages:
+            content = msg.get("content")
+            if isinstance(content, Image):
+                stripped.append({**msg, "content": "(image omitted)"})
+            elif isinstance(content, list):
+                text_parts = [
+                    item for item in content if isinstance(item, str)
+                ]
+                stripped.append({
+                    **msg,
+                    "content": "\n".join(text_parts) if text_parts else "(image omitted)",
+                })
+            else:
+                stripped.append(msg)
+        return stripped
 
     @classmethod
     def warmup(cls, model_kwargs: dict | None):
@@ -783,7 +813,18 @@ class Llm(param.Parameterized):
             previous_role = role
 
         client = await self.get_client(model_spec, **kwargs)
-        result = await client(messages=messages, **kwargs)
+        try:
+            result = await client(messages=messages, **kwargs)
+        except Exception as e:
+            if "content must be a string" in str(e):
+                log_debug(
+                    "Provider rejected multimodal content; "
+                    "stripping images and retrying."
+                )
+                messages = self._strip_images(messages)
+                result = await client(messages=messages, **kwargs)
+            else:
+                raise
         if response_model := kwargs.get("response_model"):
             log_debug(f"Response model: \033[93m{response_model.__name__!r}\033[0m")
         log_debug(f"LLM Response: \033[95m{truncate_string(str(result), max_length=1000)}\033[0m\n---")
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -269,29 +269,49 @@ class Llm(param.Parameterized):
             image = Image.from_url(image_object)
         return image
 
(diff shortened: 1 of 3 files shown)


ghostiee-11:
Yup right, the strip-after approach is a bit redundant. The cleanest option is the one you suggested: prepare both versions upfront and pass them to run_client.

Something like:

- _check_for_image stays as a pure bool check (no message mutation)
- A new _prepare_image_messages builds the image-injected version
- run_client receives both messages (plain) and messages_with_images (or None if no images), tries with images first, falls back to plain on failure

That way _check_for_image has a single responsibility, no message is mutated in place, and the retry path is straightforward. I'll rework the PR with this approach.

lumen/ai/llm.py
```diff
@@ -269,20 +269,70 @@ def _serialize_image_pane(self, image: pn.pane.image.ImageBase | Image) -> Image
...
+        """
+        Build a copy of *messages* with image content serialized.
+
+        Returns a new list; the originals are not mutated.
+        """
+        prepared = []
+        for msg in messages:
+            content = msg.get("content")
+            if isinstance(content, (Image, pn.pane.image.ImageBase)):
+                prepared.append({**msg, "content": self._serialize_image_pane(content)})
```
ahuang11: If msg only contains role and content, is there a need for **msg rather than a more readable pattern, e.g.

```python
{"role": msg["role"], "content": self._serialize_image_pane(content)}
```
Same below

ahuang11 (changes requested):
```
        output = await self.run_client(
            model_spec, messages, image_messages=image_messages,
            plain_messages=plain_messages, **kwargs
        )
```

Not a fan of this. I mistakened messages as plain_messages without messages. I suspect there are other classes that are broken now too. Perhaps as a util method that does the try/except inside `run_client` instead is a better approach.

`result = await client(messages=messages, **kwargs)`

Alternatively, we might be overcomplicating things because if we try/except, we're essentially making two client calls. Maybe it's just better to have a flag that lets users toggle whether to allow images or not in messages; if not, just return early from _prepare_vision_messages.

ghostiee-11:
You're right, the `image_messages`/`plain_messages` approach overcomplicates `run_client`. Reworked with the flag approach instead:

- Added `_supports_vision` class attribute on `Llm` (defaults to `True`). When `False`, `_prepare_vision_messages` returns early without injecting images, so the problem never arises downstream.
- Removed `_prepare_image_messages`, `_strip_images`, and the `image_messages`/`plain_messages` params from `run_client` entirely. No try/except, no double API calls, clean signature.
- The `_check_for_image` fix for preserving list items (the original bug) is still included. Previously it replaced the entire `[text, Image]` list with just the last serialized image, discarding the text.

One thing worth noting: the flag solves the case where the user explicitly knows their provider doesn't support vision. But the original bug scenario, someone pointing the `OpenAI` provider at Groq without knowing to set `_supports_vision = False`, would still hit the "content must be a string" error since vision defaults to `True`. The try/except approach auto-detected that, but I agree it was too costly (double API calls, polluted signature). Maybe the right middle ground is documenting it clearly, or should we consider auto-detecting provider capabilities in a follow-up?

Also, should `_supports_vision` be a `param.Boolean` instead of a private class attribute? That way users could configure it per-instance (e.g. via YAML) rather than needing to subclass. Happy to change that if you prefer.

ahuang11 (approved):
Much appreciated with the changes and tests!

PR #1736, opened 2026-03-11
Fix download done_callback: or -> and prevents crash and wrong message

ghostiee-11:
## Description

The done_callback in `DownloadControls._handle_urls` had two bugs caused by incorrect boolean logic (`or` instead of `and`), plus structural issues with duplicated cleanup logic scattered across multiple methods.

### Bug 1: CancelledError crash

When a download task is cancelled, `not t.cancelled()` evaluates to `False`. With `or`, Python evaluates the right side `t.exception()`. But calling `.exception()` on a cancelled task raises `asyncio.CancelledError`, crashing the callback.

### Bug 2: "Download complete." shown on failure

When a download fails with an exception, `not t.cancelled()` is `True`. With `or`, `True or ...` short-circuits to `True`, so the callback shows "Download complete." even though the download failed.

### What this PR does

1. **Extracts the inline lambda into a proper `_on_download_done` method** with explicit `if/elif` handling all three task states:
   - **Cancelled**: shows "Download cancelled."
   - **Failed**: shows "Download failed: {error}" (previously silent or showed "complete")
   - **Success**: no-op, preserves the more informative message set by `_download_and_process_urls` ("Successfully downloaded X file(s). Click 'Confirm file(s)' to process.")

2. **Adds a stale task guard** (`if task is not self._active_download_task: return`) so callbacks from replaced downloads don't interfere with newer ones.

3. **Consolidates cleanup into `_on_download_done` as single source of truth**, removes redundant `_active_download_task = None` and `progress.clear()` from `_handle_cancel` and `_download_and_process_urls`, eliminating race conditions between duplicate cleanup paths.

### Truth table

| Task state | `not t.cancelled()` | `t.exception() is None` | Old `or` (bug) | New behavior |
|---|---|---|---|---|
| Success | True | True | True (correct) | No-op, preserves coroutine message |
| Cancelled | False | *CancelledError* | **CRASH** | Shows "Download cancelled." |
| Failed | True | False | **True (wrong)** | Shows "Download failed: ..." |

### Before / After proof
The issue :

https://github.com/user-attachments/assets/9b133946-2c55-4d4a-8055-d03a3fda3e55

After  :

https://github.com/user-attachments/assets/04b25f37-9e72-4b06-8edd-2fc20c25e5ad

```
=== BEFORE (inline lambda with or) - BUGGY ===

Bug 1: Cancelled task
  Code: not t.cancelled() or t.exception() is None
  CRASH: asyncio.CancelledError raised!
  (calling t.exception() on cancelled task crashes)

Bug 2: Failed task (download error)
  Code: not t.cancelled() or t.exception() is None
  Result: True
  User sees: "Download complete." (WRONG - download failed!)

=== AFTER (_on_download_done method) - FIXED ===

Fix 1: Cancelled task
  -> Shows: "Download cancelled." (correct, no crash)

Fix 2: Failed task (download error)
  -> Shows: "Download failed: network timeout" (correct)

Fix 3: Successful task
  -> No-op, preserves: "Successfully downloaded 1 file(s). Click Confirm..." (correct)
```

### How I found this

I was reviewing the async callback patterns in the codebase and noticed the boolean logic was incorrect. I verified by tracing all task state transitions and confirmed both bugs.

## How Has This Been Tested?

Added 4 async unit tests in `test_download.py` that exercise the actual `_on_download_done` method via the `download_controls` fixture:
- `test_successful_task_preserves_coroutine_message` - callback doesn't overwrite the call-to-action message
- `test_cancelled_task_shows_cancelled` - shows cancel message without crashing
- `test_failed_task_shows_error` - shows error message, not "Download complete."
- `test_stale_task_is_ignored` - replaced task's callback is a no-op

All 26 download tests pass (22 existing + 4 new).

## Checklist

- [x] Tests added and passing
- [ ] Added documentation

## AI Disclosure

I planned, identified, and tested this fix myself. AI was used to assist with code review and verification.

### DIFF
--- a/lumen/ai/controls/download.py
+++ b/lumen/ai/controls/download.py
@@ -70,31 +70,39 @@ class DownloadControls(BaseSourceControls):
         )
 
     def _handle_cancel(self, event):
-        """Handle cancel button click by cancelling active download task"""
+        """Handle cancel button click by cancelling active download task.
+
+        Actual cleanup and messaging is handled by ``_on_download_done``
+        when the cancelled task's done callback fires.
+        """
         if self._active_download_task and not self._active_download_task.done():
             self._active_download_task.cancel()
-            self.progress.clear()
-            self._message_placeholder.param.update(
-                object="Download cancelled by user.",
-                visible=True
-            )
-        # Clear the active task so Cancel button hides
-        self._active_download_task = None
 
     def _on_download_done(self, task):
-        """Handle download task completion, cancellation, or failure."""
+        """Handle download task completion, cancellation, or failure.
+
+        Only acts on the currently active download task to avoid stale
+        callbacks from interfering with a newer download.  The success
+        path is intentionally a no-op because ``_download_and_process_urls``
+        already sets a more informative message with a call-to-action.
+        """
+        if task is not self._active_download_task:
+            return
+
         if task.cancelled():
             self._message_placeholder.param.update(
                 object="Download cancelled.",
                 visible=True
             )
         elif task.exception() is not None:
-            return
-        else:
             self._message_placeholder.param.update(
-                object="Download complete.",
+                object=f"Download failed: {task.exception()}",
                 visible=True
             )
+
+        # Clean up so the Cancel button hides for finished tasks
+        self.progress.clear()
+        self._active_download_task = None
 
     def _is_valid_url(self, url):
         """Basic URL validation"""
@@ -338,9 +346,6 @@ class DownloadControls(BaseSourceControls):
             self._message_placeholder.object = f"Successfully downloaded {success_count} file(s). Click 'Confirm file(s)' to process."
         else:
             self._message_placeholder.visible = False
-
-        # Clear the active task so Cancel button hides
-        self._active_download_task = None
 
     @param.depends("add", watch=True)
     def _on_add(self):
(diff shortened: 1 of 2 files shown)


lumen/ai/controls/download.py
```diff
@@ -70,15 +70,38 @@ def _render_layout(self):
...
+            return
+
+        if task.cancelled():
             self._message_placeholder.param.update(
-                object="Download cancelled by user.",
+                object="Download cancelled.",
                 visible=True
             )
-        # Clear the active task so Cancel button hides
+        elif task.exception() is not None:
```
philippjfr, line 97: Don't we need an else branch to add the "Download complete." message?

ghostiee-11: The success path is intentionally a no-op because `_download_and_process_urls` already sets a more informative message at line 346 before the callback fires:

```python
self._message_placeholder.object = f"Successfully downloaded {success_count} file(s). Click 'Confirm file(s)' to process."
```
Screenshot 2026-03-12 at 9 39 09 PM


Adding "Download complete." in the else branch would overwrite that call-to-action. But if you'd prefer all messaging consolidated in the callback, I can move it there.

philippjfr: Ah sweet, in that case ignore me.

philippjfr (approved):

lumen/ai/controls/download.py
ghostiee-11: Yeahh.. No issues. Thanks for the merge

PR #1739, opened 2026-03-11
Fix annotation feature failing with Google Gemini provider

ghostiee-11:
## Description

I discovered while testing Lumen AI with Google Gemini that clicking the "Add annotations" button on a Vega-Lite chart always fails. I traced three issues in the annotation pipeline:

1. **`AnnotationControls._annotate()` did not pass `self.view` to `actor.annotate()`**, so `_prepare_vision_messages()` received `None` and skipped adding the chart image. This left the LLM with no user messages, causing Google Gemini to reject the request with *"At least one user message must be included"*.

2. **`_check_for_image()` in `llm.py` destroyed list content** when serializing images. A message with content `[text, Image]` was replaced with just the serialized `Image`, losing the text instruction entirely.

3. **`Google.run_client()` let instructor auto-detect image content** and apply `HARM_CATEGORY_IMAGE_*` safety settings that are not supported by the Google GenAI v1beta API, causing a `400 INVALID_ARGUMENT` error.

### Before (error when clicking annotate)
before-annotation-bug-2

### After (annotation applied successfully)
after-annotation-working

Screenshot 2026-03-11 at 1 44 42 PM

## Fixes

- Pass `self.view` as 5th argument to `annotate()` in `revision.py`
- Add a text-only user message in `vega_lite.py` `annotate()` before `_prepare_vision_messages()` so the instruction text survives serialization
- Preserve all list items in `_check_for_image()` instead of overwriting the list with a single serialized image
- Pass `safety_settings=[]` in `Google.run_client()` to bypass instructor unsupported image safety categories

## How Has This Been Tested?

- Added regression test `test_annotate_passes_view_to_actor` verifying the view is passed as the 5th argument
- All 317 AI tests pass
- Manually tested end-to-end: uploaded CSV, created bar chart, clicked "Add annotations", typed "Highlight the highest revenue product" -- annotation was successfully applied with "Top Seller ($45.1k)" label

## Checklist

- [x] Tests added and passing
- [ ] Added documentation

## AI Disclosure

I identified the bug through hands-on testing of Lumen AI with Google Gemini, traced the root cause through the code, and designed the fix. AI tools assisted with scaffolding the implementation.

### DIFF
--- a/lumen/ai/agents/vega_lite.py
+++ b/lumen/ai/agents/vega_lite.py
@@ -741,7 +741,7 @@ class VegaLiteAgent(BaseCodeAgent):
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Revise this chart: {instruction!r}")
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
 
         vega_spec = dump_yaml(spec["spec"], default_flow_style=False)
         result = await self._invoke_prompt(
@@ -749,6 +749,7 @@ class VegaLiteAgent(BaseCodeAgent):
             messages,
             context,
             vega_spec=vega_spec,
+            instruction=instruction,
         )
         update_dict = load_yaml(result.yaml_update)
 
--- a/lumen/ai/agents/vega_lite.py
+++ b/lumen/ai/agents/vega_lite.py
@@ -297,16 +297,14 @@ class VegaLiteAgent(BaseCodeAgent):
             return None
 
     def _prepare_vision_messages(
-        self, messages: list[Message], out: LumenEditor | None, content: str,
-        text_fallback: bool = False
+        self, messages: list[Message], out: LumenEditor | None, content: str
     ) -> list[Message]:
         """Add plot image to messages for LLM vision analysis.
 
-        When *text_fallback* is True the content string is always
-        appended as a plain-text user message if vision is unavailable
-        or image export fails, so the instruction is never lost.
+        If vision is unavailable or image export fails we append the
+        content as a plain-text message.
         """
-        fallback = messages + [{"role": "user", "content": content}] if text_fallback else messages
+        fallback = messages + [{"role": "user", "content": content}]
         if not self.llm._supports_vision:
             return fallback
         if out is None or not isinstance(out, VegaLiteEditor):
@@ -748,7 +746,7 @@ class VegaLiteAgent(BaseCodeAgent):
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}", text_fallback=True)
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
 
         vega_spec = dump_yaml(spec["spec"], default_flow_style=False)
         result = await self._invoke_prompt(
--- a/lumen/ai/agents/vega_lite.py
+++ b/lumen/ai/agents/vega_lite.py
@@ -297,17 +297,24 @@ class VegaLiteAgent(BaseCodeAgent):
             return None
 
     def _prepare_vision_messages(
-        self, messages: list[Message], out: LumenEditor | None, content: str
+        self, messages: list[Message], out: LumenEditor | None, content: str,
+        text_fallback: bool = False
     ) -> list[Message]:
-        """Add plot image to messages for LLM vision analysis."""
+        """Add plot image to messages for LLM vision analysis.
+
+        When *text_fallback* is True the content string is always
+        appended as a plain-text user message if vision is unavailable
+        or image export fails, so the instruction is never lost.
+        """
+        fallback = messages + [{"role": "user", "content": content}] if text_fallback else messages
         if not self.llm._supports_vision:
-            return messages
+            return fallback
         if out is None or not isinstance(out, VegaLiteEditor):
-            return messages
+            return fallback
 
         image_bytes = self._export_plot_image(out)
         if image_bytes is None:
-            return messages
+            return fallback
 
         base64_str = base64.b64encode(image_bytes).decode('utf-8')
         plot_image = Image.from_raw_base64(base64_str)
@@ -741,7 +748,7 @@ class VegaLiteAgent(BaseCodeAgent):
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}", text_fallback=True)
 
         vega_spec = dump_yaml(spec["spec"], default_flow_style=False)
         result = await self._invoke_prompt(
@@ -749,7 +756,6 @@ class VegaLiteAgent(BaseCodeAgent):
             messages,
             context,
             vega_spec=vega_spec,
-            instruction=instruction,
         )
         update_dict = load_yaml(result.yaml_update)
 
--- a/lumen/tests/ai/test_llm.py
+++ b/lumen/tests/ai/test_llm.py
@@ -302,18 +302,11 @@ class TestPrepareVisionMessages:
             agent.llm = llm
         return agent
 
-    def test_no_vision_no_fallback(self, agent):
-        """Without text_fallback, returns messages unchanged when vision unsupported."""
-        agent.llm._supports_vision = False
-        msgs = [{"role": "user", "content": "hello"}]
-        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
-        assert result is msgs
-
     def test_no_vision_with_fallback(self, agent):
         """With text_fallback, appends text-only message when vision unsupported."""
         agent.llm._supports_vision = False
         msgs = [{"role": "user", "content": "hello"}]
-        result = agent._prepare_vision_messages(msgs, None, "Annotate this", text_fallback=True)
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
         assert len(result) == 2
         assert result[1] == {"role": "user", "content": "Annotate this"}
 
@@ -328,20 +321,9 @@ class TestPrepareVisionMessages:
         """With text_fallback, appends text-only message when editor is None."""
         agent.llm._supports_vision = True
         msgs = [{"role": "user", "content": "hello"}]
-        result = agent._prepare_vision_messages(msgs, None, "Annotate this", text_fallback=True)
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
         assert len(result) == 2
         assert result[1] == {"role": "user", "content": "Annotate this"}
-
-    def test_image_export_fails_no_fallback(self, agent):
-        """Without text_fallback, returns messages unchanged when image export fails."""
-        agent.llm._supports_vision = True
-        mock_editor = MagicMock()
-        mock_editor.__class__ = type("VegaLiteEditor", (), {})
-        msgs = [{"role": "user", "content": "hello"}]
-        with patch.object(agent, '_export_plot_image', return_value=None):
-            with patch('lumen.ai.agents.vega_lite.VegaLiteEditor', mock_editor.__class__):
-                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this")
-        assert result is msgs
 
     def test_image_export_fails_with_fallback(self, agent):
         """With text_fallback, appends text-only message when image export fails."""
@@ -351,7 +333,7 @@ class TestPrepareVisionMessages:
         msgs = [{"role": "user", "content": "hello"}]
         with patch.object(agent, '_export_plot_image', return_value=None):
             with patch('lumen.ai.agents.vega_lite.VegaLiteEditor', mock_editor.__class__):
-                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this", text_fallback=True)
+                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this")
         assert len(result) == 2
         assert result[1] == {"role": "user", "content": "Annotate this"}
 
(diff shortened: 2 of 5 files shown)


lumen/ai/agents/vega_lite.py
```diff
@@ -739,7 +739,10 @@ async def annotate(
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Revise this chart: {instruction!r}")
+        # Preserve existing chat history and append the new annotation instruction
+        messages = list(messages or [])
+        messages.append({"role": "user", "content": f"Annotate this chart: {instruction!r}"})
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
```
ahuang11: Not sure I want to preserve existing chat history since annotation step might not require original message.

ghostiee-11: Makes sense! The annotation step receives a fresh instruction. It doesn't need prior chat context. Let me fix this

lumen/ai/llm.py
```diff
@@ -1852,6 +1856,22 @@ async def run_client(self, model_spec: str | dict, messages: list[Message], **kw
...
+                        return True
+                    if isinstance(content, list):
+                        for part in content:
+                            if isinstance(part, Image):
+                                return True
+                return False
+
+            if "safety_settings" not in kwargs and _messages_contain_image(messages):
+                kwargs.setdefault("safety_settings", [])
+
```
ahuang11: Let's remove this. I think that is related to an older instructor/gemini version. Once you update it, it should work as expected.
```suggestion
```

ghostiee-11: Thanks for the review! Removed the safety_settings workaround also...

lumen/ai/agents/vega_lite.py
```diff
@@ -739,7 +739,8 @@ async def annotate(
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Revise this chart: {instruction!r}")
+        messages = [{"role": "user", "content": f"Annotate this chart: {instruction!r}"}]
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
```
ahuang11: I don't think the fix is supposed to be here...

```python

    def _prepare_vision_messages(
        self, messages: list[Message], out: LumenEditor | None, content: str
    ) -> list[Message]:
        """Add plot image to messages for LLM vision analysis."""
        if out is None or not isinstance(out, VegaLiteEditor):
            return messages

        image_bytes = self._export_plot_image(out)
        if image_bytes is None:
            return messages

        base64_str = base64.b64encode(image_bytes).decode('utf-8')
        plot_image = Image.from_raw_base64(base64_str)
        log_debug("Added plot image to messages for LLM vision analysis")
        return messages + [{
            "role": "user",
            "content": [content, plot_image]
        }]
```

ahuang11: Actually, is this change still necessary if we're passing self.view?

> AnnotationControls._annotate() did not pass self.view to actor.annotate(), so _prepare_vision_messages() received None and skipped adding the chart image. This left the LLM with no user messages, causing Google Gemini to reject the request with "At least one user message must be included".

ahuang11 (changes requested):

lumen/ai/agents/vega_lite.py
ghostiee-11: You're right, `_prepare_vision_messages` already adds the instruction alongside the image. But one gap: `annotate()` doesn't pass `instruction` to `_invoke_prompt` (unlike `revise()` which passes `feedback=instruction`), so if vision fails the LLM never sees the instruction. I'll pass `instruction` to `_invoke_prompt` and add it to `annotate_plot.jinja2`, then remove the seeded message.

ghostiee-11: Not once I wire instruction through the template. Will fix in next push.

lumen/tests/ai/test_llm.py
```diff
@@ -180,9 +182,33 @@ async def test_azure_open_ai_get_model_kwargs_individual_models():
...
+    ]
+
+    result, contains_image = llm._check_for_image(messages)
+
+    assert contains_image is True
+    content = result[0]["content"]
+    assert isinstance(content, list)
+    assert len(content) == 2
+    assert content[0] == "Annotate this chart"
+    assert content[1] == serialized_sentinel
```
ahuang11: This should be merged into the class below, and the import should be moved up.

ahuang11 (changes requested):
Sorry, one last change requested

lumen/tests/ai/test_llm.py
ghostiee-11: Suree

lumen/tests/ai/test_llm.py
```diff
@@ -180,6 +180,7 @@ async def test_azure_open_ai_get_model_kwargs_individual_models():
     assert llm._get_model_kwargs("other") == model_kwargs["other"]
 
 
+<<<<<<< HEAD
```
ahuang11:
```suggestion
```

ghostiee-11: Yup forgot to remove this. I am sorry

lumen/ai/agents/vega_lite.py
```diff
@@ -741,14 +741,15 @@ async def annotate(
...
 
         vega_spec = dump_yaml(spec["spec"], default_flow_style=False)
         result = await self._invoke_prompt(
             "annotate_plot",
             messages,
             context,
             vega_spec=vega_spec,
+            instruction=instruction,
         )
```
ahuang11, line 757: I'm not exactly following why we need to inject instructions into the system prompt. Can you help me understand?

ghostiee-11: `_prepare_vision_messages()` adds the instruction to the user messages, but it returns messages unchanged if the LLM doesn't support vision or if image export fails. The task history (`messages`) only has the original chart creation query, not the annotation instruction. So without it in the system prompt, the LLM would know how to annotate (from the template) and have the spec, but not what the user actually wants annotated in those fallback cases.

ahuang11: > but it returns messages unchanged if the LLM doesn't support vision or if image export fails

So I think it's better to handle it there instead if it's not too complex / if I'm not mistaken. Else, we're duplicating instructions (tokens) in both system and user message.

ghostiee-11: Makes sense\! I'll handle it in `_prepare_vision_messages` instead - in the fallback paths where it currently returns messages unchanged, I'll append a text-only user message with the instruction. Then I can drop it from the template so it's not duplicated. Should be a small change.

lumen/ai/agents/vega_lite.py
```diff
@@ -297,17 +297,24 @@ def _export_plot_image(self, out: VegaLiteEditor) -> bytes | None:
             return None
 
     def _prepare_vision_messages(
-        self, messages: list[Message], out: LumenEditor | None, content: str
+        self, messages: list[Message], out: LumenEditor | None, content: str,
+        text_fallback: bool = False
```
ahuang11: Do we need a flag, i.e. is there a good reason to not always fallback?

philippjfr: Agreed.

lumen/ai/agents/vega_lite.py
```diff
@@ -741,7 +748,7 @@ async def annotate(
         str
             Updated specification with annotations
         """
-        messages = self._prepare_vision_messages(messages, view, f"Revise this chart: {instruction!r}")
+        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}", text_fallback=True)
```
philippjfr:
```suggestion
        messages = self._prepare_vision_messages(messages, view, f"Annotate this chart: {instruction!r}")
```

lumen/ai/agents/vega_lite.py
philippjfr:
```suggestion
        self, messages: list[Message], out: LumenEditor | None, content: str
```

lumen/ai/agents/vega_lite.py
```diff
@@ -297,17 +297,24 @@ def _export_plot_image(self, out: VegaLiteEditor) -> bytes | None:
...
+        text_fallback: bool = False
     ) -> list[Message]:
-        """Add plot image to messages for LLM vision analysis."""
+        """Add plot image to messages for LLM vision analysis.
+
+        When *text_fallback* is True the content string is always
+        appended as a plain-text user message if vision is unavailable
+        or image export fails, so the instruction is never lost.
+        """
+        fallback = messages + [{"role": "user", "content": content}] if text_fallback else messages
```
philippjfr:
```suggestion

        If vision is unavailable or image export fails we append the
        content as a plain-text message.
        """
        fallback = messages + [{"role": "user", "content": content}]
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+        """Without text_fallback, returns messages unchanged when image export fails."""
+        agent.llm._supports_vision = True
+        mock_editor = MagicMock()
+        mock_editor.__class__ = type("VegaLiteEditor", (), {})
+        msgs = [{"role": "user", "content": "hello"}]
+        with patch.object(agent, '_export_plot_image', return_value=None):
+            with patch('lumen.ai.agents.vega_lite.VegaLiteEditor', mock_editor.__class__):
+                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this")
+        assert result is msgs
+
```
philippjfr:
```suggestion
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+    def test_image_export_fails_with_fallback(self, agent):
+        """With text_fallback, appends text-only message when image export fails."""
+        agent.llm._supports_vision = True
+        mock_editor = MagicMock()
+        mock_editor.__class__ = type("VegaLiteEditor", (), {})
+        msgs = [{"role": "user", "content": "hello"}]
+        with patch.object(agent, '_export_plot_image', return_value=None):
+            with patch('lumen.ai.agents.vega_lite.VegaLiteEditor', mock_editor.__class__):
+                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this", text_fallback=True)
```
philippjfr:
```suggestion
                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this")
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+        fake_png = (
+            b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01'
+            b'\x00\x00\x00\x01\x08\x02\x00\x00\x00\x90wS\xde\x00'
+            b'\x00\x00\x0cIDATx\x9cc\xf8\x0f\x00\x00\x01\x01\x00'
+            b'\x05\x18\xd8N\x00\x00\x00\x00IEND\xaeB`\x82'
+        )
+        msgs = [{"role": "user", "content": "hello"}]
+        with patch.object(agent, '_export_plot_image', return_value=fake_png):
+            with patch('lumen.ai.agents.vega_lite.VegaLiteEditor', mock_editor.__class__):
+                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this", text_fallback=True)
```
philippjfr:
```suggestion
                result = agent._prepare_vision_messages(msgs, mock_editor, "Annotate this")
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+        agent.llm._supports_vision = True
+        msgs = [{"role": "user", "content": "hello"}]
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
+        assert result is msgs
+
+    def test_no_editor_with_fallback(self, agent):
+        """With text_fallback, appends text-only message when editor is None."""
+        agent.llm._supports_vision = True
+        msgs = [{"role": "user", "content": "hello"}]
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this", text_fallback=True)
```
philippjfr:
```suggestion
        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+        agent.llm._supports_vision = False
+        msgs = [{"role": "user", "content": "hello"}]
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
+        assert result is msgs
+
+    def test_no_vision_with_fallback(self, agent):
+        """With text_fallback, appends text-only message when vision unsupported."""
+        agent.llm._supports_vision = False
+        msgs = [{"role": "user", "content": "hello"}]
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this", text_fallback=True)
```
philippjfr:
```suggestion
        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
```

lumen/tests/ai/test_llm.py
```diff
@@ -280,3 +283,96 @@ def test_multiple_images_in_list(self, llm):
...
+            agent.llm = llm
+        return agent
+
+    def test_no_vision_no_fallback(self, agent):
+        """Without text_fallback, returns messages unchanged when vision unsupported."""
+        agent.llm._supports_vision = False
+        msgs = [{"role": "user", "content": "hello"}]
+        result = agent._prepare_vision_messages(msgs, None, "Annotate this")
+        assert result is msgs
+
```
philippjfr:
```suggestion
```

Issue #1508, opened 2025-12-09
Add xarray-sql support


ahuang11:
Restart https://github.com/holoviz/lumen/pull/1434

ghostiee-11:
@ahuang11  Ig, This can be closed too

PR #1741, opened 2026-03-11
Add XArraySQLSource for N-dimensional scientific data

ghostiee-11:
## Description

Adds `XArraySQLSource` so Lumen can query xarray datasets (NetCDF, Zarr, HDF5, GRIB) with SQL via Apache DataFusion. Builds on the proof-of-concept from #1434 and addresses #1508.

Each data variable in the dataset becomes a SQL table with coordinate columns. For example, a NetCDF file with an `air` variable on `(time, lat, lon)` dimensions gives you a table you can query like:

```python
source = XArraySQLSource(uri="air_temperature.nc")
source.execute("SELECT lat, AVG(air) FROM air GROUP BY lat")
```

Inherits from `BaseSQLSource` to reuse schema inference, caching, SQL transforms, and `create_sql_expr_source`. Optional dependency via `pip install lumen[xarray]`.

### After

Tested with NOAA air temperature dataset (3.8M rows):

xarray_demo_full

## How Has This Been Tested?

58 tests covering construction (from dataset, file, zarr), SQL execution, schema/metadata, table normalization, serialization roundtrip, async, and resource cleanup.

```bash
pytest lumen/tests/sources/test_xarray_sql.py -v
```

## AI Disclosure

I planned the architecture and integration strategy. AI tools helped scaffold the implementation and tests based on my design decisions. I reviewed and tested all code manually.


lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,710 @@
...
+                    else:
+                        cast = lambda v: v
+                    min_val = mm[f"{col}_min"].iloc[0]
+                    max_val = mm[f"{col}_max"].iloc[0]
+                    schema[col]["inclusiveMinimum"] = min_val if pd.isna(min_val) else cast(min_val)
+                    schema[col]["inclusiveMaximum"] = max_val if pd.isna(max_val) else cast(max_val)
+
+            schemas[entry] = schema
+
+        return schemas if table is None else schemas[table]
```
ahuang11: Not a big fan of duplication. Why are we redefining get_schema?

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,710 @@
...
+            if isinstance(val, list):
+                vals_str = ", ".join(_sql_literal(v) for v in val)
+                conditions.append(f"{qcol} IN ({vals_str})")
+            elif isinstance(val, slice):
+                if val.start is not None:
+                    conditions.append(f"{qcol} >= {_sql_literal(val.start)}")
+                if val.stop is not None:
+                    conditions.append(f"{qcol} <= {_sql_literal(val.stop)}")
+            else:
+                conditions.append(f"{qcol} = {_sql_literal(val)}")
```
ahuang11: I don't understand this part.

ahuang11 (changes requested):
Didn't read through it completely; I think there's a lot of duplication instead of just properly inheriting from parent.

ahuang11:
Thanks I'll review when time allows.

In the future, there's no need to tag me or other maintainers. We'll review as soon as we can. For reference: https://holoviz.org/contribute.html#using-ai-readme

lumen/sources/xarray_sql.py
ghostiee-11: This is the xarray-specific override of `Source._get_table_metadata()`, which is called by the inherited `get_metadata()` method. DuckDBSource has the same pattern (its own `_get_table_metadata` at line 535).

For xarray datasets, the standard column-name-and-type metadata isn't very useful for AI agents. This override extracts xarray-specific attributes like units, dimension sizes, coordinate ranges, and shape so that `get_metadata("temperature")` returns something like:

```python
{
    "description": "Air Temperature [K] | dimensions: time(2920) x lat(25) x lon(53)",
    "columns": {
        "lat": {"data_type": "float64", "units": "degrees_north", "min": "15.0", "max": "75.0", ...},
        "temperature": {"data_type": "float32", "units": "K", ...}
    },
    "dimensions": ["time", "lat", "lon"],
    "shape": [2920, 25, 53]
}
```

This gives Lumen AI agents enough context to generate meaningful SQL queries against scientific data without needing to sample the dataset first.

lumen/sources/xarray_sql.py
ghostiee-11: Fixed in f812387. Now just delegates to `super().get_schema(table, limit, shuffle=False)`. DataFusion doesn't support TABLESAMPLE, so forcing `shuffle=False` makes the parent use SQLLimit instead of SQLSample. No more duplication.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,375 @@
...
+                    d["max"] = float(vals.max())
+                    d["step"] = float(np.diff(vals[:2])[0]) if len(vals) > 1 else None
+                    d["type"] = "numeric"
+                else:
+                    d["values"] = vals.tolist()
+                    d["type"] = "categorical"
+                dim_info[dim] = d
+            info[tbl] = dim_info
+
+        return info if table is None else info.get(table, {})
```
ahuang11: Are these used anywhere?

ghostiee-11: Good catch, `get_dimension_info` was not called anywhere in the codebase. Removed it along with `__repr__` (no other source defines one).

Also simplified `_get_table_metadata` to match DuckDB's output structure (just `description`, `columns`, `rows`). The previous version had extra fields like `dimensions`, `shape`, and `dataset_attrs` that nothing in the framework consumes.

Other changes in this push:
- Removed unused numpy/pandas imports
- Added xarray dependencies to pixi.toml so CI can run the tests
- Added xarray section to the sources docs
- Fixed lint warnings in tests (NPY002, missing EOF newline)

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+
+        if self.tables is None:
+            self.tables = list(self._registered_tables)
+
+    @staticmethod
+    def _resolve_chunks(dataset, chunks):
+        """Resolve chunk specification for DataFusion registration."""
+        if isinstance(chunks, dict):
+            return chunks
+        ds_chunks = getattr(dataset, "chunks", None)
```
ahuang11: Doesn't dataset always have chunks?
```
import xarray as xr

xr.Dataset().chunks
```

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @property
+    def dataset(self):
+        """Access the underlying xarray Dataset."""
+        return self._dataset
+
+    def get_schema(self, table=None, limit=None, shuffle=False):
+        # DataFusion doesn't support TABLESAMPLE; force shuffle=False
+        # so the parent uses SQLLimit instead of SQLSample.
+        return super().get_schema(table, limit, shuffle=False)
```
ahuang11, line 215: I wonder if there's another syntax that it supports? TABLESAMPLE might be coming from postgres dialect?

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+        if isinstance(self.tables, dict):
+            all_known.update(self.tables.keys())
+        elif isinstance(self.tables, (list, tuple)):
+            all_known.update(self.tables)
+        if stripped in all_known:
+            return stripped
+        lower_map = {t.lower(): t for t in all_known}
+        if stripped.lower() in lower_map:
+            return lower_map[stripped.lower()]
+        return stripped
```
ahuang11: It's a big confusing what all these ifs are doing.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def execute(self, sql_query, params=None, *args, **kwargs):
+        """Execute a SQL query against the DataFusion context."""
+        result = self._ctx.sql(sql_query)
+        return result.to_pandas()
```
ahuang11, line 240: Is it just to pandas?

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @cached
+    def get(self, table, **query):
+        query.pop('__dask', None)
+        sql_expr = self.get_sql_expr(table)
+        sql_transforms = query.pop('sql_transforms', [])
+        conditions = [
+            (col, (val.start, val.stop) if isinstance(val, slice) else val)
+            for col, val in query.items()
+            if not col.startswith("__")
```
ahuang11, line 250: What usually starts with `__`?

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,316 @@
...
+                }
+            columns[table] = {
+                "data_type": str(var.dtype),
+                "description": attrs.get("long_name", ""),
+            }
+
+            metadata[table] = {
+                "description": description,
+                "columns": columns,
+                "rows": math.prod(var.shape),
```
ahuang11: Do we need to import math for just this?

lumen/sources/xarray_sql.py
ghostiee-11: Good question\! `xr.Dataset().chunks` returns `Frozen({})`, an empty mapping that's falsy in Python. So for non-dask datasets (opened without `chunks=`), the guard catches both `None` and empty-dict cases, falling back to `dict(dataset.sizes)`. I'll add a clarifying comment.

lumen/sources/xarray_sql.py
ghostiee-11: Yes, exactly. Our `dialect = "postgres"` causes `SQLSample` to emit `TABLESAMPLE`, which DataFusion doesn't support. The generic fallback uses `ORDER BY RAND()` which DataFusion also lacks. Forcing `shuffle=False` makes the parent use `SQLLimit` instead. I'll expand the comment to explain this chain.

lumen/sources/xarray_sql.py
ghostiee-11: Fair point, I'll simplify and add comments. The three steps are: (1) strip quotes (`"air"` to `air`), (2) exact match against known tables, (3) case-insensitive fallback (`AIR` to `air`). Can't use `normalize_table_name()` since that handles file paths, not xarray variable names. I'll refactor to be clearer.

lumen/sources/xarray_sql.py
ghostiee-11: Yes. DataFusion returns an Arrow-backed DataFrame. `.to_pandas()` converts it to match `BaseSQLSource.execute()`'s `pd.DataFrame` return type, same as DuckDB's `.fetch_df()`.

lumen/sources/xarray_sql.py
ghostiee-11: Good catch. The only `__`-prefixed key is `__dask`, already popped on the line above. The filter is redundant. I'll remove it and use `conditions = list(query.items())` to match DuckDB's pattern.

lumen/sources/xarray_sql.py
ghostiee-11: Nope, good call. xarray variables have `.size` which gives total elements directly. I'll replace `math.prod(var.shape)` with `var.size` and drop the `import math`.

lumen/sources/xarray_sql.py
ahuang11: Why are we choosing postgres as the dialect again?

lumen/sources/xarray_sql.py
ahuang11: I don't follow. I don't think you need to getattr if it's always available as a property.

ahuang11: Also out of curiosity, are you using some GitHub MCP with some LLM to respond to comments?

ghostiee-11: I use the gh CLI to read and respond to PR comments from my terminal. and a bit of LLM for framing it well!

ghostiee-11: > I don't follow. I don't think you need to getattr if it's always available as a property.

Okay!! i will simplify to use that

lumen/sources/xarray_sql.py
ghostiee-11: The AI utils layer calls `get_schema(shuffle=True)` by default (`lumen/ai/utils.py:445`), which triggers `SQLSample` to emit `TABLESAMPLE` via the postgres dialect. DataFusion doesn't support that, and its generic fallback (`ORDER BY RAND()`) also isn't available. This 3-line override just downgrades to `SQLLimit`, keeping everything else inherited from the parent.

ghostiee-11: Check this, 
DataFusion doesn't natively support TABLESAMPLE yet ([apache/datafusion#16533](https://github.com/apache/datafusion/issues/16533) is still open).

ahuang11: Yes; I'm wondering though why selecting dialect = 'postgres' vs any other dialect, or custom dialects?

pyproject.toml
```diff
@@ -75,6 +75,7 @@ ai-litellm = ['litellm']
 ai-ollama = ['ollama']
 chromadb = ['chromadb']
 bigquery = ['google-cloud-bigquery', 'sqlalchemy-bigquery']
+xarray = ['xarray', 'xarray-sql', 'netCDF4', 'h5netcdf', 'zarr', 'cfgrib']
```
ahuang11: I don't think we should include all these as dependencies. Maybe xarray, xarray-sql, zarr and netcdf4/h5netcdf. The others are more for grib.

ahuang11:
I think this PR is really close; left a couple more comments; will do some testing with it and see if it's missing anything.

pixi.toml
```diff
@@ -68,6 +68,17 @@ duckdb = "*"
 intake-sql = "*"
 sqlalchemy = "*"
 
+[feature.xarray.dependencies]
+xarray = "*"
```
ahuang11, line 72: do we need xarray-sql here?

pyproject.toml
ghostiee-11: Okay got it!! will fix this

lumen/sources/xarray_sql.py
ghostiee-11: Postgres is the closest match for the SQL features that DataFusion does support, and i read somewhere 
postgres-style SQL is a safe mental model when using DataFusion.

pixi.toml
ghostiee-11: Yes, `xarray-sql` provides the `XarrayContext`/DataFusion bridge. It's PyPI-only so it's under `[feature.xarray.pypi-dependencies]`. Will trim `cfgrib`/`eccodes` from here per your other feedback.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,344 @@
...
+                        label += f" [{ac_long}]"
+                    aux_labels.append(label)
+                description += f" (auxiliary coords: {', '.join(aux_labels)})"
+
+            metadata[table] = {
+                "description": description,
+                "columns": columns,
+                "rows": var.size,
+            }
+        return metadata
```
ahuang11, line 344: Lots of indentation; ideally we reduce indentation by early continues / returns or calling helper methods

pixi.toml
ahuang11: okay I missed it; thanks!

pixi.toml
```diff
@@ -68,6 +68,15 @@ duckdb = "*"
...
+netcdf4 = "*"
+h5netcdf = "*"
```
ahuang11: Not sure it's necessary to have both h5netdf and netcdf4. I suspect netcdf4 is sufficient.

pyproject.toml
```diff
@@ -75,6 +75,7 @@ ai-litellm = ['litellm']
 ai-ollama = ['ollama']
 chromadb = ['chromadb']
 bigquery = ['google-cloud-bigquery', 'sqlalchemy-bigquery']
+xarray = ['xarray', 'xarray-sql', 'netCDF4', 'h5netcdf', 'zarr']
```
ahuang11: You could also have a `xarray-grib` and include cfgrib and eccodes

lumen/sources/xarray_sql.py
ghostiee-11: Okay, I'll refactor _get_table_metadata to extract a _build_column_metadata helper and use early continue to reduce nesting. Will push the fix.

pixi.toml
ghostiee-11: netcdf4 covers the common case. I'll drop h5netcdf from both pixi.toml and pyproject.toml.

pyproject.toml
ghostiee-11: Thanks, I'll add a separate xarray-grib extra with cfgrib and eccodes, keeping the base xarray group lean.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,352 @@
...
+    XARRAY_AVAILABLE = False
+
+# Map file extensions to xarray engines
+XARRAY_ENGINES = {
+    ".nc": "netcdf4",
+    ".nc4": "netcdf4",
+    ".netcdf": "netcdf4",
+    ".h5": "netcdf4",
+    ".hdf5": "netcdf4",
+    ".he5": "netcdf4",
```
ahuang11: Maybe these shouldn't be listed; not sure xarray supports them well. h5py would be better for these I think.

ahuang11 (approved):
Just a minor comment about the h5 extensions; otherwise, I believe this will go in v1.2.0

lumen/sources/xarray_sql.py
ghostiee-11: Removed the H5 extensions from engine detection and dropped h5netcdf from deps. Added a separate xarray-grib extra for GRIB support.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,349 @@
...
+from pathlib import Path
+from typing import Any, ClassVar
+
+import param
+
+from ..transforms.sql import SQLFilter
+from .base import BaseSQLSource, cached
+
+try:
+    import xarray as xr
```
philippjfr: Either the sources/xarray_sql.py module should not be imported by default OR these imports need to be inlined. We should not be loading xarray by default just because it's installed.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,349 @@
...
+            open_kwargs=self.open_kwargs,
+            variables=self.variables,
+        )
+
+        self._ctx = XarrayContext()
+        chunk_spec = self._resolve_chunks(self._dataset, self.chunks)
+
+        for var_name in self._dataset.data_vars:
+            var_ds = self._dataset[[var_name]]
+            self._ctx.from_dataset(var_name, var_ds, chunks=chunk_spec)
```
philippjfr, line 144: Is this expensive? Should it be done lazily?

ahuang11: I think it's lazy until collect is called: https://github.com/alxmrs/xarray-sql/issues/93#issuecomment-4015792905

lumen/sources/xarray_sql.py
ghostiee-11: Good point. I'll remove

lumen/sources/xarray_sql.py
ghostiee-11: The assign_coords itself is cheap, it just attaches small np.arange() arrays to the dataset metadata without copying data. The DataFusion context creation downstream is also lazy per xarray-sql#93 as Andrew noted.

lumen/sources/xarray_sql.py
ahuang11: Ensure it's properly reordered; perhaps even as a methods of the class.

lumen/sources/xarray_sql.py
ahuang11: Where is assign coords being used?

ghostiee-11: Ohh sorry, I added this in  AI integration, got confused

lumen/sources/xarray_sql.py
```diff
@@ -82,6 +63,23 @@ class XArraySQLSource(BaseSQLSource):
...
+        ".grb": "cfgrib",
+        ".grb2": "cfgrib",
+    }
+
+    @classmethod
+    def _detect_engine(cls, path: str) -> str | None:
+        """Detect xarray engine from file extension."""
+        suffix = Path(path).suffix.lower()
+        return cls._xarray_engines.get(suffix)
+
```
ahuang11: Really awkward place to put a method. It should be after __init__

ahuang11 (changes requested):

ahuang11 (approved):

PR #1746, opened 2026-03-13
Fix Table view horizontal overflow with long string columns

ghostiee-11:
## Description

I was exploring an Amazon product reviews dataset using Lumen AI and noticed that when the Table view renders columns containing long text (like review bodies or metadata strings), the Tabulator scrolls endlessly to the right. The table becomes unusable because columns stretch to fit the full text content, pushing the layout way beyond the viewport.

I fixed this by setting sensible defaults on the `Table` view:

- `layout='fit_data_stretch'` - columns fit their content but stretch to fill available width
- `sizing_mode='stretch_width'` - table respects the container width
- `maxInitialWidth: 300` via Tabulator's `columnDefaults` config - caps the initial column width so long strings don't blow out the layout

Columns are still resizable by dragging, so users can widen any column to see full content. If a user passes their own `configuration` kwarg, it merges on top of the defaults so nothing is lost.

### Before

https://github.com/user-attachments/assets/bd0d5b6b-b52f-48c9-bb5d-bb0084c62bd6

### After

https://github.com/user-attachments/assets/99922015-11a3-42df-a744-766beb983582

## How Has This Been Tested?

- Tested with a 999-row Amazon reviews dataset containing review text up to 11K characters
- Verified on both Lumen YAML dashboards and Lumen AI
- Added 2 unit tests: one for default layout params, one for user configuration merging
- All 25 existing view tests pass

## Checklist

- [ ] Added documentation

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1163,11 +1163,24 @@ class Table(View):
     _panel_type = pn.widgets.tables.Tabulator
 
     def get_panel(self):
-        return self._panel_type(**self._normalize_params(self._get_params()))
+        params = self._normalize_params(self._get_params())
+        config = {'columnDefaults': {'maxInitialWidth': 300}}
+        if 'configuration' in self.kwargs:
+            config.update(self.kwargs['configuration'])
+        params['configuration'] = config
+        return self._panel_type(**params)
 
     def _get_params(self):
-        return dict(value=self.get_data(), disabled=True, page_size=self.page_size,
-                    **self.kwargs)
+        kwargs = {k: v for k, v in self.kwargs.items() if k != 'configuration'}
+        params = dict(
+            value=self.get_data(),
+            disabled=True,
+            page_size=self.page_size,
+            sizing_mode='stretch_width',
+            layout='fit_data_stretch',
+        )
+        params.update(kwargs)
+        return params
 
 
 class DownloadView(View):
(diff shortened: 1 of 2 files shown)


lumen/views/base.py
```diff
@@ -1163,11 +1163,24 @@ class Table(View):
...
+        config = {'columnDefaults': {'maxInitialWidth': 300}}
+        if 'configuration' in self.kwargs:
+            config.update(self.kwargs['configuration'])
+        params['configuration'] = config
+        return self._panel_type(**params)
 
     def _get_params(self):
-        return dict(value=self.get_data(), disabled=True, page_size=self.page_size,
-                    **self.kwargs)
+        kwargs = {k: v for k, v in self.kwargs.items() if k != 'configuration'}
```
ahuang11, line 1174: Curious why we're filtering out configuration here? If for good reason, can we add an inline comment?

If user passes a configuration override, shouldn't it take precedence?

ghostiee-11: We're filtering it out because `configuration` gets handled separately in `get_panel()` where we merge it with defaults like `maxInitialWidth: 300`. If we let it pass through `_get_params()` directly, two things break - we lose the ability to set defaults underneath user values, and `_update_panel()` on data refreshes would overwrite the merged config with just the raw user value. User config does take full precedence though, `.update()` makes sure of that.

ahuang11 (approved):

lumen/views/base.py
ahuang11: Okay thanks!

ghostiee-11: No, Issue!! Thanks for the merge..😊

Issue #1436, opened 2025-10-08
Handling two separate SQL and joining the results


ahuang11:
If I ask it to compare gpt-4o-mini token usage vs 4.1-mini usage, it might invoke SQLAgent twice, but then the first SQLAgents' results get discarded.

PR #1747, opened 2026-03-13
Fix retry output accumulation and NotRequired accumulation parsing

ghostiee-11:
## Description

Two independent bug fixes found while investigating #1436.

### Fix 1: Retry output accumulation (`coordinator/base.py`)

`_retry_from_provider()` initializes `outputs = []` but never accumulates the Panel objects (charts, tables) returned by retried tasks. When a task fails with `MissingContextError` and the retry succeeds, the retried views are silently lost from the UI even though the context is correct.

Added `outputs += out` in both the pre-provider loop and the provider task execution.

### Fix 2: NotRequired accumulation parsing (`context.py`)

`_parse_accumulate_meta()` checks `get_origin(annotation) is Annotated`, but when a field is declared as `NotRequired[Annotated[list[T], ("accumulate", "key")]]`, the origin is `NotRequired`, not `Annotated`. The accumulation metadata is never extracted and the field silently falls back to LWW.

Added unwrapping of `NotRequired`/`Required` before the `Annotated` check.

## How Has This Been Tested?

Added two unit tests in `test_context.py`:
- `test_parse_accumulate_meta_notrequired_wrapped` - verifies NotRequired[Annotated[...]] is parsed correctly
- `test_merge_contexts_notrequired_annotated_accumulates` - verifies end-to-end accumulation works with NotRequired wrapping

All 21 context tests pass.

## Checklist

- [ ] Added documentation

### DIFF
--- a/lumen/ai/coordinator/planner.py
+++ b/lumen/ai/coordinator/planner.py
@@ -396,11 +396,7 @@ class Planner(Coordinator):
 
             if len(consecutive_steps) > 1:
                 # Join the consecutive steps into one
-                additional = ". ".join(step.instruction for step in consecutive_steps[1:])
-                if current_step.actor == "SQLAgent":
-                    combined_instruction = f"{current_step.instruction} Combine the following into a single query using JOIN or UNION: {additional}"
-                else:
-                    combined_instruction = f"{current_step.instruction} Additionally: {additional}"
+                combined_instruction = f"{current_step.instruction} Additionally: " + ". ".join(step.instruction for step in consecutive_steps[1:])
                 combined_title = f"{current_step.title} (combined {len(consecutive_steps)} steps)"
 
                 # Create a new step with combined instruction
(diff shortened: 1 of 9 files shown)


lumen/ai/coordinator/planner.py
```diff
@@ -396,7 +396,11 @@ async def _resolve_plan(
 
             if len(consecutive_steps) > 1:
                 # Join the consecutive steps into one
-                combined_instruction = f"{current_step.instruction} Additionally: " + ". ".join(step.instruction for step in consecutive_steps[1:])
+                additional = ". ".join(step.instruction for step in consecutive_steps[1:])
+                if current_step.actor == "SQLAgent":
+                    combined_instruction = f"{current_step.instruction} Combine the following into a single query using JOIN or UNION: {additional}"
+                else:
+                    combined_instruction = f"{current_step.instruction} Additionally: {additional}"
                 combined_title = f"{current_step.title} (combined {len(consecutive_steps)} steps)"
```
ahuang11: I don't think this is doing what you think it's doing.

ghostiee-11:
Stripped the PR down to just the two bug fixes + added tests for both. The diff is now 3 files, 25 lines.

philippjfr (approved):
Appreciate the fixes, they look good.

PR #1750, opened 2026-03-14
Fix chat file upload failing when remote data source is loaded

### LINKED ISSUE #1749: Chat file upload fails when remote data source is loaded
## Description

When lumen-ai is started with a remote data source (e.g., `penguins.csv` from URL) and a user uploads files via the chat "Attach files" button, querying the uploaded data fails on the second upload.

## Steps to Reproduce

1. Start lumen-ai with a remote data source:
   ```bash
   lumen-ai penguins.csv  # or any remote/local source
   ```
2. Upload a CSV file (e.g., `employee_data.csv`) via the chat attach button
3. Query the uploaded data (works fine for the first upload)
4. Upload a second CSV file (e.g., `test_sales.csv`)
5. Query the second uploaded file

**Expected:** Both uploads are queryable.
**Actual:** The second upload fails with "Unable to complete your request".

## Root Cause

Two issues in the file upload pipeline:

### 1. DuckDB source name collision (`_count` never incremented)

In `BaseSourceControls._process_files()`, each file upload creates a new `DuckDBSource(uri=":memory:")` with name `UploadedSource000000`. The `_count` param is never incremented, so the second upload gets the same name, causing a collision. Additionally, since each `DuckDBSource(uri=":memory:")` creates a separate in-memory connection, tables from the first upload are invisible to the second source.

### 2. MetadataLookup skips sources it has already seen (set-based tracking)

`MetadataLookup._update_vector_store()` tracks processed sources as a `set` of source names. When a source gets new tables added (e.g., second file uploaded to the same DuckDB source), the set-based check sees the source name already exists and skips re-indexing. The new tables never make it into the vector store.

## Before (bug)

On the second file upload, querying the uploaded data fails:
Image

Image

## After (fix)

After the fix, both uploads are queryable from the same DuckDB connection:

Image

## Proposed Fix

1. **Reuse the existing ephemeral DuckDB source** across uploads instead of creating a new one each time, keeping all uploaded tables in one connection
2. **Track table counts** instead of just source names in `_sources_in_progress`, so new tables trigger re-indexing
3. **Replace polling loop** with `asyncio.Event` for readiness synchronization
4. **Use `param.watch`** to keep `_ready_event` in sync with `_ready` param (eliminates 6 manual sync points)

## Environment

- lumen main branch 
- Python 3.14
- DuckDB in-memory mode

ghostiee-11:
## Description

Fixes #1749

When lumen-ai has a remote data source loaded and users upload files via chat, the second upload fails with "Unable to complete your request". Two issues cause this:

1. Each upload creates a new `DuckDBSource(uri=":memory:")`, but DuckDB memory connections are isolated, so tables from earlier uploads are invisible. Also `_count` never increments, causing name collisions.

2. `MetadataLookup._update_vector_store()` tracks sources as a `set` of names. When new tables get added to an existing source, the set check skips re-indexing because the name already exists.

I fixed (1) by reusing the existing ephemeral DuckDB source across uploads, and (2) by tracking table counts instead of just names so new tables trigger re-indexing.

### After

Both uploads queryable from the same DuckDB connection:
Screenshot 2026-03-16 at 11 30 21 PM

after

## How Has This Been Tested?

Started lumen-ai with `penguins.csv`, uploaded two files via chat sequentially, confirmed both are queryable. Added 13 new tests covering source reuse, re-indexing, and edge cases (20 total passing).

### DIFF
--- a/lumen/ai/tools/metadata_lookup.py
+++ b/lumen/ai/tools/metadata_lookup.py
@@ -127,6 +127,16 @@ class MetadataLookup(VectorLookupTool):
         super().__init__(**params)
         self._raw_metadata = {}
         self._semaphore = asyncio.Semaphore(self.max_concurrent)
+        self._ready_event = asyncio.Event()
+        self._ready_event.set()  # Initially ready (no pending sync)
+        self.param.watch(self._sync_ready_event, "_ready")
+
+    def _sync_ready_event(self, event):
+        """Keep _ready_event in sync with _ready param automatically."""
+        if event.new is False:
+            self._ready_event.clear()
+        else:
+            self._ready_event.set()
 
     async def sync(self, context: TContext):
         """
@@ -241,19 +251,21 @@ class MetadataLookup(VectorLookupTool):
 
         async with self._progress_lock:
             if vector_store_id not in self._sources_in_progress:
-                self._sources_in_progress[vector_store_id] = set()
+                self._sources_in_progress[vector_store_id] = {}
 
         log_debug(f"[MetadataLookup] Starting _update_vector_store for {len(sources)} sources")
 
         for source in sources:
             async with self._progress_lock:
-                # Skip this source if it's already being processed for this vector store
-                if source.name in self._sources_in_progress[vector_store_id]:
-                    log_debug(f"[MetadataLookup] Skipping source {source.name} - already in progress")
+                # Skip this source if already processed with the same (or more) tables
+                current_table_count = len(source.get_tables())
+                known_count = self._sources_in_progress[vector_store_id].get(source.name, 0)
+                if known_count >= current_table_count:
+                    log_debug(f"[MetadataLookup] Skipping source {source.name} - already processed {known_count} tables")
                     continue
 
-                # Mark this source as in progress for this vector store
-                self._sources_in_progress[vector_store_id].add(source.name)
+                # Mark this source as in progress with current table count
+                self._sources_in_progress[vector_store_id][source.name] = current_table_count
                 processed_sources.append(source.name)
                 log_debug(f"[MetadataLookup] Processing source {source.name}")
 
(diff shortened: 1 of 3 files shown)


lumen/tests/ai/test_coordinator.py
```diff
@@ -277,3 +277,334 @@ async def test_plan_revise(sql_plan):
...
+
+
+async def test_process_files_duplicate_table_name():
+    """
+    Uploading a file with the same table name twice should overwrite
+    the table data, not error out.
+    """
+    import io
+
+    from lumen.ai.controls.base import BaseSourceControls, UploadedFileRow
```
ahuang11: Imports can be up top

ghostiee-11: Sure.. fixing it

lumen/ai/tools/metadata_lookup.py
```diff
@@ -127,6 +127,16 @@ def __init__(self, **params):
...
+        self._ready_event = asyncio.Event()
+        self._ready_event.set()  # Initially ready (no pending sync)
+        self.param.watch(self._sync_ready_event, "_ready")
+
+    def _sync_ready_event(self, event):
+        """Keep _ready_event in sync with _ready param automatically."""
+        if event.new is False:
+            self._ready_event.clear()
+        else:
+            self._ready_event.set()
```
ahuang11, line 139: I don't understand the goal of `_ready_event`

ghostiee-11: When a file is uploaded, `_sync_metadata()` runs as a background task via `asyncio.create_task()` in controls. If the user sends a query before that sync finishes, `respond()` would search stale metadata. The `_ready_event` makes `respond()` wait (with a 30s timeout fallback) for the sync to complete first. I noticed the existing `_ready` param is set in several places but never actually checked before `respond()` runs, so the asyncio.Event fills that gap.

philippjfr: Makes sense, I had some changes in my stash to do something similar.

philippjfr (approved):

lumen/ai/tools/metadata_lookup.py
ghostiee-11: That's great, Thanks for the review and Merge!!

PR #1756, opened 2026-03-15
Add import guards for optional dependency source modules

ghostiee-11:
## Description

I was building a Panel dashboard that pulls from different Lumen sources and noticed that importing source modules like `BigQuerySource` or `SnowflakeSource` without their optional packages installed crashes immediately at module level with a raw `ModuleNotFoundError`. There's no hint about what to install or how to fix it.

For instance, just doing:
```python
from lumen.sources.bigquery import BigQuerySource
```
gives:
```
ModuleNotFoundError: No module named 'google.cloud'
```

I think it would be nicer if the import itself works fine, and the error only shows up when you actually try to instantiate the source, with a message like:
```
ImportError: BigQuerySource requires the 'google-cloud-bigquery' package. Install it with: pip install lumen[bigquery]
```

This PR wraps the optional imports in `try/except` for 4 source modules (`sqlalchemy.py`, `bigquery.py`, `snowflake.py`, `ae5.py`) and adds a check in each `__init__`. I followed the same pattern already used in `lumen/sources/base.py` for `aiohttp`.

I also noticed that `snowflake` and `ae5` didn't have corresponding extras in `pyproject.toml`, so I added those too. And `bigquery.py` was missing `from __future__ import annotations` which caused a `TypeError` on `Client | None` when the fallback `None` values kick in.

## How Has This Been Tested?

- Confirmed all 4 modules import without crashing when deps are missing
- Confirmed instantiation raises helpful `ImportError` with correct pip command
- Confirmed `SQLAlchemySource` still works normally with sqlalchemy installed
- Added 12 parametrized tests covering import success, error messages, and guard presence

```
$ python -m pytest lumen/tests/sources/test_optional_imports.py -v
12 passed in 0.31s
```

## Checklist

- [x] Tests added and passing

### DIFF
--- a/lumen/sources/ae5.py
+++ b/lumen/sources/ae5.py
@@ -9,7 +9,11 @@ import pandas as pd
 import param  # type: ignore
 import requests
 
-from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
+try:
+    from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
+except ImportError:
+    AEAdminSession = None
+    AEUserSession = None
 from panel import state
 
 from ..util import get_dataframe_schema
@@ -86,6 +90,11 @@ class AE5Source(Source):
     }
 
     def __init__(self, **params):
+        if AEAdminSession is None:
+            raise ImportError(
+                "AE5Source requires the 'ae5-tools' package. "
+                "Install it with: pip install lumen[ae5]"
+            )
         super().__init__(**params)
         self._session = AEUserSession(
             self.hostname, self.username, self.password, persist=False,
--- a/lumen/sources/ae5.py
+++ b/lumen/sources/ae5.py
@@ -11,11 +11,11 @@ import requests
 
 try:
     from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
-except ImportError:
+except ImportError as e:
     raise ImportError(
         "AE5Source requires the 'ae5-tools' package. "
         "Install it with: pip install lumen[ae5]"
-    ) from None
+    ) from e
 
 from panel import state
 
--- a/lumen/sources/ae5.py
+++ b/lumen/sources/ae5.py
@@ -12,8 +12,11 @@ import requests
 try:
     from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
 except ImportError:
-    AEAdminSession = None
-    AEUserSession = None
+    raise ImportError(
+        "AE5Source requires the 'ae5-tools' package. "
+        "Install it with: pip install lumen[ae5]"
+    )
+
 from panel import state
 
 from ..util import get_dataframe_schema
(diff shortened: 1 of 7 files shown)


lumen/sources/bigquery.py
```diff
@@ -1,19 +1,31 @@
...
+    from google.cloud.bigquery.client import Client
+    from tqdm import tqdm
+except ImportError:
+    google = None
+    Credentials = None
+    DefaultCredentialsError = None
+    bigquery = None
+    exceptions = None
+    Client = None
+    tqdm = None
```
ahuang11: tqdm isn't part of google I think. I also don't think we need to set all of these to None?

lumen/sources/snowflake.py
```diff
@@ -11,13 +11,20 @@
...
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives.serialization import (
+        Encoding, NoEncryption, PrivateFormat, load_pem_private_key,
+    )
+    from snowflake.connector.constants import QueryStatus
+except ImportError:
+    snowflake = None
+    default_backend = None
+    Encoding = NoEncryption = PrivateFormat = load_pem_private_key = None
+    QueryStatus = None
```
ahuang11: Same as above.

ahuang11 (changes requested):

lumen/sources/bigquery.py
ghostiee-11: Good catch, `tqdm` is a separate package. I'll pull it out of the google import block.

lumen/sources/snowflake.py
ghostiee-11: Will fix this the same way.

lumen/sources/ae5.py
```diff
@@ -9,7 +9,14 @@
...
 import requests
 
-from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
+try:
+    from ae5_tools.api import AEAdminSession, AEUserSession  # type: ignore
+except ImportError:
+    raise ImportError(
+        "AE5Source requires the 'ae5-tools' package. "
+        "Install it with: pip install lumen[ae5]"
+    ) from None
```
ahuang11: Why `from None` instead of `except ImportError as e` and `from e`?

lumen/sources/snowflake.py
```diff
@@ -11,13 +11,20 @@
...
+    import snowflake.connector
+
+    from cryptography.hazmat.backends import default_backend
+    from cryptography.hazmat.primitives.serialization import (
+        Encoding, NoEncryption, PrivateFormat, load_pem_private_key,
+    )
+    from snowflake.connector.constants import QueryStatus
+except ImportError:
+    raise ImportError(
+        "SnowflakeSource requires the 'snowflake-connector-python' package. "
```
ahuang11: Might want to separate cryptography or mention it

lumen/sources/ae5.py
ghostiee-11: from None keeps the traceback clean for end users since they just need the install hint, not the original error chain. Will switch to from e if you'd prefer keeping the full context though

lumen/sources/snowflake.py
ghostiee-11: True, I'll separate them so each points to the right package

ahuang11 (approved):

PR #1758, opened 2026-03-15
fix: Re-selecting the same file after chip removal in ChatAreaInput

### LINKED ISSUE #1757: Re-selecting the same file after chip removal does nothing
## How I Discovered It

While using the **Chat with Data** feature, I noticed that after attaching a file via the **attach button (+)** and then removing it by clicking the chip's **×**, trying to attach the **same file again** via the attach button does nothing — the chip never reappears.

**Steps to reproduce:**
1. Open the Chat with Data interface
2. Click **+** → select a file (e.g. `penguins_raw.csv`) → chip appears ✓
3. Click **×** on the chip to remove it
4. Click **+** → select the **same file** again → nothing happens ✗

**Screen recording:**

https://github.com/user-attachments/assets/212397bb-f43d-407d-b894-ef831158be35

**What works (workarounds):**
- Drag and drop the same file → works ✓
- Selecting a **different** file via the attach button → works ✓

SuMayaBee:
## Description

After removing a file chip (×) in the Chat with Data input, the hidden `<input type="file">` element still holds the previous filename internally. Browsers do not fire `onChange` when the same file is selected again because the value hasn't changed, causing the chip to never reappear. Drag-and-drop bypasses the file input entirely, which is why it works.

**Before:**

https://github.com/user-attachments/assets/8da2142c-9908-4f71-9290-edaa5e78929e

**After:**

https://github.com/user-attachments/assets/7b577c80-8bad-456f-8f3b-f3fa57448b6c

**How it's fixed:** 
When a chip is removed or a file is selected, we now reset the hidden `<input type="file">` value to `""` immediately after. This is done by patching the `panel_material_ui` bundle on import in `lumen/ai/__init__.py`:

```python
# Reset hidden file input when a chip is removed (× button)
content = content.replace(
    "ee.splice(W,1),v(ee),_.current=ee}",
    'ee.splice(W,1),v(ee),_.current=ee,R.current&&(R.current.value="")}',
)
# Reset hidden file input after files are selected via the picker
content = content.replace(
    "v(ue),_.current=ue}}",
    'v(ue),_.current=ue,W.target.value=""}}',
)
```

This tricks the browser into treating the next selection as a fresh one, so `onChange` fires correctly even when the same file is picked again.

Fixes #1757

## How Has This Been Tested?

1. Start the lumen AI server
2. Open Chat with Data
3. Click **+** and select any file (e.g. `penguins_raw.csv`)
4. Click **×** on the chip to remove it
5. Click **+** and select the **same file** again
6. Verify the chip reappears and the file is staged for upload

Confirmed that drag-and-drop and selecting a different file still work as expected.

## AI Disclosure

Tools and Models: Claude Code + Sonnet 4.6

## Checklist

- [ ] Added documentation

### DIFF
--- a/lumen/ai/__init__.py
+++ b/lumen/ai/__init__.py
@@ -14,3 +14,28 @@ pn.chat.message.DEFAULT_AVATARS.update({
     "sql": "🗄️",
     "router": "🚦",
 })
+
+
+def _patch_chat_area_input_file_reset():
+    try:
+        from panel_material_ui.base import BASE_PATH
+        bundle_path = BASE_PATH / "dist" / "panel-material-ui.bundle.js"
+        if not bundle_path.exists():
+            return
+        content = bundle_path.read_text(encoding="utf-8")
+        if 'R.current&&(R.current.value="")' in content:
+            return
+        content = content.replace(
+            "ee.splice(W,1),v(ee),_.current=ee}",
+            'ee.splice(W,1),v(ee),_.current=ee,R.current&&(R.current.value="")}',
+        )
+        content = content.replace(
+            "v(ue),_.current=ue}}",
+            'v(ue),_.current=ue,W.target.value=""}}',
+        )
+        bundle_path.write_text(content, encoding="utf-8")
+    except Exception:
+        pass
+
+
+_patch_chat_area_input_file_reset()


lumen/ai/__init__.py
```diff
@@ -14,3 +14,28 @@
     "sql": "🗄️",
     "router": "🚦",
 })
+
+
+def _patch_chat_area_input_file_reset():
+    try:
+        from panel_material_ui.base import BASE_PATH
```
ahuang11, line 21: Imports at the top please. Separately, I think this change should be submitted upstream to https://github.com/panel-extensions/panel-material-ui

PR #1761, opened 2026-03-16
Add edit_callback to allow direct editing of chat messages

ghostiee-11:
## Description

Addresses #1340

I added an `edit_callback` to the ChatInterface so users can edit their previously sent messages. When a user clicks the edit icon on a message, modifies the text, and submits:

1. The conversation is truncated from the edited message onward
2. All removed messages are marked as `removed=True` in the logs (if logging is enabled)
3. The edited content is re-sent as a new message, triggering a fresh AI response

Editing is blocked while a response is being generated. If a user tries to edit mid-response, a notification tells them to wait.

### How it works

Panel already has `ChatFeed.edit_callback` infrastructure that auto-shows edit icons on user messages when the callback is set. I wired a new `on_edit` async callback into `_configure_interface()` following the same closure pattern used by `on_undo`, `on_rerun`, and `on_clear`.

### Cross-repo dependency

The edit UI in `panel-material-ui` had a bug where clicking edit threw an `AttributeError` because `_handle_msg` passed a raw string to Panel's `_toggle_edit` (which expects a param Event), and Panel's `_toggle_edit` swaps `_placeholder.object` which the React frontend never observes (it renders `_object_panel` directly via `model.get_child()`). I submitted a companion fix: panel-extensions/panel-material-ui#600

**Both PRs need to be merged for the full end-to-end edit flow to work:**
- **This PR** (Lumen): Adds the callback logic to truncate and re-send
- **Companion PR** (panel-material-ui#600): Fixes the edit UI so the edit area actually appears

### Demo

https://github.com/user-attachments/assets/07960ecb-eea2-4617-bf4c-2e7baef13c42

## How Has This Been Tested?

- 7 unit tests added covering all edit callback paths (truncate+resend, last message, middle message, busy block, log updates, no-logs safety, callback wiring)
- All 63 tests in `test_ui.py` pass
- Manually tested end-to-end: sent a query, clicked edit, modified text, confirmed conversation truncated and new response generated

## Checklist

- [x] Tests added and passing
- [ ] Added documentation

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -823,6 +823,36 @@ class UI(Viewer):
                 for message in instance.objects[message_index:]:
                     self._logs.update_status(message_id=id(message), removed=True)
 
+            # If on a non-Home exploration, remove it so re-send creates
+            # a fresh exploration instead of a duplicate tab
+            current_item = self._exploration
+            current_exploration = current_item['view']
+            if current_exploration.plan is not None:
+                with hold():
+                    for child in current_item.get('items', []):
+                        if child['view'].plan is not None:
+                            child['view'].plan.cleanup()
+                        child['view'].context.clear()
+                    current_exploration.plan.cleanup()
+                    current_exploration.context.clear()
+                    if current_item in self._explorations.items:
+                        self._explorations.items = [
+                            it for it in self._explorations.items
+                            if it is not current_item
+                        ]
+                    else:
+                        parent = current_item["parent"]
+                        self._explorations.update_item(
+                            parent,
+                            items=[
+                                it for it in parent["items"]
+                                if it is not current_item
+                            ],
+                        )
+                    home_item = self._explorations.items[0]
+                    self._explorations.value = self._exploration = home_item
+                    self._last_synced = self._home
+
             # Truncate conversation and re-send edited content as a new message
             with instance.param.update(disabled=True, loading=True), hold():
                 instance.objects = instance.objects[:message_index]
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -2032,7 +2032,8 @@ class ExplorerUI(UI):
         self._update_main_view(force_report_mode=active)
 
     def _cleanup_exploration(self, item):
-        """Clean up an exploration and its children, remove from tree, switch to Home."""
+        """Clean up an exploration and its children, remove from tree,
+        and switch to parent (or Home for top-level explorations)."""
         exploration = item['view']
         if exploration.plan is None:
             return
@@ -2043,18 +2044,22 @@ class ExplorerUI(UI):
         exploration.plan.cleanup()
         exploration.context.clear()
         if item in self._explorations.items:
+            # Top-level exploration: remove and switch to Home
             self._explorations.items = [
                 it for it in self._explorations.items if it is not item
             ]
+            home_item = self._explorations.items[0]
+            self._explorations.value = self._exploration = home_item
+            self._last_synced = self._home
         else:
+            # Nested (followup) exploration: remove and switch to parent
             parent = item["parent"]
             self._explorations.update_item(
                 parent,
                 items=[it for it in parent["items"] if it is not item],
             )
-        home_item = self._explorations.items[0]
-        self._explorations.value = self._exploration = home_item
-        self._last_synced = self._home
+            self._explorations.value = self._exploration = parent
+            self._last_synced = parent['view']
 
     async def _delete_exploration(self, item):
         await self._idle.wait()
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -821,12 +821,12 @@ class UI(Viewer):
                 for message in instance.objects[message_index:]:
                     self._logs.update_status(message_id=id(message), removed=True)
 
-            self._cleanup_exploration(self._exploration)
-
-            with instance.param.update(disabled=True, loading=True), hold():
-                instance.objects = instance.objects[:message_index]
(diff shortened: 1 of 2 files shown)


ghostiee-11:
Pushed the fix + tests.

**What changed:**

The `on_edit` callback now cleans up the current exploration before re-sending. When editing on a non-Home exploration:

1. Child (followup) exploration plans are cleaned up and contexts cleared
2. Current exploration's `plan.cleanup()` is called (unwatches task watchers)
3. Exploration context is cleared (removes stale SQL results)
4. Exploration is removed from `_explorations.items` (same pattern as `_delete_exploration`)
5. Switches back to Home, then truncates + re-sends

When editing on Home (no plan yet), existing behavior is preserved - just truncate and re-send.

**Tests added (4 new, 11 total edit tests, 67/67 pass):**

- `test_edit_on_exploration_removes_old_and_switches_to_home` - verifies cleanup + Home switch
- `test_edit_on_home_preserves_behavior` - no-op when no exploration exists
- `test_edit_on_exploration_does_not_duplicate_tabs` - old exploration not in tree after edit
- `test_edit_on_parent_exploration_removes_children` - child followups cleaned up with parent

**Manual reproduction (SQLAgent + VegaLiteAgent):**

Uploaded `sales_data.csv`, queried "Show me total revenue by category as a bar chart" - SQLAgent created the query, VegaLiteAgent rendered the bar chart. Then edited the message to "Show me average price by region as a pie chart" - old exploration tab was removed, new one created with fresh SQL and pie chart. No duplicate tabs.

https://github.com/user-attachments/assets/0a45274a-d0aa-4a7d-ac53-c88d5c4be586

lumen/ai/ui.py
```diff
@@ -807,6 +807,58 @@ def on_rerun(instance, _):
...
+                        self._explorations.update_item(
+                            parent,
+                            items=[
+                                it for it in parent["items"]
+                                if it is not current_item
+                            ],
+                        )
+                    home_item = self._explorations.items[0]
+                    self._explorations.value = self._exploration = home_item
+                    self._last_synced = self._home
```
ahuang11: Not a big fan of this being so indented / nested. It's hard to read/understand

lumen/tests/ai/test_ui.py
```diff
@@ -1057,3 +1057,424 @@ def test_resolve_data_pipeline_and_files(self):
...
+    assert ui._exploration['view'] is ui._home
+
+    # Exploration 2 context should be cleared
+    assert len(exploration_2_context) == 0
+
+    # Exploration 1 must still exist and be untouched
+    assert ui._explorations.items[1] is exploration_1_item
+    assert exploration_1_item['view'] is exploration_1
+    assert exploration_1.plan is not None
+    assert len(exploration_1.context) > 0  # sql_1 context preserved
```
ahuang11: I feel like this should be more explicit, e.g. the context keys and values.

ahuang11 (changes requested):
Thanks; I've been wanting this! Few comments and questions.

ghostiee-11:
When you edit on the parent (e.g. "Group by Season"), all its children (e.g. "Subset Season") get cleaned up too  `plan.cleanup()` + `context.clear()` on each child, then the parent itself. Both are removed from the tree and the UI switches back to Home. The edited message then goes through the normal `_chat_invoke` -> `_coordinator.respond` path to create a fresh exploration.

`test_edit_on_parent_exploration_removes_children` covers this: creates a parent, adds a child followup under it, edits on the parent, asserts both are gone and child context is cleared.

lumen/ai/ui.py
ghostiee-11: Agreed, will extract the exploration cleanup into a helper method on `ExplorerUI` (something like `_cleanup_exploration(item)`) and flatten the conditionals with early returns. That'll keep `on_edit` short and readable.

lumen/tests/ai/test_ui.py
ghostiee-11: Good point, will replace `len(exploration_1.context) > 0` with explicit key/value checks -- e.g. asserting the `sql_query` key and specific table slug are still present in exploration 1's context.

lumen/ai/ui.py
```diff
@@ -2062,6 +2031,32 @@ def _toggle_report_mode(self, active: bool):
         """Toggle between regular and report mode."""
         self._update_main_view(force_report_mode=active)
 
+    def _cleanup_exploration(self, item):
+        """Clean up an exploration and its children, remove from tree, switch to Home."""
+        exploration = item['view']
+        if exploration.plan is None:
+            return
+        with hold():
```
ahuang11: Maybe move hold outside _cleanup_exploration so less indented.

ghostiee-11: Yupp.. true fixed it

lumen/ai/ui.py
```diff
@@ -2008,6 +2031,31 @@ def _toggle_report_mode(self, active: bool):
...
+                it for it in self._explorations.items if it is not item
+            ]
+        else:
+            parent = item["parent"]
+            self._explorations.update_item(
+                parent,
+                items=[it for it in parent["items"] if it is not item],
+            )
+        home_item = self._explorations.items[0]
+        self._explorations.value = self._exploration = home_item
```
ahuang11: Do we need to switch to home, especially if it's nested?

ghostiee-11: Updated - nested explorations now switch to the parent instead of Home. Top-level ones still go to Home. Added a test for the nested case.

ahuang11 (approved):

PR #1764, opened 2026-03-17
Add follow-up suggestion chips after successful queries

### LINKED ISSUE #1763: Add dataset-specific follow-up suggestion chips after successful queries
## Problem

After every successful query in Lumen AI, users stare at results wondering "what next?" They have to think of and type every follow-up manually. This creates friction especially for users unfamiliar with their dataset's columns and structure.

## Proposed Solution

After each successful query, show 2-3 clickable suggestion chips that reference **actual column names and values** from the result. For example, after "Show total revenue by product" with columns `product`, `revenue`, `region`, `date`:

- (compare_arrows) "Compare total revenue between Widget A and Widget C"
- (filter_list) "Filter to products with total revenue greater than 30000"
- (bar_chart) "Show a bar chart of total revenue by product"

Clicking a chip sends it as a user query. Suggestions are generated via a single LLM call using a Jinja2 prompt template that includes the data summary, SQL executed, and conversation history (to avoid repeating past questions).

## Demo

https://github.com/user-attachments/assets/f191f361-2c39-43e7-8e91-773ac6160539

## Implementation Notes

- Hooks into `_postprocess_exploration()` after `_add_analysis_suggestions()`
- Uses `asyncio.create_task()` (fire-and-forget) so suggestions are non-blocking
- Reuses existing `_add_suggestions_to_footer()` with a new `group_name` parameter to coexist with analysis suggestions
- Adds a `follow_up_suggestions` param (default `True`) to toggle the feature off
- Stale message guard prevents appending to wrong message if user sends a new query while suggestions are generating
- Silent failure on LLM errors — suggestions are optional, never break the flow

ghostiee-11:
While exploring data in Lumen AI, I noticed that after each query I'd stare at the results trying to figure out what to ask next. Users unfamiliar with their dataset's columns have it worse - they don't know what's possible.

This adds 2-3 suggestion chips after each successful query that reference actual column names and values from the result. Clicking one sends it as a new query. A "Follow-Up Suggestions" toggle in Settings lets users turn it off.

## Demo

https://github.com/user-attachments/assets/c5eeaedc-1739-4d68-9f58-ee85e78bb5ef

## How to reproduce

1. `ExplorerUI().servable()`, upload a CSV
2. Ask "Show total revenue by product"
3. After the result, suggestion chips appear like "Compare revenue across region" or "Filter to top products by revenue"
4. Click one - it executes as a follow-up query
5. Toggle off in Settings > Follow-Up Suggestions

Closes #1763

## AI Disclosure

I identified the UX gap while testing Lumen AI , designed the approach (hooking into `_postprocess_exploration`, reusing `_add_suggestions_to_footer` with a `group_name` param), and verified through 9 new tests + manual end-to-end testing. AI helped scaffold the implementation.


lumen/ai/ui.py
```diff
@@ -1553,6 +1561,65 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+                data_summary=data_summary,
+                sql=sql,
+                output_type=output_type,
+                conversation_history=conversation_history,
+            )
+            result = await self.llm.invoke(
+                messages=[{"role": "user", "content": "Generate follow-up suggestions."}],
+                system=prompt,
+                response_model=FollowUpSuggestions,
+            )
```
ahuang11: I think we should use self._invoke_prompt instead, or even self._stream_prompt, so each suggestion appears as it's generated.

lumen/ai/models.py
```diff
@@ -29,6 +29,28 @@ def model_post_init(self, __context):
             raise MissingContextError(self.insufficient_context_reason)
 
 
+class FollowUpSuggestion(BaseModel):
+    icon: str = Field(
+        description="Material UI icon name for the button",
+        examples=["bar_chart", "pie_chart", "timeline", "map", "filter_list", "compare_arrows"]
```
ahuang11: Not sure we need this many examples

lumen/ai/models.py
```diff
@@ -29,6 +29,28 @@ def model_post_init(self, __context):
...
+    query: str = Field(
+        description="The exact natural language query to send to Lumen AI",
+        max_length=100
+    )
+
+
+class FollowUpSuggestions(BaseModel):
+    chain_of_thought: str = Field(
+        description="Brief reasoning about what follow-ups would be most useful given the data and current query"
+    )
```
ahuang11: Not sure we need chain of thought for suggestions especially for speed

lumen/ai/ui.py
ghostiee-11: Will refactor to use `_invoke_prompt` for consistency. Since `UI` doesn't inherit from `Actor` (where `_invoke_prompt` lives), I'll route it through `self._coordinator._invoke_prompt`. For streaming, structured output with a Pydantic model needs the full response to parse, so I'll start with `_invoke_prompt` and can explore streaming as a follow-up if needed.

lumen/ai/models.py
ghostiee-11: Yeah!! sure will strip to 2 or 3

lumen/ai/models.py
ghostiee-11: Fair point - the extra field adds latency without much benefit here since the suggestions are straightforward. I'll remove chain of thought from the model.

ahuang11: One other thought is using a smaller model for this as well, maybe a new follow_up key, or use the existing ui key.

ghostiee-11: Yeah, makes sense.. lemme give this a try and then push again

lumen/ai/ui.py
```diff
@@ -1553,6 +1562,63 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+
+        suggestions = [(s.icon, s.query) for s in result.suggestions]
+        self._add_suggestions_to_footer(
+            suggestions,
+            append_demo=False,
+            analysis=False,
+            hide_after_use=True,
+            num_objects=num_objects,
+            group_name="FollowUp",
+        )
```
ahuang11: Can the logic live in coordinator, i.e. Coordinator has a method to suggest_followup based on last message, then UI simply renders the output?

lumen/ai/coordinator/base.py
```diff
@@ -284,6 +284,10 @@ class Coordinator(Viewer, VectorLookupToolUser):
                 "template": PROMPTS_DIR / "Coordinator" / "tool_relevance.jinja2",
                 "response_model": ThinkingYesNo,
             },
+            "follow_up_suggestions": {
+                "template": PROMPTS_DIR / "ExplorerUI" / "follow_up_suggestions.jinja2",
```
ahuang11: Should live in Coordinator imo

lumen/ai/prompts/ExplorerUI/follow_up_suggestions.jinja2
```diff
@@ -0,0 +1,47 @@
...
+- "Show me more data"
+- "Analyze the results"
+- "Create a visualization"
+
+Given data with columns: name, age, salary, department
+User asked: "Show average salary by department"
+Good suggestions:
+- (scatter_plot) "Plot age vs salary colored by department"
+- (bar_chart) "Show employee count by department"
+- (filter_list) "Filter departments where avg salary > 50000"
```
ahuang11: Shorten good suggestions I think

lumen/ai/prompts/ExplorerUI/follow_up_suggestions.jinja2
```diff
@@ -0,0 +1,47 @@
...
+
+## User's Question
+{{ user_query }}
```
ahuang11: This is unnecessary; it's already in messages.

lumen/ai/prompts/ExplorerUI/follow_up_suggestions.jinja2
```diff
@@ -0,0 +1,47 @@
...
+
+## SQL Executed
+{{ sql }}
+
+## Output Type
+{{ output_type }}
+{% if conversation_history %}
+
+## Previous Questions in This Exploration (do NOT repeat these)
+{% for msg in conversation_history %}- {{ msg }}
```
ahuang11: not really a "conversation" if we're filtering by user

ahuang11 (changes requested):
Still not happy with how it looks under the message; looks really messy / imbalanced, especially with the horizontal scroll bar. It might be because the icons are above the buttons (maybe reverse them in panel-material-ui).

Instead, maybe we can have a automagic (AI) icon button added to the footer (like pencil edit in your other PR), and then on click it replaces/populates the ChatAreaInput with a suggestion instead?

lumen/ai/ui.py
ghostiee-11: Yes, lemme fix it

lumen/ai/prompts/ExplorerUI/follow_up_suggestions.jinja2
ghostiee-11: Okay sure

lumen/ai/coordinator/base.py
```diff
@@ -284,6 +284,11 @@ class Coordinator(Viewer, VectorLookupToolUser):
...
```
ahuang11: Not sure if this PR is ready, but this should point to Coordinator

ghostiee-11: Thanks for the heads-up!! fixed it!

ahuang11:
Np rush. It'll probably go into the next release.

lumen/ai/prompts/Coordinator/follow_up_suggestions.jinja2
```diff
@@ -0,0 +1,28 @@
...
+## SQL Executed
+{{ sql }}
+
+## Output Type
+{{ output_type }}
+{% if previous_queries %}
+
+## Previous Queries (do NOT repeat these)
+{% for q in previous_queries %}- {{ q }}
+{% endfor %}{% endif %}
```
ahuang11: I think we should use the structure of Actor.jinja2

lumen/ai/models.py
```diff
@@ -29,6 +29,25 @@ def model_post_init(self, __context):
...
+        max_length=100
+    )
+
+
+class FollowUpSuggestions(BaseModel):
+    suggestions: list[FollowUpSuggestion] = Field(
+        description="2-3 dataset-specific follow-up suggestions",
+        min_length=1,
+        max_length=3
+    )
```
ahuang11: I still think it's better to populate the `ChatAreaInput.value` with one follow up suggestion for efficiency and look.

ahuang11 (commented):
Sorry if I'm being unclear; I drew a picture to illustrate what I desire from follow up suggestions:


A light bulb/auto awesome lightbulb icon next to the edit/copy/etc icon, and on click, it would populate the chat area input with the suggested text.

Image

ghostiee-11:
Thanks for the drawing Andrew, that made it very clear! Pushed a rework:

- **Single suggestion** - returns one query string, populates `ChatAreaInput.value_input` on click, hides button after
- **Template extends `Actor/main.jinja2`** - uses `instructions`, `examples`, `context` blocks
- **Logic in Coordinator** - `suggest_followup(plan)` returns `str | None`, UI just renders
- **`llm_spec="ui"`** - cheapest model, no chain of thought, model is just `query: str`
- **Lightbulb in `footer_actions`** - inline with edit/copy icons
      (needs panel-extensions/panel-material-ui#605), falls back to `footer_objects`
- Squashed to 1 commit, docs updated

ahuang11 (changes requested):
1. I don't think there should be an auto-generate follow up suggestion; it should just be an icon button that users click when they're out of ideas
2. Inconsistent naming: follow_up is sometimes followup
3. Lightbulb should always show; no need for a toggle

Toggles should be reserved for anything that auto-generates, e.g. Validation agent or SQL planning.

lumen/ai/coordinator/base.py
```diff
@@ -509,6 +514,40 @@ async def _check_tool_relevance(self, tool: Tool, tool_output: str, actor: Actor
...
+                messages=[{"role": "user", "content": "Generate a follow-up suggestion."}],
+                context=plan.out_context,
+                data_summary=data_summary,
+                sql=sql,
+                output_type=output_type,
+                previous_queries=previous_queries,
+            )
+            return result.query
+        except Exception:
+            return None
```
ahuang11, line 629: Maybe it should raise a notification perhaps through pn.state.notifications instead of failing silently.

lumen/ai/ui.py
```diff
@@ -1596,6 +1596,65 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+            if len(self.interface.objects) > num_objects:
+                follow_up_button.visible = False
+
+        follow_up_button = IconButton(
+            icon="lightbulb",
+            description="Suggest a follow-up question",
+            size="small",
+            icon_size="0.9em",
+            margin=(5, 0),
+            on_click=lambda _: state.execute(_generate_suggestion),
```
ahuang11: Is it necessary to call state.execute() vs just _generate_suggestion. Also for consistency, I think it should be called _generate_follow_up. And nested functions should be right under function signature, i.e. move 
```
        if not plan.out_context.get("pipeline"):
            return
        if not plan.out_context.get("data"):
            return

        num_objects = len(self.interface.objects)
```
below

lumen/ai/ui.py
```diff
@@ -1596,6 +1596,65 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+                    with edit_readonly(self._chat_input):
+                        self._chat_input.value_input = suggestion
+                follow_up_button.visible = False
+            except Exception:
+                follow_up_button.disabled = False
+                follow_up_button.description = "Click to retry"
+
+        def hide_follow_up(_=None):
+            if len(self.interface.objects) > num_objects:
+                follow_up_button.visible = False
```
ahuang11: Maybe this can be a rx one-liner, e.g. `IconButton(visible=self...)`; see https://medium.com/@pYdeas/rx-marks-the-spot-a-prescription-for-reactive-python-bdf25e5bf727

Alternatively, there might not be a need to hide it. Maybe user wants to keep stemming from the same conversation.

ghostiee-11:
Addressed all three inline comments + the top-level review:

- **base.py:549** - replaced silent `except` with `state.notifications.warning()` so users see why suggestion failed
- **ui.py:1635** -  renamed to `_generate_follow_up`, moved nested function right under signature, kept `state.execute()` (needed for async on the event loop, same pattern as `replan_button`/`rerun_button`)
- **ui.py:1627** - removed hide logic entirely, button stays visible and re-enables after each click so users can keep generating
- **Naming** - standardized on `follow_up` everywhere
- Removed bare `except Exception: pass` in UI - coordinator already notifies on failure, UI uses `try/finally` to re-enable the button

lumen/ai/coordinator/base.py
```diff
@@ -284,6 +284,11 @@ class Coordinator(Viewer, VectorLookupToolUser):
                 "template": PROMPTS_DIR / "Coordinator" / "tool_relevance.jinja2",
                 "response_model": ThinkingYesNo,
             },
+            "follow_up_suggestions": {
```
ahuang11:
```suggestion
            "follow_up": {
```

lumen/ai/coordinator/base.py
```diff
@@ -509,6 +514,45 @@ async def _check_tool_relevance(self, tool: Tool, tool_output: str, actor: Actor
...
+        sql = plan.out_context.get("sql", "")
+        previous_queries = [
+            m["content"] for m in plan.history
+            if m.get("role") == "user"
+        ]
+        output_type = "chart" if plan.out_context.get("view") else "table"
+
+        try:
+            result = await self._invoke_prompt(
+                "follow_up_suggestions",
```
ahuang11:
```suggestion
                "follow_up",
```

lumen/ai/ui.py
```diff
@@ -1596,6 +1596,58 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+        """
+
+        async def _generate_follow_up():
+            follow_up_button.disabled = True
+            follow_up_button.description = "Generating suggestion..."
+            try:
+                suggestion = await self._coordinator.suggest_follow_up(plan)
+                if suggestion:
+                    with edit_readonly(self._chat_input):
+                        self._chat_input.value_input = suggestion
```
ahuang11: Do we need edit read only? Can't we just do `self._chat_input.value = suggestion`
```suggestion
                    self._chat_input.value = suggestion
```

lumen/ai/ui.py
```diff
@@ -1596,6 +1596,58 @@ async def _add_analysis_suggestions(self, plan: Plan):
...
+            except AttributeError:
+                existing = message.footer_objects or []
+            existing = [
+                obj for obj in existing
+                if getattr(obj, 'name', '') != "FollowUp"
+            ]
+            try:
+                message.footer_actions = existing + [follow_up_button]
+            except AttributeError:
+                message.footer_objects = existing + [follow_up_button]
```
ahuang11: Let's just assume footer actions is implemented, and then bump panel-material-ui pinned requirement.

pixi.toml
```diff
@@ -20,7 +20,7 @@ lint = ["py312", "lint"]
 [dependencies]
 bokeh = "*"
 holoviews = ">=1.17.0"
-panel-material-ui = ">=0.9.0"
+panel-material-ui = ">=0.8.1"
```
ahuang11, line 23: Why pin lower?
https://github.com/panel-extensions/panel-material-ui/releases

ghostiee-11: ohh yes yes i am sorry, fixing this

pixi.toml
```diff
@@ -20,7 +20,7 @@ lint = ["py312", "lint"]
 [dependencies]
 bokeh = "*"
 holoviews = ">=1.17.0"
-panel-material-ui = ">=0.8.1"
+panel-material-ui = ">=0.9.0"
```
ahuang11: https://github.com/panel-extensions/panel-material-ui

Tbh we should probably pin 0.9.2

ghostiee-11: Okay, will pin this to 0.9.2

ghostiee-11: added 0.9.1
>https://github.com/panel-extensions/panel-material-ui/releases/tag/v0.9.1

PR #1778, opened 2026-03-20
feat: Add XArraySource for querying N-dimensional scientific datasets 

SuMayaBee:
## Description:

Adds `XArraySource`, a new Lumen data source for multidimensional scientific datasets (NetCDF, Zarr, HDF5, GRIB), backed by [`xarray-sql`](https://github.com/alxmrs/xarray-sql) and Apache DataFusion. Rebuilds #1434 with full AI pipeline integration.

### What's new:

- New `XArraySource` data source with SQL query support via DataFusion, lazy Dask loading, and CF-convention metadata enrichment
- AI pipeline integration: schema introspection, spatial/temporal context, and derived source creation
- Several fixes to the AI upload flow to ensure the source is correctly registered before the first query executes

This is an initial implementation and there are still edge cases to improve, happy to get feedback! 😀

### Prototype Demo:

https://github.com/user-attachments/assets/7b90f0de-64ac-4652-a84e-69b77d8df18b

---

## How Has This Been Tested?

```bash
pixi run --environment test-312 -- pytest lumen/tests/sources/test_xarray_sql.py -v
```

38 tests covering `get`, `get_schema`, `execute`, schema-prefix stripping, CF role detection, metadata, chunking, and `create_sql_expr_source`. All passing.

---

## AI Disclosure

Tools and Models: Github Copilot + GPT 5.1
---

## Checklist

- [x] Added documentation


lumen/ai/controls/base.py
```diff
@@ -578,6 +587,65 @@ def _add_table(
...
+
+            # Write to a named temp file with the correct extension so xarray
+            # can detect the backend (netcdf4 / zarr)
+            suffix = f".{extension}"
+            tmp = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
+            tmp.write(raw)
+            tmp.flush()
+            tmp.close()
+            tmp_path = tmp.name
+            self._xarray_temp_files.add(pathlib.Path(tmp_path))
```
ahuang11, line 614: This seems highly undesired, especially for large datasets in the 100s of GBs.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,677 @@
...
+    })
+    # Common coordinate name patterns (fallback when CF attrs are absent)
+    _LATLON_NAME_PATTERNS = {
+        'latitude': re.compile(r'^(lat|latitude|y)$', re.IGNORECASE),
+        'longitude': re.compile(r'^(lon|lng|longitude|x)$', re.IGNORECASE),
+        'time': re.compile(r'^(time|t|date|datetime|forecast_time)$', re.IGNORECASE),
+    }
+    # CF unit patterns for coordinate detection
+    _CF_LAT_UNITS = re.compile(r'degrees?\s*_?(north|N)\b', re.IGNORECASE)
+    _CF_LON_UNITS = re.compile(r'degrees?\s*_?(east|E)\b', re.IGNORECASE)
```
ahuang11: Not a huge fan of this; other libraries, like metpy.parse_cf already do this.

lumen/sources/xarray_sql.py
```diff
@@ -0,0 +1,677 @@
...
+    @staticmethod
+    def _epsg_from_wkt(wkt: str) -> str | None:
+        """Try to extract an EPSG code from a WKT CRS string."""
+        match = re.search(r'AUTHORITY\s*\[\s*"EPSG"\s*,\s*"(\d+)"\s*\]', wkt)
+        if match:
+            return f'EPSG:{match.group(1)}'
+        match = re.search(r'ID\s*\[\s*"EPSG"\s*,\s*(\d+)\s*\]', wkt)
+        if match:
+            return f'EPSG:{match.group(1)}'
+        return None
```
ahuang11: Same as above; this shouldn't be Lumen's job I think.

ahuang11 (commented):
Thanks for sharing your proposed PR; really appreciate the work that went into it. At the moment, it feels like the separation of concerns between Lumen and the other packages isn’t quite clear yet.

Also, just a heads up to hopefully save some extra work: we’re currently leaning toward merging https://github.com/holoviz/lumen/pull/1741 soon, which provides essentially the same functionality but without the AI component yet. However really cool demo!

Marking PR as a draft so I remember what I reviewed.

lumen/ai/controls/base.py
SuMayaBee: Good catch! Fixed, switched from read() + write() to shutil.copyfileobj to avoid the redundant in-memory copy. That said, for 100GB+ files the upload control isn't really the right path anyway since the browser has to buffer the transfer. For large datasets, XArraySource accepts file paths and remote URLs directly.

lumen/sources/xarray_sql.py
SuMayaBee: Switched to cf_xarray, removed the custom CF detection and let cf_xarray handle it via ds.cf.coordinates, with a small datetime dtype fallback for bare coords that have no CF attributes. Also added cf-xarray as a dependency. Thanks for the suggestion!

lumen/sources/xarray_sql.py
SuMayaBee: Same as above, Switched to cf_xarray.

SuMayaBee:
> Thanks for sharing your proposed PR; really appreciate the work that went into it. At the moment, it feels like the separation of concerns between Lumen and the other packages isn’t quite clear yet.
> 
> Also, just a heads up to hopefully save some extra work: we’re currently leaning toward merging #1741 soon, which provides essentially the same functionality but without the AI component yet. However really cool demo!
> 
> Marking PR as a draft so I remember what I reviewed.

Thanks for the heads up, really appreciate it!

I had a look at #1741 and there are a few things our `xarray_sql.py` adds that I think would be worth including in future:
1. **Multi-dataset support**: a tables dict that lets you register multiple files/datasets in a single source and query across all of them, rather than one uri at a time
2. **CF metadata via cf_xarray**: automatically detects coordinate roles, spatial bounds and temporal range so the AI can understand the data structure without the user having to describe it
3. **Schema prefix stripping**: LLMs tend to generate FROM SourceName.table style SQL which DataFusion rejects, this handles it transparently

Happy to rebase on top of #1741 once it lands and submit a follow-up PR in future.

ahuang11:
Superseded by https://github.com/holoviz/lumen/pull/1741

PR #1784, opened 2026-03-22
Fix export to image not capturing current view position

### LINKED ISSUE #1773: Export to image doesn't capture current view position
In bokeh, when you save PNG, it saves what is currently visible, but that's because it uses browser's screenshot.

Not sure if we can capture current bounds and inject into the spec before saving?

What I wanted exported in the pane:
Image

Reality:
Image

ghostiee-11:
Fixes #1773

The fix captures the current view bounds from the live panel before the export recreates the component, and ensures the export dimensions match the browser viewport instead of the spec's smaller defaults.

After fix:

vega_lite_editor

### How it was tested

Uploaded a fireball dataset in Lumen AI, generated a world map with `projection: equalEarth`, and exported as PNG. Before the fix, the export showed only North America. After the fix, the full world map is visible matching the browser view. Also added 21 unit tests.


lumen/ai/editors.py
```diff
@@ -290,6 +290,152 @@ class VegaLiteEditor(LumenEditor):
...
+        or None if no zoom state is available.
+        """
+        panel = getattr(self.component, '_panel', None)
+        if panel is None or not hasattr(panel, 'selection') or panel.selection is None:
+            return None
+        bounds = {}
+        for param_name in panel.selection.param:
+            if param_name == 'name':
+                continue
+            value = getattr(panel.selection, param_name)
```
ahuang11: Not a big fan of all the getattr and hasattr

lumen/ai/editors.py
```diff
@@ -290,6 +290,152 @@ class VegaLiteEditor(LumenEditor):
...
+        VegaLiteEditor._inject_encoding_domains(spec, bounds)
+
+        # Recurse into concatenated specs
+        for key in ('hconcat', 'vconcat', 'concat'):
+            if key in spec:
+                spec[key] = [
+                    VegaLiteEditor._apply_bounds_to_spec(sub, bounds)
+                    for sub in spec[key]
+                ]
+        return spec
```
ahuang11: Is there a way to simplify?

lumen/ai/editors.py
```diff
@@ -299,13 +445,45 @@ def export(self, fmt: str) -> StringIO | BytesIO:
...
+        export_height = spec["height"]
         with self.param.update(spec=dump_yaml(spec)):
-            out = self.component.get_panel().export(render_fmt, **kwargs)
+            panel = self.component.get_panel()
+            # Set dimensions directly on the pane so Panel's Vega.export()
+            # uses them instead of its own defaults
+            if hasattr(panel, 'width'):
+                panel.width = export_width
+            if hasattr(panel, 'height'):
+                panel.height = export_height
```
ahuang11: I think all panels should have width.

ahuang11 (commented):
I think the result is desired. However, I'm not sure this is the right implementation; seems overly complex.

lumen/ai/editors.py
ghostiee-11: Yeah sure will fix it

PR #1785, opened 2026-03-22
Add explain query icon to SQL editor

### LINKED ISSUE #1781: Should add an icon to explain query
I think having AI break down and explain the query would be nice.

Image

ghostiee-11:
Closes #1781

When the AI generates a SQL query, there is no easy way to understand what the query does without reading the SQL yourself. This PR adds a brain icon button to the SQL editor toolbar that asks the AI to explain the query in plain language. The explanation streams progressively into the chat.

## Demo

https://github.com/user-attachments/assets/98010211-06b0-40ac-a1cf-b1f90231b43b

## AI Disclosure

I planned the approach, reviewed the existing control patterns (CopyControls, RetryControls, AnnotationControls), and guided the implementation. AI assisted with scaffolding the code and tests.


lumen/ai/prompts/ExplainControls/main.jinja2
```diff
@@ -0,0 +1,12 @@
...
+{%- if language == "sql" %}
+Break down each clause (SELECT, FROM, WHERE, JOIN, GROUP BY, ORDER BY, etc.) and explain what it does.
+{%- elif language == "yaml" %}
+Break down each top-level key and explain what it configures.
+{%- elif language == "json" %}
+Break down each top-level field and explain its purpose.
+{%- else %}
+Break down each section and explain what it does.
+{%- endif %}
+Keep the explanation concise but thorough.
```
ahuang11: Wondering if we should inherit from Actor instead, or instead, a new base blocks_template.jinja2?

```
{%- block global %}
{%- endblock %}

{%- block instructions -%}
{%- endblock -%}

{%- block examples -%}
{%- endblock -%}

{%- block context -%}
{%- endblock %}

{% block footer -%}
{%- endblock -%}
```

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,114 @@
...
+            return
+
+        language = getattr(self.view, "language", "yaml")
+        self._icon.disabled = True
+        message = None
+        try:
+            llm = self.task.actor.llm
+            system = self._render_system_prompt()
+            messages = [
+                {"role": "user", "content": f"Explain this {language}:\n\n```{language}\n{spec}\n```"}
```
ahuang11: `{language}` is overused I think. Mentioned many times in the system prompt and user prompt wasting tokens.

Also I think we should migrate this to the system prompt?

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,114 @@
...
+
+    task = param.Parameter()
+
+    view = param.Parameter(doc="The View containing the spec to explain")
+
+    _clicking = param.Boolean(default=False)
+
+    def __init__(self, **params):
+        super().__init__(**params)
+        language = getattr(self.view, "language", "yaml").upper()
```
ahuang11: In your claude.md or whatever agent file, can you reduce the use of `getattr`?

Since this is on SQLEditor, it should have language.

lumen/ai/editors.py
```diff
@@ -505,6 +507,8 @@ class SQLEditor(LumenEditor):
 
     language = "sql"
 
+    _controls = [RetryControls, ExplainControls, CopyControls]
```
ahuang11: Maybe we can have it on LumenEditor too? Vega lite can be confusing as well.

ahuang11 (changes requested):
I'm excited for this! Perhaps it works like retry/annotation control, allowing users to specify which part they don't understand, but if they don't specify anything (leave blank), then explain the full spec. If so, please submit a new video.

lumen/ai/prompts/ExplainControls/main.jinja2
ghostiee-11: Good call, I'll refactor the prompt to extend from Actor/main.jinja2 using the standard blocks instead of a flat template. Keeps it consistent with how everything else works.

lumen/ai/controls/explain.py
ghostiee-11: Yeah that's true. I'll consolidate so the system prompt mentions the language once, and the user message just sends the raw spec.

lumen/ai/controls/explain.py
ghostiee-11: sure

lumen/ai/editors.py
ghostiee-11: Makes sense, I'll move it to LumenEditor's controls so all editors get it by default.

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,83 @@
...
+        self._text_input.value = ""
+        self.param.update(active=False, _explain_count=self._explain_count + 1)
+
+    @param.depends("_explain_count", watch=True)
+    async def _explain(self):
+        spec = self.view.spec
+        if not spec or not spec.strip():
+            return
+
+        focus = getattr(self, "_focus", "")
```
ahuang11: Is there a need for getattr? If it's undefined initially, I prefer in __init__ self._focus = "". Would appreciate it if you can double check these AI generated code to ensure there's no excess getattr and hasattr.

Also better naming; I was confused as to what `focus` means.

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,89 @@
...
+        message = None
+        try:
+            async for chunk in llm.stream(messages=messages, system=system):
+                if chunk:
+                    message = self.interface.stream(
+                        chunk,
+                        replace=True,
+                        message=message,
+                        user="Assistant",
+                    )
```
ahuang11: In revise(), we have:
```
                new_spec = await self.task.actor.revise(
                    self.instruction, messages, self.task.out_context, self.view
                )
```

I wonder if we should have something like that instead: `self.task.actor.explain()` and under `BaseLumenAgent`, next to `revise_output` we can have `explain_output`?

Theoretically, we can even incorporate `_prepare_vision_messages` if we add an explain method

Just be sure to migrate `lumen/ai/prompts/ExplainControls/main.jinja2` to the proper directory.

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,57 @@
...
+    async def _explain(self):
+        messages = list(self.task.history)
+        message = None
+        try:
+            with self.task.actor.param.update(interface=self.interface):
+                async for chunk in self.task.actor.explain(
+                    self._user_question, messages, self.task.out_context, self.view
+                ):
+                    if chunk:
+                        message = self.interface.stream(
```
ahuang11:
```suggestion
                    if not chunk:
                        continue
                    message = self.interface.stream(
```

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,57 @@
...
+        self._user_question = self._text_input.value_input.strip()
+        self._text_input.value = ""
+        self.param.update(active=False, _explain_count=self._explain_count + 1)
+
+    @param.depends("_explain_count", watch=True)
+    async def _explain(self):
+        messages = list(self.task.history)
+        message = None
+        try:
+            with self.task.actor.param.update(interface=self.interface):
```
ahuang11: Try should only wrap the likely failure part, e.g. explain/stream

ghostiee-11: will fix and push !!

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,58 @@
...
+    """Controls for explaining editor specs using AI.
+
+    Works with any LumenEditor that has a ``spec`` and ``language``
+    attribute. Clicking the icon opens a text input where the user can
+    optionally describe which part to explain. Leaving it blank
+    explains the overall goal.
+    """
+
+    input_kwargs = {
+        "placeholder": "Which part? Leave blank for overview.",
```
ahuang11:
```suggestion
        "placeholder": "Explain which part? Leave blank for all.",
```

lumen/ai/controls/explain.py
```diff
@@ -0,0 +1,58 @@
...
+    optionally describe which part to explain. Leaving it blank
+    explains the overall goal.
+    """
+
+    input_kwargs = {
+        "placeholder": "Explain which part? Leave blank for all.",
+    }
+
+    toggle_kwargs = {
+        "icon": "psychology",
```
ahuang11:
```suggestion
        "icon": "contact_support",
```

ahuang11 (approved):

ghostiee-11:
Great, Thanks for review and auto merge... Lemme fix the tests according to the icon and then ci will be green

PR #1808, opened 2026-04-13
Allow providing LLM tools per Actor

philippjfr:
- Add ability to define `llm_tools` per `Actor`
- Refactor RAG handling to allow LLM to request document lookup


lumen/ai/coordinator/planner.py
```diff
@@ -105,6 +106,7 @@ class Planner(Coordinator):
     and then executes it.
     """
 
+
```
philippjfr:
```suggestion
```

PR #1823, opened 2026-04-18
Support plotting xarray grids (hvPlot + VegaLite + DeckGL)

### LINKED ISSUE #1821: Support Plotting XArray Grids
Right now, I don't think we have support for gridded plotting (or examples for Vega-Lite/DeckGL) so it results in Image

ghostiee-11:
Asking Lumen AI to plot a 2D xarray field (like air temperature over lon/lat) used to fall back to scatter or an empty chart, because the value column had nowhere to go and the agents had no signal the data came from an xarray source. This adds gridded support across hvPlot (quadmesh/image/heatmap), VegaLite (rect heatmap), and DeckGL, so the AI now picks the right 2D encoding automatically. Supersedes #1824 and #1825.

## After

https://github.com/user-attachments/assets/60c0836d-9385-454f-8cdd-ccb1d636d233

Closes #1821.


ahuang11:
I suppose heatmap isn't the right type for plotting gridded data; the better one is hvplot.image or hvplot.quadmesh, and I wonder if there's a way to detect if isinstance(XArraySQLSource) or something.

ghostiee-11:
Good call, thanks. Pushed `f6e188f0` that swings the default toward quadmesh and image.

What changed:

1. hvPlotBaseView.get_plot now pivots long-form pandas to a 2D xarray DataArray (via set_index([y, x])[z].to_xarray()) when kind is in (quadmesh, image, contourf). Without this, hvplot raises NotImplementedError on pandas for those kinds since XArraySQLSource returns pandas via SQL. Verified by test_hvplot_quadmesh_from_longform_df and test_hvplot_image_from_longform_df, both render real hv.QuadMesh and hv.Image instances.
2. Prompt guidance now recommends kind='quadmesh' as the primary choice for xarray-derived data, with image/contourf as alternatives and heatmap as a fallback for ordinal/categorical axes.

3. The isinstance(XArraySQLSource) detection already exists as detect_gridded() in lumen/ai/gridded.py (from the original commit on this PR). It drives the BaseViewAgent prompt's gridded block, so the LLM only gets quadmesh-preferring instructions when the pipeline is actually xarray-backed.

Scope choice to flag: the pivot itself is source-agnostic, it fires whenever kind is quadmesh/image/contourf and x/y/z columns are present and unambiguous. That way a plain CSV with lat/lon/value columns also works if the user explicitly picks quadmesh. If you'd prefer to restrict the pivot to isinstance(pipeline.source, XArraySQLSource) only, happy to tighten it.

Safety guards in _to_gridded:
- xarray passthrough when the input is already a DataArray/Dataset
- _hvplot_xarray_available flag set at __init__, skips pivot if the xarray accessor is unavailable
- _GRIDDED_MAX_CELLS = 10M cap on pivoted grid size, raises ValueError  with a clear message rather than risking OOM
- pivot skipped if duplicate (y, x) pairs make it ambiguous

43 tests pass in lumen/tests/views/ and lumen/tests/ai/test_gridded.py.

lumen/ai/gridded.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+from typing import Any
+
+from ..pipeline import Pipeline
+from ..sources.xarray_sql import XArraySQLSource
+from ..util import check_xarray_available
+
+
+def detect_gridded(pipeline: Pipeline) -> dict[str, Any] | None:
```
philippjfr: Can we just move this into a lumen/util.py?

philippjfr: Actually, `lumen/ai/utils.py`.

ghostiee-11: Suree!!

lumen/views/base.py
```diff
@@ -1029,6 +1042,46 @@ def __init__(self, **params):
...
+            return df
+        if not self._hvplot_xarray_available:
+            return df
+        if not (self.x and self.y and self.z):
+            return df
+        required_cols = {self.x, self.y, self.z}
+        if not required_cols.issubset(df.columns):
+            return df
+        if df.duplicated(subset=[self.y, self.x]).any():
+            return df
```
ahuang11: I feel like this should be consolidated into one `if ... or ...`

lumen/views/base.py
```diff
@@ -938,19 +939,31 @@ class hvPlotBaseView(View):
...
+    _hvplot_xarray_available = False
+
     def __init__(self, **params):
         import hvplot.pandas  # type: ignore
+        try:
+            import hvplot.xarray  # type: ignore
+        except Exception:
+            type(self)._hvplot_xarray_available = False
+        else:
+            type(self)._hvplot_xarray_available = True
```
ahuang11: Slightly confused why `type(self)`._hvplot_xarray_available = False

vs just self._hvplot_xarray_available

lumen/ai/utils.py
```diff
@@ -1409,3 +1410,23 @@ def result_to_dataframe(result) -> pd.DataFrame | None:
...
+    """Return gridded metadata if pipeline.source is an XArraySQLSource, else None.
+
+    Returns a dict with keys ``dims``, ``coords``, ``data_vars``, ``source_type``,
+    or ``None`` for tabular pipelines.
+    """
+    source = pipeline.source
+    if not isinstance(source, XArraySQLSource):
+        return None
+    if not check_xarray_available():
+        return None
```
ahuang11: Combine into one.

lumen/ai/prompts/hvPlotAgent/main.jinja2
```diff
@@ -5,4 +5,16 @@ Generate the plot the user requested. Note that `x`, `y`, `by` and `groupby` fie
 no repeated columns are allowed. Do not arbitrarily set `groupby` and `by` fields unless explicitly requested. If a
 histogram is requested, use `y` instead of `x`. If x is categorical or strings, prefer barh over bar, and use `y` for
 the values.
+{%- if gridded is defined and gridded %}
+
+For gridded xarray data:
+- Prefer kind='quadmesh'. It preserves continuous coordinate axes and is
+  the right default for xarray-derived grids. Use 'image' only for strictly
+  regular grids, 'contourf' for filled contours, or 'heatmap' if you
+  specifically want ordinal/categorical axes.
```
ahuang11: How does the LLM know if it's regular grid or not?

lumen/ai/utils.py
```diff
@@ -1409,3 +1410,23 @@ def result_to_dataframe(result) -> pd.DataFrame | None:
             return None
 
     return None
+
+
+def detect_gridded(pipeline: Pipeline) -> dict[str, Any] | None:
```
ahuang11: This util sounds like it would return a bool instead of a dict... Rename?

lumen/views/base.py
```diff
@@ -1029,6 +1042,46 @@ def __init__(self, **params):
...
+
+        Returns the original df unchanged if x, y, or z are not set, if
+        required columns are missing, if duplicate (y, x) combinations are
+        present, or if the input is already an xarray object (falls back
+        to hvPlot's own path so it raises the informative error).
+
+        Raises ValueError if the pivoted grid would exceed
+        _GRIDDED_MAX_CELLS to prevent OOM.
+        """
+        import pandas as pd
```
ahuang11: Import up top please

lumen/views/base.py
```diff
@@ -1039,8 +1092,14 @@ def get_plot(self, df):
             processed[k] = v
         if self.streaming:
             processed['stream'] = self._data_stream
+        if self.z is not None:
+            processed['C' if self.kind == 'heatmap' else 'z'] = self.z
+
+        plot_source = df
+        if self.kind in self._GRIDDED_KINDS:
+            plot_source = self._to_gridded(df)
```
ahuang11: What happens if returned is just a df instead of a xarray dataset?

ahuang11 (changes requested):

ahuang11:
I'm confused about your screenshots:

image

Are those both showing the before? If it's after, why doesn't it match
image

lumen/views/base.py
ghostiee-11: okay!!

ghostiee-11:
Sorry for the confusion on those, you're right that they didn't actually demonstrate the fix. The original demo used a tiny 4x5 synthetic grid which made the after look mostly uniform and almost identical to the before.

I've redone the demo with a real dataset (`xr.tutorial.air_temperature`, 25 lat by 53 lon) and pulled the comparison back to a single line of view code so the difference is unambiguous:

**Before** (main): the same view spec raises a clear `ValueError` because `'quadmesh'` isn't in the allowed kinds and there's no `z=` param to route the value column.

programmatic-before-main

**After** (this branch): the view pivots the long-form pandas frame to a 2D xarray DataArray and renders the quadmesh with the proper color scale, matching what `ds.hvplot.image("lon", "lat", "air")` produces directly.


programmatic-after-feat



```python
view = hvPlotView(pipeline=pipe, kind="quadmesh", x="lon", y="lat", z="air")
view.get_panel()
```

ghostiee-11:
Pushed d99b5306 which addresses the two follow-ups from your last pass:

1. Collapsed the two early-return guards in `_to_gridded` into a single `if not isinstance(df, pd.DataFrame) or self._gridded_pivot_blocker(df) is not None: return df`, matching the consolidation style you flagged.
2. Added `test_hvplot_gridded_xarray_dataset_renders` to confirm the non-DataFrame pass-through actually produces a `hv.QuadMesh` downstream, not just that the object is returned unchanged. Standard Lumen sources only emit DataFrames, so the test feeds the Dataset directly to `get_plot` to exercise the path.

ghostiee-11:
Hey @ahuang11, Combined the other 2 PR into this one PR so it's easier to review in one place. Folded in the VegaLite rect heatmap guidance and the DeckGL ColumnLayer support (plus a 250k row guard on DeckGLView) that were in #1824 and #1825.

ghostiee-11:
Hey @ahuang11, can you check this once!!

lumen/ai/agents/hvplot.py
```diff
@@ -64,6 +64,9 @@ async def _extract_spec(self, context: TContext, spec: dict[str, Any]):
 
         # Add defaults
         spec["responsive"] = True
+        if spec.get("kind") == "heatmap":
+            spec.pop("by", None)
+            spec.pop("groupby", None)
```
ahuang11: Groupby I think is still valid for heatmap

lumen/views/base.py
```diff
@@ -1030,6 +1042,43 @@ def __init__(self, **params):
         self._linked_objs = []
         super().__init__(**params)
 
+    _GRIDDED_KINDS = ('quadmesh', 'image', 'contourf')
```
ahuang11: Can we import these from hvplot?

lumen/views/base.py
```diff
@@ -45,6 +46,12 @@
 )
 from ..validation import ValidationError
 
+try:
+    import hvplot.xarray  # type: ignore  # noqa: F401
+    _HVPLOT_XARRAY_AVAILABLE = True
+except Exception:
+    _HVPLOT_XARRAY_AVAILABLE = False
```
ahuang11: Can we use the same pattern as try_import_xarray, or try_import_xarray_hvplot?

Alternatively, we can unify with importlib, `def try_import_module(name): importlib ...`

lumen/views/base.py
```diff
@@ -1346,8 +1410,21 @@ class DeckGLView(View):
 
     _panel_type = pn.pane.DeckGL
 
+    # deck.gl serialises every row into the browser as JSON. Beyond a few
+    # hundred thousand records this reliably OOMs the tab, so reject it with a
+    # clear error rather than hanging the browser. Gridded xarray sources make
+    # this easy to hit, since a modest grid expands to lon*lat rows.
+    _MAX_ROWS = 250_000
```
ahuang11: Semi-weird to have ALL_CAPS under a clas

lumen/views/base.py
```diff
@@ -1030,6 +1045,57 @@ def __init__(self, **params):
...
 
+    @staticmethod
+    def _gridded_kinds() -> tuple[str, ...]:
+        """hvPlot kinds we pivot long-form data into a 2D xarray grid for.
+
+        Taken from hvPlot's own gridded-type list so it tracks upstream,
+        limited to the 2D scalar-field kinds that map cleanly from x/y/z
+        columns (points/dataset/rgb are gridded to hvPlot but need no pivot).
+        """
+        from hvplot.converter import HoloViewsConverter
```
ahuang11: Is it necessary to have this nested under a function? I think we import hvplot elsewhere maybe?

lumen/views/base.py
```diff
@@ -1030,6 +1045,57 @@ def __init__(self, **params):
...
+        return tuple(
+            kind for kind in HoloViewsConverter._gridded_types
+            if kind in ("contour", "contourf", "image", "quadmesh")
+        )
```
ahuang11: Isn't this simply doing `return ("contour", "contourf", "image", "quadmesh")` but in a convoluted way?

lumen/ai/prompts/VegaLiteAgent/main_altair.jinja2
```diff
@@ -16,6 +16,12 @@ Generate Altair code. Data is `df`, assign to `chart`, end with `.interactive()`
 - When the n_rows > 5000:
   - You **MUST** include `alt.data_transformers.disable_max_rows()` and `alt.renderers.set_embed_options(renderer="canvas")`
   - Prefer aggregated and binned plots
+{%- if gridded is defined and gridded %}
+
+## Gridded xarray data
+`df` is a long-form grid from an xarray dataset (dimensions: {{ gridded.dims | join(', ') }}; variables: {{ gridded.data_vars | join(', ') }}).
+Use `mark_rect()` for a heatmap: encode `x`/`y` as ordinal coordinate dimensions and `color` as the quantitative value with `alt.Scale(scheme='viridis')`. Do not use geographic projections.
```
ahuang11: Why no geographic projections?

ahuang11:
Thanks! I think this is close. Can you run through xr.tutorial.open_dataset("air_temperature") example and show a video of it being plotted thru Lumen?

ghostiee-11:
Sure..

ghostiee-11:
Ran it through air_temperature

That dataset is ~3.87M rows long-form. gridded kinds are fine (pivot to a small grid), but scatter/table would ship every row to the browser and freeze the tab. deck.gl already caps rows for this, so i added the same cap to hvPlotView/hvPlotUIView. now a per-row render past the limit raises a clear "reduce via LIMIT/aggregation" error instead of hanging. gridded/aggregating kinds + rasterize/datashade exempt.

https://github.com/user-attachments/assets/ad8693a0-cd42-42c6-afd2-c9e46fa34438

ahuang11:
Wait, I thought we're using xql.to_dataset() already; if not, I think this PR should support that or else gridded datasets won't be usable in Lumen.

> That dataset is ~3.87M rows long-form

Separately, does this include time?

ghostiee-11:
Ohh yup, I forgot the xarray-sql thing, this pr was previous of that release will fix this on top of that

ghostiee-11:
yeah it includes time: 2920 time × 25 lat × 53 lon = 3.87M. that's one row per (time, lat, lon) cell in long form, so kept gridded it's just a (2920, 25, 53) array.

ahuang11:
So besides hvplot, I don't think Vegalite and Deckgl can utilize timesteps? If accurate, we should instruct the agent to subset only 1 time slice (or more generally, keep it 2D unless colorby/groupby)

ghostiee-11:
https://github.com/user-attachments/assets/edb880b9-1ec7-4657-84b5-7daf5bbc8fd5

ahuang11:
image
I wonder why the latitudes are flipped

lumen/ai/utils.py
```diff
@@ -1521,45 +1523,60 @@ def _is_regular(values) -> bool:
...
-def subset_gridded_to_2d(pipeline: Pipeline) -> Pipeline:
-    """Pin each extra (non-spatial) dimension of a gridded xarray-backed
-    pipeline to its first value, so views that cannot page a dimension
-    (VegaLite, DeckGL) render a single 2D slice instead of the full grid.
-
-    Returns the pipeline unchanged when it is not gridded, has no extra dims,
-    or xarray is unavailable. Prompt guidance alone cannot do this -- the view
-    agents emit a spec over the pipeline and do not control its rows -- so the
-    reduction has to happen here, deterministically.
+def _spec_field_references(spec: Any) -> set[str]:
```
ahuang11: Should this be more explicit with a package (or something) Literal['vega-lite', 'deckgl'] Literal instead

ahuang11:
Does `gridded_metadata` work with line plots for xarray datasets?

Also, can we do a find_all and replace rename `gridded_metadata` to something that starts with a verb which makes code read more like choppy English, e.g. get_gridded_metadata(pipeline)

lumen/pipeline.py
```diff
@@ -251,15 +251,37 @@ def refs(self) -> list[str]:
...
-
-        # Compute Filter query
         for filt in self.filters:
             filt_query = filt.query
             if (filt_query is not None and not getattr(filt, 'disabled', None) and
                 (filt.table is None or filt.table == self.table)):
                 query[filt.field] = filt_query
+        return query
+
+    def get_dataset(self):
```
ahuang11, line 264: Is this used anywhere?

ghostiee-11: yeah it is, `_source_dataset()` in `views/base.py` wraps it and that gets called in three places: `__panel__`, `get_panel` (so an xarray pipeline gets `hvGridExplorer` instead of the dataframe explorer) and `_gridded_plot_source`.

it's there so gridded kinds can use the source's compact `to_dataset` grid instead of pivoting the long form frame back into xarray, which is what _gridded_pivot_blocker guards against (dupe rows, cell cap).

ahuang11:
Oops. Merge conflicts again :( I should focus on this PR first next time

ghostiee-11:
lol, lemme fix this :)

ghostiee-11:
@ahuang11

ahuang11 (approved):

PR #1824, opened 2026-04-18
Add gridded heatmap guidance to VegaLiteAgent prompt

ghostiee-11:
When the AI receives a gridded xarray-derived dataset and the user asks for a visualisation, the VegaLite agent has no guidance on how to represent 2D fields. It falls back to geographic point charts (longitude/latitude encoding) which render empty for non-geographic grids, or picks an inappropriate mark type.

This adds a guarded block to the VegaLiteAgent prompt that fires only when the pipeline originates from an xarray source. It instructs the LLM to use `mark: rect` with ordinal x/y axes mapped to the coordinate dimensions and a
viridis colour scale on the value column — the correct Vega-Lite pattern for a
2D heatmap from long-form data.

## How to reproduce

1. Load a lat/lon/value CSV and ask "plot air temperature over lat and lon"
2. Before: VegaLite agent uses geographic encoding, chart renders empty
3. After: agent picks `mark: rect` with correct x, y, color mapping

## How Has This Been Tested?

```bash
pytest lumen/tests/ai/test_vega_lite_gridded.py -v    # 2 new prompt tests
pytest lumen/tests/ -x                                 # full suite, no regressions
```

Addresses #1821 (2/3)


lumen/views/base.py
```diff
@@ -1039,6 +1043,8 @@ def get_plot(self, df):
             processed[k] = v
         if self.streaming:
             processed['stream'] = self._data_stream
+        if self.C is not None:
+            processed['C'] = self.C
```
philippjfr: hvPlot generally uses `z`, only heatmap support `C`.

ghostiee-11: Sure thanks will change to use z

lumen/ai/prompts/BaseViewAgent/main.jinja2
```diff
@@ -9,6 +9,14 @@ The current SQL table name: {{ memory['pipeline'].table }}.
 Here is the current dataset to use:
 {{ memory['data'] }}
 
+{%- if gridded is defined and gridded %}
+
+Data originates from an xarray dataset.
+- dimensions: {{ gridded.dims | join(', ') }}
+- data variables: {{ gridded.data_vars | join(', ') }}
+Use kind='heatmap' (or 'quadmesh'/'image'/'contourf') with x/y set to coordinate dimensions and z set to the value column.
```
ahuang11: This isn't a valid kind across VegaLiteView I think...

ghostiee-11:
Folding this into #1823, which combines the hvPlot, VegaLite, and DeckGL gridded support into one PR. Closing to keep it in a single reviewable PR.

PR #1825, opened 2026-04-18
feat: add DeckGL gridded guidance and browser row-count guard

ghostiee-11:
When a user loads an xarray dataset and asks for a DeckGL visualisation, the agent picks a generic scatter spec with no value encoding, producing an empty or broken map.

This adds a conditional block to the DeckGL prompt templates that fires only for xarray sources, steering the LLM toward `ColumnLayer`/`HexagonLayer` with `getPosition` and `getElevation` wired to the actual dimension and variable names. It also raises an explicit `ValueError` in `DeckGLView._get_params` when the frame exceeds 250,000 rows instead of silently OOMing the browser.

## How to reproduce

1. Load a lat/lon/value xarray dataset and ask "show a deck.gl view of the temperature grid"
2. Before: generic ScatterplotLayer, no value encoding
3. After: ColumnLayer with `getElevation` mapped to the value field

## How Has This Been Tested?

```bash
pytest lumen/tests/ai/test_deckgl_gridded.py -v
pytest lumen/tests/ -x
```

Addresses #1821 (3/3)


philippjfr:
Looking good but I'm a little confused. Seems like the three PRs are overlapping to some degree. Could you combine them or indicate in what order I should review/merge them?

ghostiee-11:
These three PRs are a stack, not independent changes. Each one adds a layer on top of the previous:

1. #1823 is the base — it adds the shared `detect_gridded()`  helper, the `C` param on `hvPlotBaseView` (now `z` per your review), and the hvPlotAgent gridded prompt. Nothing else depends on this, it stands alone.

2. #1824 builds on #1823 and adds the VegaLiteAgent prompt block plus its tests. It contains #1823's diff.

3. #1825 builds on #1824 and adds the DeckGLAgent prompt block plus a browser row-count guard on `DeckGLView._get_params`. It contains #1824's diff.

Review/merge order: 1823 → 1824 → 1825. After #1823 lands I will rebase #1824 onto main (which will make it a clean diff against #1823's merged state), same for #1825 after #1824 lands.

If you'd prefer a single combined PR I'm happy to collapse them just let me know.

ghostiee-11:
Folding this into #1823, which includes the DeckGL gridded guidance and the DeckGLView 250k-row guard. Closing to keep it in a single reviewable PR.

PR #1827, opened 2026-04-23
Fix validation agent

### LINKED ISSUE #1822: Chat response should be invalidated (or follow provenance)
The assistant response is outdated; ChatAgent was from two explorations ago, but still injected into ValidationAgent recently, causing ValidationAgent to raise.

Image

```
Assistant Response:
"""
The query retrieves all rows from the **analysed_sst** dataset, skipping the first row with `OFFSET 1`. In simple terms, it’s pulling a table with **942,731 rows and 4 columns**: `time`, `latitude`, `longitude`, and `analysed_sst`.

The data looks like a **sea surface temperature dataset mapped by time and location**. The sample spans **2018-01-01 12:00:00 to 2018-12-01 12:00:00**, with `latitude` from **17.024999618530273** to **30.024999618530273**, `longitude` from **195.02499389648438** to **210.02499389648438**, and `analysed_sst` from **18.3** to **28.4**. **One issue to note:** `analysed_sst` has **38 nulls**, so a few points may be missing from the map.
"""

### PR DESCRIPTION
**Pull request by ahuang11**
## Description

Fixes #1822

## How Has This Been Tested?

image

## AI Disclosure

Tools and Models: {e.g., Cursor + Sonnet 4.6, Claude Code + Opus 4.6, Antigravity + Gemini Flash 3, ChatGPT, etc.}

Opus 4.6

## Checklist

- [ ] Added documentation

### DIFF
--- a/lumen/ai/actor.py
+++ b/lumen/ai/actor.py
@@ -228,11 +228,21 @@ class LLMUser(param.Parameterized):
             msgs = self.interface.serialize(
                 filter_by=lambda msgs: [msg for msg in msgs if isinstance(msg.object, str)]
             )
+            # Walk backward collecting assistant messages.
+            # Messages after the last user message are "current";
+            # messages before it (but after an earlier user message)
+            # are labeled "from previous" so downstream consumers
+            # can distinguish stale context from fresh output.
             history = []
+            found_user = False
             for msg in reversed(msgs):
                 if msg["role"].lower() == "user":
-                    break
-                history.append(f"{msg['role']}: {msg['content']}")
+                    if found_user:
+                        break
+                    found_user = True
+                    continue
+                prefix = "(from previous) " if found_user else ""
+                history.append(f'{prefix}{msg["role"]}: """\n{msg["content"]}\n"""\n')
             prompt_context["chat_history"] = "\n".join(reversed(history))
         return prompt_context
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by philippjfr — CHANGES_REQUESTED** [review]
I deliberately introduced `chat_history` because other agents can generate chat outputs that should be included. So I don't think this will do it. The chat history should always only capture chat outputs since the last user prompt.

**Comment by ahuang11** [comment]
Now shows:

```
Chat History:
(from previous) assistant:"""
This query simply retrieves **all rows and all columns** from the `data_oni_csv` table, so it is showing the full dataset rather than a filtered subset. The table contains **899 rows and 8 columns**. The columns include `year`, `sst_c`, `anom_c`, `cumulative`, `ntotal`, `season`, `threshold`, and `oni`.

In plain language, the data looks like a seasonal climate dataset spanning **1950 to 2024**. `sst_c` ranges from **24.4 to 29.3**, `anom_c` ranges from **-2.0 to 2.6**, and `oni` has three categories: **la_nina**, **neutral**, and **el_nino**. The `season` field has **12 unique values**, and `threshold` has **below**, **between**, and **above**.

**Key takeaway:** the dataset appears to track seasonal sea-surface temperature conditions and related classifications over time. One thing to note is that `anom_c` has a mean of **0.0**, which suggests the anomalies are centered around zero, while `cumulative` and `ntotal` both span from **1 to 50**, indicating a bounded count-like measure rather than a continuous metric.
"""
```

philippjfr (approved):

PR #1848, opened 2026-05-06
Add support for Kaggle datasets under DownloadSourceControls

ahuang11:
## Description

I was trying to find more datasets to experiment with, and I realized Kaggle is a great source for example datasets, so I added it to `DownloadSourceControls`. I imagine we can use it for tutorials / examples in docs, and maybe even down the line, add training ML-models as part of Lumen?? It can also be helpful for us to experiment with Lumen and see if it can reproduce some kaggle analysis.

I was debating whether to have a separate `KaggleSourceControl` but I decided to merge these for convenience. Alternatively, I suppose we could have a separate `KaggleSourceControls` that inherits from DownloadSourceControls, and then it would be the default for UI so new users can easily try out Lumen with Kaggle datasets.

Separately, this PR addresses duplications of translating files into dataframes -> `read_file_to_dataframes` 

## How Has This Been Tested?

image

## AI Disclosure

Used Opus 4.6 to plan and imp

## Checklist

- [x] Added documentation

### DIFF
--- a/docs/configuration/controls.md
+++ b/docs/configuration/controls.md
@@ -29,9 +29,27 @@ Common use cases include:
 | Control | Use for |
 |---------|---------|
 | `UploadControls` | Uploading local files (CSV, Excel, etc.) |
-| `DownloadControls` | Fetching data from URLs |
+| `DownloadControls` | Fetching data from URLs (including Kaggle datasets) |
 
 `UploadControls` stages selected files before processing. After selecting files, click `Confirm file(s)` to process them, or `Clear selected` to reset the staged selection.
+
+#### Kaggle datasets
+
+`DownloadControls` can fetch datasets directly from Kaggle. Paste a Kaggle dataset URL into the URL input field:
+
+```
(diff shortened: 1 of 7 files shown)


docs/configuration/controls.md
```diff
@@ -29,10 +29,28 @@ Common use cases include:
...
+
+Each file in the dataset becomes a separate table in the same data source. This also works through the agent — users can ask the AI to load a Kaggle dataset by providing the URL.
+
+Requires the `kagglehub` package:
+
+```bash
+pip install kagglehub
+```
+
+On first use, `kagglehub` will prompt for Kaggle API credentials (or reads them from `~/.kaggle/kaggle.json`). If `kagglehub` is not installed, Kaggle URL support is silently disabled.
```
ahuang11:
```suggestion
If `kagglehub` is not installed, Kaggle URL support is silently disabled.
```

philippjfr (approved):

PR #1850, opened 2026-05-07
fix: stop max_retries leaking to bare SDK client in _run_tool_loop

ghostiee-11:
When the planner asks for both a structured response and tools, `_run_tool_loop` drops `response_model` and falls back to the bare SDK client, but `max_retries` was left in `kwargs` and reached `AsyncMessages.create` / `AsyncCompletions.create`, which don't accept it. Result: any Anthropic chat that engages tools fails with `Planner could not settle on a plan of action`. Pops it alongside `response_model` on every bare-client branch, and stashes/restores it so the final structured round-trip still gets the caller's retry count.

**Before:**

```
TypeError: AsyncMessages.create() got an unexpected keyword argument 'max_retries'
```

**After:**

```
lumen/tests/ai/test_llm.py::test_run_tool_loop_drops_max_retries_on_bare_client PASSED
```

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -431,27 +431,27 @@ class Llm(param.Parameterized):
         max_tool_rounds: int = 16,
         **kwargs
     ) -> BaseModel | str:
+        # ``max_retries`` is consumed by the instructor wrapper; on bare-client
+        # paths it leaks to the SDK and raises ``TypeError``. Pop once and
+        # re-attach only to the instructor (response_model) round-trip.
+        max_retries = kwargs.pop("max_retries", None)
         requested_stream = bool(kwargs.get("stream", False))
         if requested_stream and structured_model is None:
             # In stream mode we can't know upfront whether a tool will be called.
             # Return the provider stream and let stream() inspect chunks for tool calls.
             kwargs.pop("response_model", None)
-            kwargs.pop("max_retries", None)
             messages = self._normalize_multimodal_messages(messages)
             return await self.run_client(model_spec, messages, **kwargs)
 
         stream = False
-        # ``max_retries`` is consumed by the instructor wrapper; on the raw
-        # client path it leaks through to the SDK and raises ``TypeError``.
-        # Stash it so the final structured-output round-trip can still use it.
-        saved_max_retries = None
         if structured_model is not None and not tool_instances:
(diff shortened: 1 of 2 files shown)


lumen/ai/llm.py
```diff
@@ -470,6 +476,8 @@ async def _run_tool_loop(
 
         if structured_model:
             kwargs["response_model"] = structured_model
+            if saved_max_retries is not None:
+                kwargs["max_retries"] = saved_max_retries
             output = await self.run_client(model_spec, messages_curr, stream=stream, **kwargs)
```
ahuang11: I wonder if we can simplify:

`max_retries = kwargs.pop("max_retries", None)`

at the top, and then where it needs to be consumed:

`output = await self.run_client(model_spec, messages_curr, stream=stream, max_retries=max_retries, **kwargs)`

ahuang11 (changes requested):

lumen/ai/llm.py
ghostiee-11: Sure, we can consolidate it!!

ghostiee-11: Done in the latest push. One small deviation worth flagging: I pop `max_retries` once at the top as you suggested, but only re-attach it on the structured (instructor) calls when it was actually passed in:

```python
max_retries = kwargs.pop("max_retries", None)
...
if structured_model:
    kwargs["response_model"] = structured_model
    if max_retries is not None:
        kwargs["max_retries"] = max_retries
    output = await self.run_client(model_spec, messages_curr, stream=stream, **kwargs)
```

The reason is that `create_kwargs` defaults to `{"max_retries": 1}` and gets baked into the partial in `_get_cached_client`. If I pass `max_retries=max_retries` unconditionally, callers who never set it would end up sending `max_retries=None` and override that default `1`, which would silently disable instructor retries.

Happy to drop the guard and pass it unconditionally if you'd rather have the simpler line, just let me know.

ahuang11 (approved):

lumen/ai/llm.py
ahuang11: Thanks this reads a lot better already.

ghostiee-11: Thanks for the review and merge!! 😊

PR #1852, opened 2026-05-07
Add ASCII logo

ahuang11:
Inspired by running other CLI tools; made by Claude

image


ghostiee-11:
imo, this will look good

PR #1867, opened 2026-05-19
Add STACSource to query STAC catalogs

ghostiee-11:
## Description

Adds `STACSource` so Lumen can point at a SpatioTemporal Asset Catalog (STAC) API and query any cataloged collection as a table. Builds on `XArraySQLSource` (#1741): a collection is resolved to xarray via `xpystac` and the resolved dataset is delegated to an `XArraySQLSource`, so there is no duplicated SQL/schema logic, it just inherits the parent.

This turns the climate-scientist onboarding flow from "download a multi-GB NetCDF, then write a YAML spec" into pointing Lumen at a catalog:

```python
src = Source.from_spec({"type": "stac",
    "url": "https://planetarycomputer.microsoft.com/api/stac/v1"})
src.get_tables()                       # 135 collections, no data fetched
src.execute("SELECT time, x, y, prcp FROM prcp WHERE prcp > 0 LIMIT 5")
```

`get_tables()`/`get_metadata()` stay lazy (no data fetched for discovery). Optional dependency via `pip install lumen[stac]`; `pip install lumen[stac-planetary]` adds Planetary Computer asset signing.

It also fixes two `XArraySQLSource` scalability issues this surfaced: a global chunk spec applied to variables of differing dimensionality, and `get_schema` running `COUNT(*)/DISTINCT/MIN/MAX` over the full grid (now a bounded sample plus an O(1) metadata row count).

### After

Run live against the public Microsoft Planetary Computer STAC API (135 collections discovered; real `daymet-daily-hi` precipitation returned). Executed notebook with real outputs: https://gist.github.com/ghostiee-11/ee42a767740e2720e339b5cf76962654

## How Has This Been Tested?

34 tests covering construction, lazy listing, resolution and delegation, asset selection/signing, datacube metadata, schema, caching, and the optional-import guard, plus xarray_sql coverage for the chunk/schema fixes.

```bash
pytest lumen/tests/sources/test_stac.py -v
```

## AI Disclosure

I planned the architecture and integration strategy. AI tools helped scaffold the implementation and tests based on my design decisions. I reviewed and tested all code manually.


ahuang11:
Thanks! This looks good. Couple questions:
1. Is there a way that the SourceAgent can query this?  (See https://github.com/holoviz/lumen/blob/main/lumen/ai/controls/ingest/catalog.py or https://github.com/holoviz/lumen/blob/main/lumen/ai/controls/ingest/parametric.py for example)
2. I'm wondering if this should be an extension e.g. lumen-stac (I don't know). What are your thoughts? Is pystac a tiny dependency?

ghostiee-11:
Firstly thanks for the review, According to your questions : 
> 1. Is there a way that the SourceAgent can query this?  (See https://github.com/holoviz/lumen/blob/main/lumen/ai/controls/ingest/catalog.py or https://github.com/holoviz/lumen/blob/main/lumen/ai/controls/ingest/parametric.py for example)

`CatalogSourceControls` looks like the right fit. A small `STACCatalogControls` subclass would map cleanly (`_load_catalog `lists the collections, `as_tools()` gives `SourceAgent` the search path), and STAC would be the first concrete user of that base class.

Heads up: I have a follow-up wiring PR nearly ready, but it's a different thing. It adds `catalog <stac-url>` to `lumen-ai serve `so the source loads at startup and shows up in the explorer dropdown. The `SourceAgent` controls would be the in-chat discovery path. Complementary, not overlapping. Happy to tuck `STACCatalogControls` into that follow-up PR

ghostiee-11:
> 2. I'm wondering if this should be an extension e.g. lumen-stac (I don't know). What are your thoughts? Is pystac a tiny dependency?

`pystac` is tiny: `pystac `~210KB, `pystac-client` ~42KB, `xpystac `~9KB. Required deps are `requests / python-dateutil / jsonschema.``xarray` is the only heavy transitive and `lumen[xarray]` already pulls it (STACSource materializes through `XArraySQLSource`), so the marginal cost is ~260KB of pure Python.

Either way if you want lumen-stac, Happy to break it out

lumen/sources/stac.py
```diff
@@ -0,0 +1,425 @@
...
+        collection_id, variable = self._split(table)
+        delegate = self._resolve(collection_id)
+        if variable is None:
+            variable = delegate.get_tables()[0]
+        return delegate, variable
+
+    # ---- BaseSQLSource overrides (fully delegated) ----
+
+    def normalize_table(self, table: str) -> str:
+        return table.strip('"').strip("'").strip("`")
```
ahuang11, line 326: Is there a reason why we override here

lumen/sources/stac.py
```diff
@@ -0,0 +1,425 @@
...
+
+Points Lumen at a SpatioTemporal Asset Catalog (STAC) API. Every cataloged
+collection is exposed as a selectable dataset via ``get_tables()`` without
+fetching any data. When a collection is queried (by the AI agent or directly),
+it is resolved to an xarray dataset through xpystac and handed to an
+``XArraySQLSource`` under the hood, with no further configuration.
+
+To use this source, install the optional STAC dependencies::
+
+    pip install lumen[stac]
```
ahuang11:
```suggestion
    pip install 'lumen[stac]'
```

lumen/sources/stac.py
```diff
@@ -0,0 +1,425 @@
...
+try:
+    import pystac_client
+    import xarray as xr
+    import xpystac  # noqa: F401  (registers the ``stac`` xarray engine)
+
+    from pystac.extensions.datacube import DatacubeExtension
+except ImportError as e:
+    raise ImportError(
+        "STACSource requires the 'pystac-client', 'xpystac' and 'xarray' "
+        "packages. Install them with: pip install lumen[stac]"
```
ahuang11:
```suggestion
        "packages. Install them with: pip install 'lumen[stac]'"
```

ahuang11:
> > 2. I'm wondering if this should be an extension e.g. lumen-stac (I don't know). What are your thoughts? Is pystac a tiny dependency?
> 
> `pystac` is tiny: `pystac `~210KB, `pystac-client` ~42KB, `xpystac `~9KB. Required deps are `` requests / python-dateutil / jsonschema.``xarray `` is the only heavy transitive and `lumen[xarray]` already pulls it (STACSource materializes through `XArraySQLSource`), so the marginal cost is ~260KB of pure Python.
> 
> Either way if you want lumen-stac, Happy to break it out

What about planetary-computer?

We also need to add docs

ahuang11:
```
src._resolve("daymet-daily-hi")  # resolve once via xpystac (cached)
src.execute(
    "SELECT time, x, y, prcp FROM prcp WHERE prcp > 0 LIMIT 5"
)
```

I'm wondering what happens if you `src.get("daymet-daily-hi")`?

image

Should we automatically limit the results in get() so we're not fetching TBs or PBs of data?

Also why does _resolve need to run before execute? It seems like a private method. Shouldn't execute resolve automatically?

lumen/sources/stac.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def get(self, table: str, **query):
+        delegate, variable = self._delegate_table(table)
+        return delegate.get(variable, **query)
+
+    def execute(self, sql_query: str, params=None, *args, **kwargs):
+        if not self._delegates:
+            raise ValueError(
+                "No STAC collection has been resolved yet; query a collection "
+                "table first so it can be opened before running raw SQL."
```
ahuang11: How do you query a collection table first?

I think this should be more actionable

ahuang11:
In addition to the above comments,

> Heads up: I have a follow-up wiring PR nearly ready, but it's a different thing. It adds catalog <stac-url> to lumen-ai serve so the source loads at startup and shows up in the explorer dropdown. The SourceAgent controls would be the in-chat discovery path. Complementary, not overlapping. Happy to tuck STACCatalogControls into that follow-up PR

I'd love to see this alongside this PR first before committing to merging this.

Also, is adlfs needed as dependency? I tried pulling in daymet dataset and it errored.

lumen/sources/stac.py
ghostiee-11: I just mirrored `XArraySQLSource.normalize_table` `_split` uses the bare `collection_id` as a cache key, so a quoted `"daymet-daily-hi:precipitation"` would split to (`'"daymet-daily-hi'`, `'precipitation"')`

Will drop this override anyways!!

ghostiee-11:
> > > 2. I'm wondering if this should be an extension e.g. lumen-stac (I don't know). What are your thoughts? Is pystac a tiny dependency?
> > 
> > 
> > `pystac` is tiny: `pystac `~210KB, `pystac-client` ~42KB, `xpystac `~9KB. Required deps are ` requests / python-dateutil / jsonschema.`xarray `` is the only heavy transitive and `lumen[xarray]` already pulls it (STACSource materializes through `XArraySQLSource`), so the marginal cost is ~260KB of pure Python.
> > Either way if you want lumen-stac, Happy to break it out
> 
> What about planetary-computer?
> 
> We also need to add docs

`planetary-computer ` is already a separate extra (`lumen[stac-planetary]`), and the source auto-wires `planetary_computer.sign` for PC URLs only. Kept it separate since not every STAC catalog is PC, happy to fold it into `lumen[stac]`

Docs: yep, will add

ghostiee-11:
Yep, both of those are bugs:

**`src.get("daymet-daily-hi")`**: today it returns the full materialized xarray grid (straight delegation to `XArraySQLSource.get`), so for daymet you get the whole thing in memory. I'll add an opt-out-able default row limit on `STACSource.get`; explicit `limit=` overrides.

**`_resolve` before `execute`**: agree, private leakage. I'll parse the SQL with `sqlglot` (already a dep) to find referenced tables and auto-resolve them before dispatching, so `execute()` just works on a fresh source. The "no collection resolved" branch then only fires when the table genuinely isn't in the catalog, and I'll rewrite that message to be actionable.

ghostiee-11:
> In addition to the above comments,
> 
> > Heads up: I have a follow-up wiring PR nearly ready, but it's a different thing. It adds catalog  to lumen-ai serve so the source loads at startup and shows up in the explorer dropdown. The SourceAgent controls would be the in-chat discovery path. Complementary, not overlapping. Happy to tuck STACCatalogControls into that follow-up PR
> 
> I'd love to see this alongside this PR first before committing to merging this.
> 
> Also, is adlfs needed as dependency? I tried pulling in daymet dataset and it errored.

will tuck `STACCatalogControls` into this PR so you can see the whole surface before merge.

On adlfs: yes, daymet's COGs live on Azure Blob (abfs://), so adlfs is genuinely required. Will add it to lumen[stac-planetary] since most PC datasets are on Azure.

docs/configuration/sources.md
```diff
@@ -213,6 +214,65 @@ source = XArraySQLSource(
...
+from lumen.sources.stac import STACSource
+import lumen.ai as lmai
+
+source = STACSource(
+    url='https://planetarycomputer.microsoft.com/api/stac/v1'
+)
+source.get_tables()  # list of collection IDs
+
+ui = lmai.ExplorerUI(data=source)
+ui.servable()
```
ahuang11: This doesn't really match the table I description.

`get_tables` doesn't get shown because it's not the last item

docs/configuration/sources.md
```diff
@@ -213,6 +214,65 @@ source = XArraySQLSource(
...
+### STAC catalogs
+
+Point Lumen at a [SpatioTemporal Asset Catalog](https://stacspec.org/) and
+pull each cataloged collection in as an xarray-backed SQL table:
+
+``` bash title="Install dependencies"
+pip install 'lumen[stac]'
+```
+
+For Planetary Computer collections (Azure-hosted), also install the signing
```
ahuang11: What's the signing extra? Is it just stac-planetary?

docs/configuration/sources.md
```diff
@@ -213,6 +214,65 @@ source = XArraySQLSource(
...
+ui = lmai.ExplorerUI(data=source)
+ui.servable()
+```
+
+Each cataloged collection becomes a queryable table. Address a specific
+variable from a multi-variable collection with the `collection:variable`
+table syntax:
+
+``` py
+df = source.get('daymet-daily-hi:prcp')
```
ahuang11, line 246: Where do you get a list of the variables?

lumen/ai/controls/ingest/stac.py
```diff
@@ -0,0 +1,85 @@
...
+
+    search_columns = ["id", "title", "description", "keywords"]
+
+    filter_columns = {
+        "license": {
+            "type": "input", "func": "like", "placeholder": "License...",
+        },
+    }
+
+    detail_columns = ["description", "keywords"]
```
ahuang11, line 53: Should these be private? Or is that inherited from the parent class?

lumen/ai/controls/ingest/stac.py
```diff
@@ -0,0 +1,85 @@
...
+        rows = []
+        for collection in client.get_collections():
+            rows.append({
+                "id":          collection.id,
+                "title":       collection.title or collection.id,
+                # Trim long descriptions so the Tabulator row stays readable;
+                # the full description is available via the detail panel.
+                "description": (collection.description or "")[:300],
+                "license":     collection.license or "",
+                "keywords":    ", ".join(collection.keywords or []),
```
ahuang11: Let's follow PEP-8 ideally

lumen/ai/controls/ingest/stac.py
```diff
@@ -0,0 +1,85 @@
...
+    url = param.String(default=None, allow_None=True, doc="""
+        Root URL of a STAC API endpoint (e.g. a STAC API landing page).""")
+
+    label = "STAC catalog"
+
+    display_columns = {
+        "id":          {"title": "Collection",  "width": "25%"},
+        "title":       {"title": "Title",       "width": "30%"},
+        "description": {"title": "Description", "width": "35%"},
+        "license":     {"title": "License",     "width": "10%"},
```
ahuang11: PEP-8

lumen/sources/stac.py
```diff
@@ -0,0 +1,497 @@
...
+
+    source_type: ClassVar[str | None] = "stac"
+
+    # Asset media-type / role hints that indicate an xarray-openable asset.
+    _xarray_media_tokens: ClassVar[tuple[str, ...]] = (
+        "zarr", "netcdf", "x-hdf", "kerchunk", "reference",
+    )
+    _xarray_roles: ClassVar[frozenset[str]] = frozenset(
+        {"data", "zarr", "references", "index"}
+    )
```
ahuang11, line 88: Run 
https://holoviz-dev.github.io/holoviz-skills/contributing-to-holoviz/cleanup/

docs/configuration/sources.md
```diff
@@ -213,6 +214,65 @@ source = XArraySQLSource(
...
+`get()` is row-capped at `default_limit=100_000` so naive calls against a
+TB-scale collection do not materialize the full xarray grid. Pass
+`default_limit=None` on the source, or pass `sql_transforms=[...]`
+explicitly per call to override.
+
+Public (non-Planetary-Computer) STAC catalogs work the same way:
+
+``` py
+source = STACSource(url='https://earth-search.aws.element84.com/v1')
+```
```
ahuang11, line 268: I couldn't get it working
```
from lumen.sources.stac import STACSource
import lumen.ai as lmai

source = STACSource(url='https://earth-search.aws.element84.com/v1')
source.get("naip")
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/repos/lumen/lumen/sources/stac.py:250, in STACSource._resolve(self, collection_id)
    249     signed = signer(obj) if signer is not None else obj
--> 250     ds = self._open_dataset(signed)
    251 except Exception as e:

File ~/repos/lumen/lumen/sources/stac.py:236, in STACSource._open_dataset(self, obj)
    235     kwargs["patch_url"] = self.patch_url
--> 236 return xr.open_dataset(
    237     obj, engine="stac", chunks=self.chunks, **kwargs
    238 )

File ~/miniconda3/envs/lumen/lib/python3.12/site-packages/xarray/backends/api.py:613, in open_dataset(filename_or_obj, engine, chunks, cache, decode_cf, mask_and_scale, decode_times, decode_timedelta, use_cftime, concat_characters, decode_coords, drop_variables, create_default_indexes, inline_array, chunked_array_type, from_array_kwargs, backend_kwargs, **kwargs)
    607 backend_ds = backend.open_dataset(
    608     filename_or_obj,
    609     drop_variables=drop_variables,
    610     **decoders,
    611     **kwargs,
    612 )
--> 613 ds = _dataset_from_backend_dataset(
    614     backend_ds,
    615     filename_or_obj,
    616     engine,
    617     chunks,
    618     cache,
    619     overwrite_encoded_chunks,
    620     inline_array,
    621     chunked_array_type,
    622     from_array_kwargs,
    623     drop_variables=drop_variables,
    624     create_default_indexes=create_default_indexes,
    625     **decoders,
    626     **kwargs,
    627 )
    628 return ds

File ~/miniconda3/envs/lumen/lib/python3.12/site-packages/xarray/backends/api.py:300, in _dataset_from_backend_dataset(backend_ds, filename_or_obj, engine, chunks, cache, overwrite_encoded_chunks, inline_array, chunked_array_type, from_array_kwargs, create_default_indexes, **extra_tokens)
    296     raise ValueError(
    297         f"chunks must be an int, dict, 'auto', or None. Instead found {chunks}."
    298     )
--> 300 _protect_dataset_variables_inplace(backend_ds, cache)
    302 if create_default_indexes:

File ~/miniconda3/envs/lumen/lib/python3.12/site-packages/xarray/backends/api.py:111, in _protect_dataset_variables_inplace(dataset, cache)
    110 def _protect_dataset_variables_inplace(dataset: Dataset, cache: bool) -> None:
--> 111     for name, variable in dataset.variables.items():
    112         if name not in dataset._indexes:
    113             # no need to protect IndexVariable objects

AttributeError: 'NoneType' object has no attribute 'variables'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In[4], line 5
      1 from lumen.sources.stac import STACSource
      2 import lumen.ai as lmai
      3 
      4 source = STACSource(url='https://earth-search.aws.element84.com/v1')
----> 5 source.get("naip")

File ~/repos/lumen/lumen/sources/stac.py:326, in STACSource.get(self, table, **query)
    325 def get(self, table: str, **query):
--> 326     delegate, variable = self._delegate_table(table)
    327     if self.default_limit is not None and "sql_transforms" not in query:
    328         query["sql_transforms"] = [SQLLimit(limit=self.default_limit)]

File ~/repos/lumen/lumen/sources/stac.py:296, in STACSource._delegate_table(self, table)
    294 def _delegate_table(self, table: str) -> tuple[XArraySQLSource, str]:
    295     collection_id, variable = self._split(table)
--> 296     delegate = self._resolve(collection_id)
    297     if variable is None:
    298         variable = delegate.get_tables()[0]

File ~/repos/lumen/lumen/sources/stac.py:259, in STACSource._resolve(self, collection_id)
    257     last_error = last_error or ValueError("no data variables resolved")
    258 if dataset is None:
--> 259     raise ValueError(
    260         f"STAC collection {collection_id!r} could not be opened as an "
    261         f"xarray dataset via xpystac ({type(last_error).__name__}: "
    262         f"{last_error}). Its assets may need a reader (e.g. "
    263         "stackstac/odc-stac for COGs) or storage credentials/extra "
    264         "filesystem packages (e.g. adlfs) not configured here."
    265     ) from last_error
    266 # CF "bounds" variables (a small vertices dim like nv/bnds) are not
    267 # queryable tables and break tabular chunk handling; drop them.
    268 bounds_dims = [
    269     d for d in dataset.dims if d in ("nv", "bnds", "bounds", "vertices")
    270 ]

ValueError: STAC collection 'naip' could not be opened as an xarray dataset via xpystac (AttributeError: 'NoneType' object has no attribute 'variables'). Its assets may need a reader (e.g. stackstac/odc-stac for COGs) or storage credentials/extra filesystem packages (e.g. adlfs) not configured here.
```

ahuang11 (changes requested):
I think needs cleanup, clearer instructions, and fixing tests.

lumen/ai/controls/ingest/stac.py
ghostiee-11: Inherited. `CatalogSourceControls` declares `display_columns`, `search_columns`, `filter_columns`, and `detail_columns` as public `param.Dict` / `param.List` ([`lumen/ai/controls/ingest/catalog.py:42-55`](https://github.com/holoviz/lumen/blob/main/lumen/ai/controls/ingest/catalog.py#L42-L55)), so I'm overriding the defaults under the same public names. Keeping them public to match the base class.

docs/configuration/sources.md
ghostiee-11: Fixed: split into its own block with `print(source.get_tables())` so the listing actually shows in a REPL / notebook.

docs/configuration/sources.md
ghostiee-11: Renamed: now says "install `lumen[stac-planetary]` to pull in `planetary-computer` and `adlfs`" instead of vaguely calling it "the signing extra".

docs/configuration/sources.md
ghostiee-11: Added a `source.get_schema(collection)` step before the `collection:variable` example so users can see what is queryable on a given collection.

docs/configuration/sources.md
ghostiee-11: Fixed. `_is_xarray_asset` used to trust the `data` role even when `media_type` said `image/tiff; profile=cloud-optimized`, so COG assets reached xpystac and blew up. Tightened it so `media_type` is authoritative, and skipped the first-item fallback when the item has no xarray-openable asset. Your snippet now returns:

```
ValueError: STAC collection 'naip' has no xarray-openable asset (saw media_types: ['application/xml', 'image/tiff; application=geotiff; profile=cloud-optimized']); for COG or other image collections use stackstac or odc-stac instead.
```

NAIP is COG-only so xpystac can't open it regardless. For an end-to-end STACSource example use a Zarr-backed PC collection like `daymet-daily-hi`. Native COG support (stackstac fallback) is a planned follow-up.

PR #1869, opened 2026-05-19
show popout only if >1

ahuang11:
Only show if > 1 tabs
image

Also change to primary color to match other icons
image


lumen/ai/ui.py
```diff
@@ -2447,8 +2447,18 @@ def pop_out(event):
...
         )
+
+        def _sync_pop_button_visibility(*events):
+            if _tab_index() is None:
+                # Popped out: always show (to allow closing the split)
+                pop_button.visible = True
+            else:
+                pop_button.visible = len(tabs) > 1
+
+        tabs.param.watch(_sync_pop_button_visibility, 'objects')
```
```suggestion
            margin=(5, 0, 0, 0), on_click=pop_out, styles={"margin-left": "auto"},
            visible=tabs.param['objects'].rx.len() > 1, color="primary"
)
```

philippjfr: Ah, `_tab_index()`.

PR #1871, opened 2026-05-20
Fix tests to address deprecation warnings

ahuang11:
Fix deprecation tests:
1. name -> label
2. button_type -> color
3. sqlglot newer version fixes


philippjfr:
My bad, didn't see this one.

PR #1872, opened 2026-05-20
Fix TableListAgent to list one or more data sorces available

ciaograsso06:
# Ensure suggestions use synced sources and allow table listing for single-source contexts

## Summary

This PR improves how `ExplorerUI` handles data sources provided at initialization time when the user interacts with default suggestions.

It introduces two small changes:

1. Before submitting a suggestion, `ExplorerUI` syncs the current sources when a coordinator and sources are available.
2. `TableListAgent.applies()` now supports single-table contexts and falls back to checking `source` / `sources` when `visible_slugs` is not available yet.

## Motivation

In simple `ExplorerUI` apps, data can be provided directly through the `data` argument:

```python
ui = lmai.ExplorerUI(
    data=("penguins", "penguins.csv"),
    llm=llm,
)
```

The data source is available, and SQL-style prompts such as:

```text
show me the first rows of data
```

can work correctly.

However, clicking the default suggestion:

```text
What data is available?
```

may fail to list the available data, especially when only one table is present or when the source visibility state has not been synced before the suggestion is submitted.

### Sync sources before executing suggestions

When a suggestion is clicked, the UI now ensures sources are synced before submitting the suggestion text:

```python
if hasattr(self, "_coordinator") and self.context.get("sources"):
    await self._sync_sources(global_context=self.context)
```

This helps ensure that source-related context, such as available sources and visible tables, is up to date before the suggestion is processed.

### Make `TableListAgent` apply to single-table contexts

Previously, `TableListAgent` only applied when more than one table was visible. This excluded valid single-table cases.

The new logic applies when:

- `visible_slugs` is present and non-empty;
- or `source` exists and has tables;
- or any source in `sources` has tables.

```python
async def applies(cls, context: TContext) -> bool:
    visible_slugs = context.get("visible_slugs")
    if visible_slugs:
        return True

    source = context.get("source")
    if source is not None:
        return bool(source.get_tables())

    sources = context.get("sources", [])
    return any(source.get_tables() for source in sources)
```

## Result

With this change, default suggestions such as:

```text
What data is available?
```

can correctly detect and list available data in simple `ExplorerUI` apps, including cases where only one table/source is loaded.

## Notes

This PR does not change how data sources are created. It only makes the existing source context more reliably available before suggestions are submitted, and makes `TableListAgent` applicable to valid single-table contexts.


ahuang11:
Hi! Thanks for contributing. Three requests:
1. can you follow the PR template?
2. run cleanup skill https://holoviz-dev.github.io/holoviz-skills/contributing-to-holoviz/cleanup/
3. share screenshots / video captures of this in action
Also, I think TableListAgent shouldn't show for single tables.

ciaograsso06:
Okay, I will consider all the 3 points and make the necessary changes.

lumen/ai/agents/table_list.py
```diff
@@ -86,7 +86,16 @@ def row_content(row):
 
     @classmethod
     async def applies(cls, context: TContext) -> bool:
-        return len(context.get('visible_slugs', set())) > 1
+        visible_slugs = context.get("visible_slugs")
+        if visible_slugs:
```
philippjfr:
```suggestion
        if visible_slugs and len(visible_slugs) > 1:
```

lumen/ai/agents/table_list.py
```diff
@@ -86,7 +86,16 @@ def row_content(row):
...
+            return True
+
+        source = context.get("source")
+        if source is not None:
+            return bool(source.get_tables())
```
philippjfr:
```suggestion
            return len(source.get_tables()) > 1
```

lumen/ai/agents/table_list.py
```diff
@@ -86,7 +86,16 @@ def row_content(row):
...
+
+        sources = context.get("sources", [])
+        return any(source.get_tables() for source in sources)
```
philippjfr:
```suggestion
        return sum(len(source.get_tables()) for source in sources) > 1
```

lumen/ai/ui.py
```diff
@@ -1604,6 +1604,14 @@ async def use_suggestion(event):
...
                         return
                 self._update_main_view()
+                if hasattr(self, "_coordinator") and self.context.get("sources"):
+                    await self._sync_sources(global_context=self.context)
+
+                if not analysis:
+                    with edit_readonly(self._chat_input):
+                        self._chat_input.value_input = contents
+                    self._on_submit()
+                    return
```
```suggestion
```

ahuang11 (approved):

ahuang11:
Thanks for contributing!

PR #1879, opened 2026-06-01
Fix SQLAgent crash on sources with the legacy 'any' dialect

ghostiee-11:
## Description

While testing `lumen-ai serve --catalog <stac-url>` I hit `ValueError: Unknown dialect 'any'` from `clean_sql` → `sqlglot.transpile(read='any')`. sqlglot 28.x removed the legacy `'any'` value while `BaseSQLSource.dialect` still ships it as the default, so any source without a concrete dialect crashes the `SQLAgent` chat path before the SQL is dispatched. Normalise `'any'` → `None` (dialect-agnostic mode) inside `clean_sql`; backward compatible.


ahuang11:
Would it be better to make dialect default to None instead of any?

ghostiee-11:
Yeah, we can turn to None, it will be cleaner i think to None instead any

philippjfr (approved):

PR #1889, opened 2026-06-12
feat: add Arraylake source control

ghostiee-11:
Adds `ArraylakeSourceControls`, a parametric input control that opens an Arraylake repository (repo / branch / group) and registers it as an `XArraySQLSource`, so Lumen AI can query Icechunk/Zarr stores the same way uploaded CSVs become a `DuckDBSource`. It drops 0-dim scalar metadata vars, unifies inconsistent chunks, and sizes the Icechunk snapshot cache so real-world stores like GOES register cleanly and fast (~200s to ~16s open).

```python
from lumen.ai.controls import ArraylakeSourceControls

ui = ExplorerUI(source_controls=[ArraylakeSourceControls])
# User enters repo="earthmover-public/goes-16",
# group="ABI-L2-MCMIPF/post-2023-04-19"  ->  registered as a queryable source
```
**Result (live `earthmover-public/goes-16`):** 

arraylake_goes


lumen/ai/controls/ingest/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .arraylake import ArraylakeSourceControls
```
ahuang11: Should this be guarded if imports are available? Try/except?

lumen/ai/controls/ingest/arraylake.py
```diff
@@ -0,0 +1,148 @@
...
+        group = (params.get("group") or self.group).strip() or None
+
+        try:
+            import arraylake as al
+            import xarray as xr
+        except ImportError:
+            return SourceResult.empty(
+                "Arraylake support requires `pip install lumen[arraylake]` "
+                "(Python >=3.12)."
+            )
```
ahuang11, line 99: Maybe use as util?

ahuang11:
Can you try it out on the UI? Ask the agent to query for something.

lumen/ai/controls/ingest/__init__.py
ghostiee-11: Yess!! we can guard it and, add try/except.

lumen/ai/controls/ingest/arraylake.py
ghostiee-11: Yeah, will add this in util...

ghostiee-11:
Dug into the slow load and validated it. The open is fine now (sizing the Icechunk snapshot cache took it from ~205s to ~16s, that config is in the control). The rest is registration, and it is an xarray-sql scaling issue rather than a control problem.

xarray-sql creates one scan partition per native chunk and enumerates them all at registration. I measured it on tiny synthetic data (same data, only the chunking changes):

| partitions | register time |
| --- | --- |
| 1,000 | 0.06s |
| 10,000 | 0.24s |
| 100,000 | 2.45s |

So ~25us per partition, linear in the number of chunks.

GOES is chunked one timestep per chunk: `CMI_C13` is `t=102,988` chunks of size 1, `y`/`x` are `24` chunks of 226. That is `102988 x 24 x 24 = ~59M` partitions for a single variable, which is ~25-30 min just to register one variable. `XArraySQLSource` registers all 188 variables separately on top of that, so it never finishes. Coord pre-loading and coarse time-chunking on the full store did not help (still minutes).

The real fix is upstream in xarray-sql: coalesce native chunks into a bounded number of scan partitions (or enumerate lazily) so registration cost is independent of how finely the store is chunked. That fixes it for the current `XArraySQLSource` without changes here. A secondary Lumen improvement would be registering per dimension-group instead of per-variable, which also gives cleaner tables. Filed the xarray-sql issue with the repro: alxmrs/xarray-sql.

The control itself works well on a subset (registers in ~16s, queries in ~1s), and it shows up and loads in the UI under its own Arraylake tab.

ahuang11:
Nice find! Can you show a screen recording of it please?

ghostiee-11:
Suree!!

ghostiee-11:
https://github.com/user-attachments/assets/edff6bf8-e246-46ee-a61a-b86f190ac494

ahuang11:
Thanks for the screenshot; is there a way for users to discover these values? i.e. can we somehow have a dropdown instead? Perhaps there are central repos like earthmover? I'm thinking a new Select, EarthMover, NWS, AWS, Custom, and for Custom, users input as you did, but for the others, ideally there would be a dropdown.

image

ghostiee-11:
Hey @ahuang11, On making the dropdowns dynamic: the group can be auto-discovered, but the repo list can't.

`list_repos` returns 403 for any org you don't own, so `earthmover-public` can't be enumerated, and S3 buckets aren't listable. So the provider/repo list stays curated.

Groups are dynamic. Once a repo is open I walk its zarr tree and populate the Group dropdown from the groups that actually hold data, so no group paths are hardcoded (they were fragile anyway, ERA5's real group is `single/spatial`, not `single`).

So, How should we move forward ?

ahuang11:
Can you share how you manually discover the repos? Is there a link?

ghostiee-11:
No link to list them. `list_repos` 403s for `earthmover-public` even when logged in (it needs org *membership*, not just auth). Public repos are readable one at a time via `get_repo`, but the org's repo list stays private, so it can't be enumerated.

So I found them by probing names with `get_repo`. `goes-16` is in their [virtual-zarr blog](https://www.earthmover.io/blog/virtual-zarr); `era5`/`gfs`/`hrrr` I confirmed the same way. Net: repos stay curated, groups stay dynamic.

ahuang11:
Do you mind asking around in https://discourse.pangeo.io/?

ghostiee-11:
> Do you mind asking around in https://discourse.pangeo.io/?

Yup, Thanks for sharing this, will create a Topic here

ahuang11:
I created one https://discourse.pangeo.io/t/how-to-see-all-the-repos-in-arraylake/5764

ghostiee-11:
Thanks a lot, Andrew!!

ahuang11:
image

Can the tooltip give an example like the repo?

image

Also, in my testing, it doesn't look like it's responsive after clicking Fetch Data, which is poor UX. We need to allow the main thread to trigger loading=True by asyncio.sleep(0.01).

https://github.com/holoviz/lumen/pull/1889#issuecomment-4694498730

Also, would it be possible to subset tables so we don't have to wait for all the tables to load?

PR #1890, opened 2026-06-13
fix: surface valid table names on SQL validation 'table not found' retry

ghostiee-11:
When the SQL agent generates a query that references a table that doesn't exist (e.g. the **source name** instead of one of its tables, which the model does for multi-table sources like `XArraySQLSource` where each variable is its own table), validation fails with "table not found". The retry feedback was just the raw error string, so the model kept making the same mistake and exhausted its retries (`RetriesExceededError`). This surfaces the source's actual table names in the retry feedback so the model recovers.

**Before**
```text
table 'UploadedSource000002' not found
  -> retry feedback = raw error -> model guesses again -> RetriesExceededError
```

**After**
```text
table 'UploadedSource000002' not found
  -> retry feedback: "Use one of these exact table names, not the source name:
     'CMI_C01', 'DQF_C01', ..."
  -> model selects CMI_C01 -> validates
```

Surfaced while testing the Arraylake source control (#1889) against a real multi-table GOES `XArraySQLSource`. Adds two deterministic tests for `_validate_sql` (no LLM needed).

### DIFF
--- a/lumen/ai/agents/sql.py
+++ b/lumen/ai/agents/sql.py
@@ -377,6 +377,18 @@ class SQLAgent(BaseLumenAgent):
                 feedback = f"{type(e).__name__}: {e!s}"
                 if "KeyError" in feedback:
                     feedback += " The data does not exist; select from available data sources."
+                elif "not found" in str(e).lower():
+                    # A referenced table does not exist (e.g. the model used the
+                    # source name instead of one of its tables, common for
+                    # multi-table sources like XArraySQLSource). Surface the
+                    # actual table names so the retry stops guessing.
+                    available = source.get_tables()
+                    if available:
+                        names = ", ".join(repr(t) for t in available[:50])
+                        feedback += (
+                            f" Use one of these exact table names, not the "
+                            f"source name: {names}."
+                        )
 
                 retry_result = await self.revise(
                     feedback, messages, context, spec=sql_query, language=f"sql.{source.dialect}",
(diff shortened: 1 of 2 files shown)


lumen/ai/agents/sql.py
```diff
@@ -377,6 +377,18 @@ async def _validate_sql(
                 feedback = f"{type(e).__name__}: {e!s}"
                 if "KeyError" in feedback:
                     feedback += " The data does not exist; select from available data sources."
+                elif "not found" in str(e).lower():
+                    # A referenced table does not exist (e.g. the model used the
+                    # source name instead of one of its tables, common for
+                    # multi-table sources like XArraySQLSource). Surface the
+                    # actual table names so the retry stops guessing.
+                    available = source.get_tables()
```
ahuang11, line 385: What happens if there are over 100 of tables in a database? Won't that consume too much context? I think we may need to do similarity search, perhaps through Metaset?

ahuang11 (changes requested):

lumen/ai/agents/sql.py
ghostiee-11: Good catch on the ordering. It's already capped at `[:50]` so context stays bounded, but `get_tables()` comes back in arbitrary order, so with a lot of tables the slice can drop the one the model actually needs.

The metaset's a better fit, and it's already in context here. `get_top_tables()` ranks by similarity to the query, so I'll pull the relevant names for that source from it and only fall back to `get_tables()` when it's empty. Pushing that shortly.

ahuang11: I think we need to mention it's truncated if it's over 50

PR #1599, opened 2026-01-14
Add the ability to add filters to exploration

philippjfr:
Adds UI to add and remove filters from a `SQLOutput` in the UI

filters

PR #1894, opened 2026-06-23
Add coordinate-aware filters to exploration outputs (completes #1599)

ghostiee-11:
## Overview

Completes Philipp's stalled draft #1599: exploration outputs get an "Add Filter" menu, and for xarray sources it surfaces coordinate dimensions (time, lat, lon) first and filters them through the existing pipeline. Also adds an `apply_filter` tool the chat agent can use. Built on the existing filter machinery, not a parallel widget stack.

## Demo
https://github.com/user-attachments/assets/8a128f08-b75d-46f3-89b2-29e50e474259

### DIFF
--- a/lumen/ai/editors.py
+++ b/lumen/ai/editors.py
@@ -523,15 +523,18 @@ class SQLEditor(LumenEditor):
         self._filters: dict[str, WidgetFilter] = {}
         self._filter_area = FlexBox(
             sizing_mode="stretch_width", justify_content="space-evenly",
-            max_height=300, styles={"overflow-y": "auto"},
+            styles={"gap": "8px", "align-content": "flex-start"},
         )
         # The filter widgets live in a Paper that the exploration view places
         # above the editor/table split (see ExplorerUI._render_view), so adding
         # filters pushes both the SQL editor and the results table down. Only
-        # shown once at least one filter has been added.
+        # shown once at least one filter has been added. The Paper fills its
+        # split pane and scrolls internally (padding for margins) so tall
+        # multi-select filters stay contained instead of overlapping the editor.
         self._filter_paper = Paper(
-            self._filter_area, elevation=2, margin=(5, 10),
-            sizing_mode="stretch_width", visible=False,
+            self._filter_area, elevation=2, margin=(8, 10),
+            sizing_mode="stretch_both", styles={"overflow-y": "auto", "padding": "10px"},
+            visible=False,
         )
         return editor
 
@@ -595,14 +598,16 @@ class SQLEditor(LumenEditor):
                 # Cap each filter's width so two fit per row; the FlexBox's
                 # space-evenly justification gives equal gaps. Use the compact
                 # size, hide the always-on value, and surface the range as a
-                # hover tooltip (description). size/show_value are slider-only,
-                # so guard for non-slider filters (e.g. Select).
+                # hover tooltip (description).
                 widget_opts = {
                     "width": 180, "margin": (8, 5),
                     "description": self._filter_tooltip(self.component.schema[field]),
                 }
                 params = filt.widget.param
-                if "size" in params:
+                # `size` is a small/medium/large Selector on sliders but a
+                # visible-rows Integer on MultiChoice/TextInput, so only set it
+                # where "small" is a valid choice. show_value is slider-only.
+                if "size" in params and "small" in (getattr(params["size"], "objects", None) or []):
                     widget_opts["size"] = "small"
                 if "show_value" in params:
                     widget_opts["show_value"] = False
--- a/lumen/ai/editors.py
+++ b/lumen/ai/editors.py
@@ -549,15 +549,23 @@ class SQLEditor(LumenEditor):
             return "text_fields"
         return self._filter_icons.get(col_type, "help")
 
-    def _filter_items(self) -> list[dict[str, str]]:
+    def _filter_items(self) -> list[dict[str, Any]]:
         # Coordinate dimensions (if any) are listed first, then data variables /
         # tabular columns. Tabular sources carry no "dimension" flag, so their
-        # ordering is unchanged.
+        # ordering is unchanged. Columns with an active filter show a filled
+        # check (review: make the active state visible in the menu).
+        active = {filt.field for filt in self.component.filters}
         dimensions, variables = [], []
         for col, col_schema in self.component.schema.items():
             if col == "__len__":
                 continue
-            item = {"label": col, "icon": self._filter_icon(col_schema)}
+            item = {
+                "label": col,
+                "icon": self._filter_icon(col_schema),
+                "active_icon": "check_circle",
+                "active_color": "primary",
+                "toggled": col in active,
+            }
             (dimensions if col_schema.get("dimension") else variables).append(item)
         return dimensions + variables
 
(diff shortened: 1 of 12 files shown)


ahuang11:
Looks cool! Maybe the widgets should have some margins + stretch width? Also, maybe we can put it on a `pmui.Paper` so it looks more in place?

ghostiee-11:
> Looks cool! Maybe the widgets should have some margins + stretch width? Also, maybe we can put it on a `pmui.Paper` so it looks more in place?

Sounds good, Andrew!! Will surely do that

ghostiee-11:
Hey @ahuang11, your views on this!!

https://github.com/user-attachments/assets/f3fbc23a-bc98-4f52-8a90-58489686bfee

ahuang11:
Nice! My first impressions, it's quite a long widget. Maybe we should set a max width, especially if users expand the right side? Also, maybe we can utilize https://panel-material-ui.holoviz.org/reference/layouts/FlexBox.html 

Also, if a column widget is visible/invisible, maybe make it more intuitive, e.g. having a MenuToggle instead of a MenuButton (or whatever is being used). https://panel-material-ui.holoviz.org/reference/menus/MenuToggle.html

Lastly, given that you can just hide the widgets by dragging up/dragging down, I'm torn on whether users need to add filters or no, or simply just add all the filters by default (I guess if there's 100 columns, then probably not).

ghostiee-11:
Okay Andrew, lemme look to it..

also idts we should just add all the filters by default

ghostiee-11:
How about this @ahuang11 

https://github.com/user-attachments/assets/b620fc08-6ed2-40f5-af5c-0acf69e392ed

ghostiee-11:
i should reduce the vertical space when i add any filter...

ahuang11:
Wonder if `show_value=False` would be cleaner. I also kind of expected the menu icons to toggle between filled / not-filled or even simple check boxes.

ghostiee-11:
https://github.com/user-attachments/assets/a51cce2a-99f2-4792-9d37-51e7b5980b7d

ghostiee-11:
https://github.com/user-attachments/assets/b952d298-c70c-43b1-8f8e-a493036f213d

ahuang11:
Nice! Thanks for the screenshares. I wonder if we can strip the `:` from the labels if there's no value (maybe a pmui change)?

ghostiee-11:
Yeah sure!!

ghostiee-11:
Hey @ahuang11, Can you check this once

ahuang11:
image

Functionally is great! However, I think visually, the filters need more UI tweaks, specifically margins and maybe max height?

image

ahuang11:
Could also play with size
image

And instead of the i icon, maybe can just always show tooltip.
image

ghostiee-11:
Whoa.. nice ideas!!

ghostiee-11:
In my opinion, it doesn't look clean, even after removing the values after the for. eg, lat:  then `tooltips = True`


Screenshot 2026-07-01 at 2 35 42 AM

ahuang11:
Okay, yeah remove that tooltips=True. I also noticed the values re-appeared.

ghostiee-11:
Screenshot 2026-07-02 at 6 55 32 PM
Screenshot 2026-07-02 at 6 55 39 PM

ghostiee-11:
Hey @ahuang11, In my opinion feels clean. If any more suggestions surely love to iterate over that.

https://github.com/user-attachments/assets/681da0a5-48ed-47a7-8be2-cbe686f31591

ahuang11:
image

image

It doesn't seem to work for strings

image

lumen/ai/editors.py
```diff
@@ -510,6 +511,128 @@ class SQLEditor(LumenEditor):
     export_formats = ("sql", "csv", "xlsx", "json", "markdown")
     _label = "Table"
 
+    # Icon shown in the "Add Filter" menu for each schema type.
+    _filter_icons = {
+        "number": "calculate",
+        "integer": "numbers",
+        "boolean": "toggle_on",
+    }
```
ahuang11, line 519: No support for strings?

ahuang11:
Also, I am wondering, if I interact with the sliders to subset the dataset, would/should subsequent queries see what you subset, or would it see the entire dataset?

lumen/ai/editors.py
ahuang11: It's:
```

/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/bokeh/util/serialization.py:247: UserWarning: no explicit representation of timezones available for np.datetime64
  return convert(array.astype("datetime64[us]"))
/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/bokeh/util/serialization.py:247: UserWarning: no explicit representation of timezones available for np.datetime64
  return convert(array.astype("datetime64[us]"))
Traceback (most recent call last):
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/panel/io/state.py", line 741, in execute
    callback()
  File "/Users/ahuang/repos/lumen/lumen/ai/editors.py", line 609, in _add_filter
    filt.widget.param.update(**widget_opts)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 2788, in update
    restore = dict(self_._update(arg, **kwargs))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 2814, in _update
    setattr(self_or_cls, k, v)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 629, in _f
    instance_param.__set__(obj, val)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 631, in _f
    return f(self, obj, val)
           ^^^^^^^^^^^^^^^^^
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameters.py", line 541, in __set__
    super().__set__(obj,val)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 631, in _f
    return f(self, obj, val)
           ^^^^^^^^^^^^^^^^^
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameterized.py", line 1815, in __set__
    self._validate(val)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameters.py", line 816, in _validate
    self._validate_value(val, self.allow_None)
  File "/Users/ahuang/miniconda3/envs/lumen/lib/python3.12/site-packages/param/parameters.py", line 844, in _validate_value
    raise ValueError(
ValueError: Integer parameter 'MultiChoice.size' must be an integer, not <class 'str'>.
```

ghostiee-11:
https://github.com/user-attachments/assets/94ce7c77-d796-4375-9815-48655ba98bc2

lumen/ai/editors.py
ghostiee-11: Fixed it Andrew

ghostiee-11:
> Also, I am wondering, if I interact with the sliders to subset the dataset, would/should subsequent queries see what you subset, or would it see the entire dataset?

Right now the sliders only subset the table in this view (the pipeline), so a follow-up query still reads the full source and doesn't carry the subset over. I leaned that way to keep filtering a quick visual thing without quietly changing what the LLM sees, but feeding the active filters into the next query's context is reasonable too.

ghostiee-11:
https://github.com/user-attachments/assets/84f568c4-ecae-4d59-a8d5-7ab935c58558

ahuang11:
Great it looks awesome!

Funny thing is the plot is not in order from earliest to latest (2000 -> 2012 -> 2008 -> 1996 -> 2004), but unrelated to this PR.

>  if I interact with the sliders to subset the dataset, would/should subsequent queries see what you subset, or would it see the entire dataset?

I'm slightly torn between having the LLM generate an equivalent filter expression with SQL, or we just programmatically re-apply the same filters. Thoughts?

ghostiee-11:
Yeah, I guess we should keep the LLM generating the WHERE. Follow-ups aggregate/group/join, so the filter has to land in the right spot and the LLM just does that, plus it stays visible in the SQL. Re-applying it programmatically gets messy the second the filtered column doesn't survive an aggregation.

catch is that the model might drop one now and then, but it's visible in the SQL so it's not silent. If that gets flaky we can add a guard that checks the filters made it into the query and retries, rather than maintaining a separate re-apply path.

ghostiee-11:
> Funny thing is the plot is not in order from earliest to latest (2000 -> 2012 -> 2008 -> 1996 -> 2004), but unrelated to this PR.

Yeahh 😅 Will raise and issue about this and fix this in seperate PR

ahuang11 (approved):

PR #1896, opened 2026-06-23
Fix misleading limit

ahuang11:
Was testing a 1M+ row dataset, and the LLM cited it only had 1M rows, so added a disclaimer about a row limit.
image

Removed truncation in log_debug for better visibility into these things.

### DIFF
--- a/lumen/ai/editors.py
+++ b/lumen/ai/editors.py
@@ -534,12 +534,17 @@ class SQLEditor(LumenEditor):
         )
 
     async def render_context(self):
+        row_limit = next(
+            (t.limit for t in self.component.sql_transforms
+             if isinstance(t, SQLLimit) and t.limit),
+            None
+        )
         return {
             "sql": self.spec,
             "pipeline": self.component,
             "table": self.component.table,
             "source": self.component.source,
-            "data": await describe_data(self.component.data)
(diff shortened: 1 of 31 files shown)


lumen/ai/editors.py
```diff
@@ -534,12 +534,17 @@ def render_explorer(self):
...
+             if isinstance(t, SQLLimit) and t.limit),
+            None
+        )
         return {
             "sql": self.spec,
             "pipeline": self.component,
             "table": self.component.table,
             "source": self.component.source,
-            "data": await describe_data(self.component.data)
+            "data": await describe_data(self.component.data, row_limit=row_limit)
```
ghostiee-11, line 547: nice fix. same thing shows up in a couple other spots: `sql.py:434` and `dbtsl.py:209` both run `describe_data` on a pipeline that already has `SQLLimit(1_000_000)` applied but don't pass `row_limit`, so they would still miss the disclaimer. maybe pull the row_limit lookup into a small helper and reuse it across these?

philippjfr: Those other instances aren't actually used for context. I'll eliminate them in a separate PR.

ghostiee-11: Okay

PR #1897, opened 2026-06-24
Structure prompts

ahuang11:
Builds on https://github.com/holoviz/lumen/pull/1896

Combed through the prompts with Opus 4.8:

This PR is a consistency and quality pass over the `lumen/ai/prompts` templates (limit/row-count handling is excluded and lives on a separate branch). It standardizes the heading hierarchy across all 38 templates—`##` for top-level sections, `###` for subsections, with no stray single-`#` headers or trailing-colon headings—and adds "illustrative" markers to example blocks so the model doesn't mistake sample tables, metrics, and numbers for the user's live data.

It fixes a handful of small but real issues (a `can will` grammar bug in DependencyResolver, a SQLAgent dash typo, a stray quote in DbtslAgent, a Jinja precedence bug in AnalysisAgent's docstring rendering, and contradictory font sizes between prose and example in VegaLite's `revise_output`), tightens a few guardrails (SourceAgent "at most one tool," FunctionTool "don't fabricate missing args," DocumentSummarizer "only cite filenames present in context"), and harmonizes duplicated wording such as the Snowflake upper-case note.

Finally, it documents the conventions in a new `GUIDANCE.md` and enforces the load-bearing ones with a new `tests/ai/test_prompts.py` that parses every template and checks the heading rules, so future prompts stay consistent.

### DIFF
--- a/lumen/ai/prompts/GUIDANCE.md
+++ b/lumen/ai/prompts/GUIDANCE.md
@@ -1,0 +1,100 @@
+# Prompt authoring guidance
+
+Design notes and conventions for the Jinja2 prompt templates in this directory.
+The goal is one consistent house style grounded in current prompt/context-engineering
+practice. When in doubt, prefer the smallest, clearest prompt that reliably produces
+the behavior you want.
+
+## Core principles
+
+1. **Aim for the right altitude.** Be specific enough to steer behavior, flexible
+   enough to leave the model good heuristics. Avoid both extremes: brittle if-else
+   logic hardcoded into prose, and vague hand-waving that assumes shared context.
+   If a rule encodes a one-off workaround, it probably belongs in code, not the prompt.
+
+2. **Minimal high-signal context, not maximal.** The model has a finite attention
+   budget and recall degrades as the context grows ("context rot"). Minimal does not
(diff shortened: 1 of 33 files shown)


ghostiee-11:
The description mentions a new `tests/ai/test_prompts.py` but it's not in the diff

lumen/ai/prompts/GUIDANCE.md
```diff
@@ -0,0 +1,100 @@
...
+
+6. **Iterate from a minimal baseline against real failures.** Start minimal, observe
+   actual failure modes, then add the smallest instruction or example that fixes the
+   observed failure — not a speculative rule for a failure you haven't seen.
+
+## Repo conventions
+
+### Structure & headings
+
+- Use `##` for top-level sections and `###` for subsections. Do **not** use single `#`
```
ghostiee-11, line 42: by this rule, `VectorStore/main.jinja2` still has single `#` headers that slipped through (only `should_situate.jinja2` was touched here). These can be done
```
## Full Document
## Previous Chunk Context
## Metadata
```

philippjfr (approved):

PR #1902, opened 2026-07-05
Detect encoding when reading uploaded CSV/JSON/HTML

### LINKED ISSUE #1901: AI file ingest assumes UTF-8 and fails on non-UTF-8 CSVs (ignores detect_file_encoding)
Loading a non-UTF-8 CSV through Lumen AI fails to parse. For example a Latin-1 encoded file such as the US Census county estimates (`co-est2024-alldata.csv`, which contains bytes like `0xf1` for names like "Dona Ana"):

```
Exception: Could not parse 'co-est2024-alldata.csv':
'utf-8' codec can't decode byte 0xf1 in position 6703: invalid continuation byte
```

The failure is wrapped and retried, so it surfaces as `RetriesExceededError` and an apparently hung spinner rather than a clear message.

#### Cause

`lumen/ai/controls/ingest/utils.py` hardcodes UTF-8:

- line 158: `content = content.decode("utf-8")`
- line 192: `content = content.decode("utf-8")`
- line 263: `df = pd.read_csv(file_obj, **csv_kwargs)`  (no `encoding`)

Lumen already has `lumen.util.detect_file_encoding`, used correctly in `lumen/ai/controls/ingest/file.py:91`, `lumen/transforms/sql.py:169`, and `lumen/sources/duckdb.py:229`, but not in `utils.py`.

#### Suggested fix

Use `detect_file_encoding` in `ingest/utils.py` (falling back to latin-1) as the sibling modules do, and avoid retry-looping a hard decode failure so the user gets a clear error.

ghostiee-11:
## Description

Fixes #1901

The AI ingest readers hardcoded UTF-8, so a non-UTF-8 upload (e.g. a latin-1 CSV) failed to parse and surfaced as a retry loop. Use the existing `detect_file_encoding` util in the CSV/JSON/HTML readers, as the sibling ingest modules already do.

### DIFF
--- a/lumen/ai/controls/ingest/utils.py
+++ b/lumen/ai/controls/ingest/utils.py
@@ -13,7 +13,7 @@ from urllib.parse import parse_qs, urlparse
 import httpx
 import pandas as pd
 
-from ....util import normalize_table_name
+from ....util import detect_file_encoding, normalize_table_name
 from .constants import (
     CONTENT_TYPE_TO_EXTENSION, METADATA_EXTENSIONS, TABLE_EXTENSIONS,
     DownloadConfig,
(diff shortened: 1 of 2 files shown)


ahuang11 (approved):

ahuang11:
Thanks!

Issue #1809, opened 2026-04-13
Support for Geopandas


ahuang11:
Not sure if it already works, but just wanted to label it here to try

ghostiee-11:
checked on current main, it doesn't work yet. querying a geometry table crashes (`source.get`/`execute` both raise `NotImplementedException: Unsupported type "GEOMETRY"` on the DuckDB -> numpy conversion), and `get_dataframe_schema` mis-detects the geometry column as a string enum, dumping raw shapely objects into the schema.

opened #1903 to fix it end to end: geometry-safe fetch (GeoDataFrame out), a proper geometry schema entry (type/crs/geographic), and per-view rendering (table -> wkt, deckgl -> geojson, hvplot -> polygons/paths/points). upload a geojson and ask for a choropleth and it renders a map.

ghostiee-11:
Ig, This can be closed @ahuang11

PR #1903, opened 2026-07-05
Support geometry columns end-to-end so GeoDataFrames render as maps

ghostiee-11:
## Overview

Addresses #1809. Geometry columns were invisible to the schema, so a GeoDataFrame never routed to a map. This recognizes geometry in the schema, preserves the GeoDataFrame through the source and pipeline, and lets the Table, hvPlot and DeckGL views plus the map agents render it, so uploading a GeoJSON and asking for a choropleth now returns a map.

## Demo

https://github.com/user-attachments/assets/7a545186-cc57-4d18-91c1-c1877d2e10eb

https://github.com/user-attachments/assets/7a5d6331-3ed3-4554-b3d3-446d7a322e7e


lumen/sources/duckdb.py
```diff
@@ -500,11 +507,40 @@ def create_sql_expr_source(
...
+            return rel.fetch_df(date_as_object=date_as_object)
+        selected = ', '.join(
+            f'ST_AsWKB("{d[0]}"::GEOMETRY) AS "{d[0]}"'
+            if d[0] in geom_cols else f'"{d[0]}"'
+            for d in rel.description
+        )
+        wrapped = f'SELECT {selected} FROM ({sql_expr})'
+        rel = cursor.execute(wrapped, params) if params else cursor.execute(wrapped)
+        df = rel.fetch_df(date_as_object=date_as_object)
+        if check_geopandas_available():
```
ahuang11: I think this would be nicer

```python
if gpd := try_import_geopandas():
   ...
```

lumen/sources/duckdb.py
```diff
@@ -500,11 +507,40 @@ def create_sql_expr_source(
...
+
+        DuckDB cannot convert a native GEOMETRY column to NumPy, so re-select
+        any geometry columns as WKB via ST_AsWKB before fetching, then rebuild
+        a GeoDataFrame when geopandas is available (WKB bytes otherwise).
+        """
+        rel = cursor.execute(sql_expr, params) if params else cursor.execute(sql_expr)
+        geom_cols = [d[0] for d in rel.description if str(d[1]) == 'GEOMETRY']
+        if not geom_cols:
+            return rel.fetch_df(date_as_object=date_as_object)
+        selected = ', '.join(
```
ahuang11, line 525: Nit for readability (add extra line spacing so it's detectable that return early) 
```suggestion
            return rel.fetch_df(date_as_object=date_as_object)

        selected = ', '.join(
```

lumen/views/base.py
```diff
@@ -962,6 +964,22 @@ def __init__(self, **params):
             params["kind"] = "points"
         super().__init__(**params)
 
+    @staticmethod
+    def _is_geodata(df) -> bool:
+        """Return True if df is a GeoDataFrame with an active geometry column."""
+        return is_geodataframe(df) and df.geometry.name in df.columns
```
ahuang11: Should this be consolidated with is_geodataframe?

lumen/views/base.py
```diff
@@ -962,6 +964,22 @@ def __init__(self, **params):
...
+    @staticmethod
+    def _geometry_kind(df) -> str:
+        """Pick an hvplot kind from the geometry type of a GeoDataFrame."""
+        geom_types = df.geometry.geom_type.dropna().unique()
+        geom_type = geom_types[0] if len(geom_types) else ''
+        if 'Polygon' in geom_type:
+            return 'polygons'
+        if 'Line' in geom_type:
+            return 'paths'
+        return 'points'
```
ahuang11: I wonder if there's a better way to do this 🤔 or if hvplot already has a method to do this

lumen/util.py
```diff
@@ -501,3 +518,52 @@ def check_xarray_available():
...
         return False
+
+
+def try_import_geopandas():
+    """Import and return geopandas, or None if it is not installed."""
+    try:
+        import geopandas
+        return geopandas
+    except ImportError:
+        return None
```
ahuang11, line 529: Nice! Can we also do this for check_xarray in a separate PR?

lumen/views/base.py
ghostiee-11: done, dropped it and just using `is_geodataframe`. the extra `geometry.name` in columns check was redundant, would raise instead of returning false anyway.

lumen/views/base.py
ghostiee-11: hvplot infers the kind for the plain view so i dropped the mapping there. the explorer defaults to scatter and doesn't infer.

lumen/util.py
```diff
@@ -501,3 +518,52 @@ def check_xarray_available():
...
+
+
+def check_geopandas_available():
+    """Check if geopandas is installed."""
+    return try_import_geopandas() is not None
+
+
+def is_geodataframe(df):
+    """Return True if df is a geopandas GeoDataFrame."""
+    gpd = try_import_geopandas()
```
philippjfr: Since this is called speculatively, i.e. even when it isn't needed, we should not attempt to import geopandas here. Either we check if the dataframe has a geometry column or we check if `"geopandas" in sys.modules`.

philippjfr (commented):
Looks great but let's please make sure that we don't speculatively import geopandas if it's not needed.

lumen/util.py
ghostiee-11: Hmm, Makes sense, Will change this to `gpd = sys.modules["geopandas"]`

lumen/views/base.py
ahuang11: Is there a way to have an LLM generate this instead? basically a prompt if geodataframe

lumen/util.py
```diff
@@ -501,3 +518,58 @@ def check_xarray_available():
...
+    try:
+        import geopandas
+        return geopandas
+    except ImportError:
+        return None
+
+
+def check_geopandas_available():
+    """Check if geopandas is installed."""
+    return try_import_geopandas() is not None
```
ahuang11: Can we remove this function

ahuang11 (changes requested):

lumen/util.py
```diff
@@ -501,3 +518,58 @@ def try_import_xarray():
         return xarray
     except ImportError:
         return None
+
+
+def check_xarray_available():
+    """Check if xarray and xarray-sql are installed."""
+    return try_import_xarray() is not None
```
ahuang11: This was accidentally reintroduced back.

ghostiee-11: Yupp might be in fixing conflicts...

lumen/tests/views/test_base.py
```diff
@@ -364,3 +364,118 @@ def test_vega_datasets(set_root):
...
+
+
+def test_deckgl_view_geometry_as_geojson():
+    """DeckGLView emits a GeoDataFrame as a GeoJSON FeatureCollection for layers."""
+    gpd = pytest.importorskip("geopandas")
+    pytest.importorskip("duckdb")
+    from shapely.geometry import Polygon
+
+    from lumen.sources.duckdb import DuckDBSource
+    from lumen.views.base import DeckGLView
```
ahuang11: I notice imports are nested in these tests

lumen/tests/ai/test_agents.py
```diff
@@ -277,6 +277,15 @@ class ExtendedChatAgent(ChatAgent):
         assert "Footer appended." in prompt
 
 
+def test_map_agents_route_geometry_columns():
+    """hvPlot and DeckGL agents advertise a geometry-column condition so the
+    coordinator routes GeoDataFrame data to a map-capable view."""
+    from lumen.ai.agents.deck_gl import DeckGLAgent
+    from lumen.ai.agents.hvplot import hvPlotAgent
```
ahuang11: I still see some nested imports; try `git diff main...HEAD > GEOPANDAS.diff` and pass that file to Claude to clean it all up.

lumen/tests/views/test_base.py
```diff
@@ -10,7 +13,28 @@
...
+
+requires_geo = pytest.mark.skipif(
+    not _GEO_DEPS, reason="geopandas, shapely or duckdb not installed"
+)
+
+# geoviews is checked without importing it: importing geoviews (-> matplotlib)
+# at collection under xdist can deadlock the font cache on macOS. hvplot imports
+# it lazily when the geometry plot is actually rendered.
+requires_geoviews = pytest.mark.skipif(
+    importlib.util.find_spec("geoviews") is None, reason="geoviews not installed"
```
ahuang11, line 37: duckdbsource is not part of geo deps
and also, doesn't gpd already require shapely? 
are we ableto simplify to
requires_geoviews = pytest.mark.skipif(
    importlib.util.find_spec("geopandas") is None, reason="geopandas not installed"
)

lumen/tests/test_pipeline.py
```diff
@@ -15,6 +15,13 @@
 except Exception:
     duckdb = None
 
+try:
+    import geopandas as gpd
+
+    from shapely.geometry import Polygon
+except Exception:
+    gpd = None
```
ahuang11, line 23: We can create a test util under conftest or something

lumen/util.py
```diff
@@ -501,3 +518,53 @@ def try_import_xarray():
...
+
+
+def is_geodataframe(df):
+    """Return True if df is a geopandas GeoDataFrame.
+
+    Uses an already-imported geopandas rather than importing it: a df cannot be
+    a GeoDataFrame unless geopandas is loaded, so this stays cheap on the hot
+    path and never speculatively imports geopandas.
+    """
+    gpd = sys.modules.get("geopandas")
```
ahuang11, line 539: We should add a kwarg to `try_import_geopandas(load=True)` and if not load, `return sys.modules.get("geopandas")`

Do the same for xarray??

ahuang11 (approved):
Talked to Aman; planning to fix the comments downstream since other PRs also need fixing.

PR #1905, opened 2026-07-05
Load geospatial files (.geojson/.wkt/.shp) at startup

### LINKED ISSUE #1900: Lumen AI cannot load .geojson/.shp data at startup (_resolve_data)
Passing a geospatial file to Lumen AI at startup fails:

```
lumen-ai serve counties.geojson --provider openai
...
ValueError: Could not determine how to load /path/counties.geojson file.
```

`ExplorerUI(data=['counties.geojson'])` fails the same way.

#### Cause

`_resolve_data` in `lumen/ai/ui.py` only recognizes a fixed set of extensions before
raising:

```python
if src.endswith(('.parq', '.parquet', '.csv', '.json', '.tsv', '.jsonl', '.ndjson')):
    table = src
else:
    raise ValueError(f"Could not determine how to load {src} file.")
```

Geospatial extensions (`.geojson`, `.wkt`, zipped `.shp`) are not handled. The interactive upload path already supports them: `read_geo_file` in `lumen/ai/controls/ingest/utils.py` reads them with geopandas and registers them via DuckDB spatial (`ST_GeomFromWKB`). The startup path could route geo extensions through the same code so uploaded and startup data behave consistently.

#### Suggested fix

Add a geo branch in `_resolve_data` that dispatches geospatial extensions to the existing `read_geo_file` -> DuckDB path.

ghostiee-11:
## Overview

Fixes #1900. Passing a geospatial file at startup (`lumen-ai serve x.geojson`, or `ExplorerUI(data=[...])`) raised "Could not determine how to load", even though the interactive upload already reads it. This routes local `.geojson`/`.wkt`/zipped-shapefile through the same `read_geo_file` to DuckDB spatial path so startup and upload behave the same. Pairs with #1903, which makes the loaded geometry queryable/renderable.

## Before / After

Before:
```
$ lumen-ai serve counties.geojson --provider openai
...
ValueError: Could not determine how to load counties.geojson file.
```

After:
```
$ lumen-ai serve counties.geojson --provider openai
# loads a source with a `counties` table (geometry registered via ST_GeomFromWKB),
# available at startup just like a .csv / .nc / .db
```

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -37,7 +37,7 @@ from ..pipeline import Pipeline
 from ..sources import Source
 from ..sources.duckdb import DuckDBSource
 from ..sources.xarray_sql import XArraySQLSource
-from ..util import check_xarray_available, log
+from ..util import check_xarray_available, log, normalize_table_name
 from .agents import (
     AnalysisAgent, BaseCodeAgent, ChatAgent, DocumentListAgent,
     DocumentSummarizerAgent, SourceAgent, SQLAgent, TableListAgent,
@@ -53,6 +53,7 @@ from .controls import (
     SourceCatalog, TableExplorer, UploadSourceControls,
 )
 from .controls.ingest.constants import XARRAY_EXTENSIONS
+from .controls.ingest.utils import read_geo_file
 from .coordinator import Coordinator, Plan, Planner
 from .editors import AnalysisOutput, LumenEditor, SQLEditor
 from .export import export_notebook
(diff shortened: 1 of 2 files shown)


lumen/ai/ui.py
```diff
@@ -661,6 +668,45 @@ def _resolve_data(
...
+    @classmethod
+    def _resolve_geo_source(cls, src: str, extension: str) -> Source:
+        """Load a geospatial file into an in-memory DuckDBSource.
+
+        Routes GeoJSON/WKT/zipped-shapefile through the same read_geo_file ->
+        DuckDB spatial (ST_GeomFromWKB) path the interactive upload uses, so
+        startup and uploaded geospatial data behave identically (gh-1900).
+        """
+        from ..util import normalize_table_name
+        from .controls.ingest.utils import read_geo_file
```
ahuang11: Maybe move these up

ahuang11 (changes requested):
Teeny change

lumen/ai/ui.py
ghostiee-11: sure

ahuang11 (approved):

PR #1909, opened 2026-07-08
feat: add a reusable normalize_vegalite_spec function

ghostiee-11:
Lumen's Vega-Lite spec normalization is locked inside the private `VegaLiteAgent._extract_spec`, reachable only by spinning up a full agent. This PR lifts it into a standalone `normalize_vegalite_spec` function so specs can be normalized headlessly; `_extract_spec` keeps its parsing and now delegates to it, so existing behavior is unchanged. I need this for lumen-mcp, which renders host-written Vega-Lite specs without an agent and relies on this step to inject `$schema` (which `pn.pane.Vega` requires).

### DIFF
--- a/lumen/ai/agents/vega_lite.py
+++ b/lumen/ai/agents/vega_lite.py
@@ -30,6 +30,109 @@ from ..vector_store import DuckDBVectorStore
 from .base_code import BaseCodeAgent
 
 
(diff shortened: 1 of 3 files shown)


ahuang11:
Thanks! I think this should be under utils if so.

ahuang11 (approved):

PR #1917, opened 2026-07-08
Use hsplit for Navigation

ahuang11:
Ported styling to https://github.com/panel-extensions/panel-splitjs/pull/30 and made Navigation bar also SplitJS, so it can be resized, and doesn't need to be in the left bar, which was confusing.

https://github.com/user-attachments/assets/72c1cd4e-1af1-4075-9353-78eb0b135224


philippjfr:
I'd also like to explore using a [docked inline Drawer](https://panel-material-ui.holoviz.org/reference/layouts/Drawer.html#inline-drawers).

ahuang11:
https://github.com/user-attachments/assets/91a95f43-8bc1-4283-9720-bd9769dd3b8e

Looks clean!

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
     "pandas",
     "panel >=1.7.5",
     "panel-graphic-walker[kernel] >=0.6.4",
-    "panel-material-ui >=0.8.1",
+    "panel-material-ui >=0.14.0",
     "panel-splitjs >=0.3.2",
```
ghostiee-11: `collapse_threshold=15` in `_render_main` was only added in panel-splitjs 0.4.0 (0.3.4 raises `TypeError: unexpected keyword argument 'collapse_threshold'`). CI passes because it resolves to 0.4.0, but anyone on 0.3.2-0.3.4 will crash when the UI builds. Should this be `>=0.4.0`?

ghostiee-11 (changes requested):

pyproject.toml
ahuang11: Thanks for reviewing, yes!

ahuang11:
```suggestion
    "panel-splitjs >=0.4.0",
```

ghostiee-11 (approved):

PR #1918, opened 2026-07-09
Fix editor layout and filters

ahuang11:
Handles:
1. spaces in column names
2. Improved responsiveness
3. Reorder filter icon to near last
4. NaNs or None values
5. slider max < 1 while step == 1

Remaining issue: https://github.com/panel-extensions/panel-material-ui/issues/704

https://github.com/user-attachments/assets/06b74f8a-a85b-47b2-a82e-d96cfc49f14d

### DIFF
--- a/lumen/transforms/sql.py
+++ b/lumen/transforms/sql.py
@@ -519,10 +519,12 @@ class SQLFilterBase(SQLTransform):
         filters = []
 
         for col, val in conditions:
-            # Wrap the column name in a quoted Identifier so names containing
-            # spaces or other special characters (e.g. "Smoking Status") are
-            # emitted as valid, quoted identifiers rather than bare tokens.
-            column_expr = Column(this=Identifier(this=col, quoted=True))
+            # Wrap the column name in an Identifier. Only force quoting when
+            # the name actually needs it (e.g. contains spaces or other
+            # special characters); otherwise leave it unquoted so generated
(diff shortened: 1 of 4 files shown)


lumen/transforms/sql.py
```diff
@@ -519,7 +519,12 @@ def _build_filter_conditions(self, conditions: list) -> list:
         filters = []
 
         for col, val in conditions:
-            column_expr = Column(this=col)
+            # Wrap the column name in an Identifier. Only force quoting when
+            # the name actually needs it (e.g. contains spaces or other
+            # special characters); otherwise leave it unquoted so generated
+            # SQL matches plain column references like `A` or `obs_id`.
+            needs_quoting = not re.match(r'^[A-Za-z_][A-Za-z0-9_]*$', col)
```
ghostiee-11: Poked at this against duckdb, and most reserved words are actually fine here since sqlglot auto-quotes them even with `needs_quoting=False` (`order`/`select` come out as `"order"`/`"select"`, both parse). But a few slip through, e.g. a column literally named `end` emits unquoted and hits a `Parser Error`. Since always-quoting works for everything (`"end"`, `"obs_id"`, `"A"` all bind fine), might just be simpler to do `quoted=True` and drop the regex?

ghostiee-11 (changes requested):

ghostiee-11 (approved):

PR #1919, opened 2026-07-09
Fix describe data

ahuang11:
Rank which columns to show enums. I noticed wide dataframes wasted cols with many string values, but there were cols with limited enums that is very useful for filtering.

### DIFF
--- a/lumen/ai/utils.py
+++ b/lumen/ai/utils.py
@@ -888,9 +888,7 @@ def describe_data_sync(
         summary["columns_shown"] = len(selected_columns)
         notes.append(
             f"Showing {len(selected_columns)} of {n_cols} columns, chosen by "
-            "relevance (query/filter columns and low-cardinality fields first); "
-            "use the schema tools for the rest."
-        )
+            "relevance (query/filter columns and low-cardinality fields first).")
     if notes:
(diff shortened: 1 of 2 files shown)


lumen/ai/utils.py
```diff
@@ -600,11 +606,116 @@ def get_data_sync():
...
+
+    if nunique <= 1:
+        # Constant (or all-null) column carries no signal.
+        return 0.0
+
+    ratio = nunique / n_rows if n_rows else 0.0
+
+    # Low-cardinality categorical/enum (incl. bool and small int codes):
+    # cheap to summarise and usually the most query-relevant.
+    if nunique <= LOW_CARDINALITY_MAX and ratio < 0.5:
```
ghostiee-11: One edge: the `ratio < 0.5` gate drops a low-card categorical to 1.5 on small frames (8 distinct in 12 rows = 0.67), which is exactly the filter column you're trying to surface. Only bites on wide + few-row results, but since `nunique <= LOW_CARDINALITY_MAX` already caps it, maybe drop the ratio for that bucket?

ghostiee-11 (changes requested):

ghostiee-11 (approved):

PR #1922, opened 2026-07-09
Allow selecting which report sections to keep on export

### LINKED ISSUE #1911: Allow users to select charts to keep / discard in exported Report
Currently, it's not possible to subset the results upon download. Ideally, there'd be checkboxes next to the things to keep.

Image

ghostiee-11:
Adds a checkbox to each report section card so users can pick which sections to keep, and the Notebook/HTML export includes only the checked ones. Default keeps every section, so existing exports are unchanged, and selection affects the export only, not the live report. Closes #1911.

https://github.com/user-attachments/assets/c018a98e-707a-44e5-b1f9-716c011c9a72

### DIFF
--- a/lumen/ai/report.py
+++ b/lumen/ai/report.py
@@ -625,76 +625,6 @@ class TaskGroup(Task):
                     task.reset()
             self._populate_view()
 
-    @property
-    def _export_views(self):
-        """
-        The flat list of views to include when exporting. Overridden by
-        subclasses that support selecting which parts to export.
-        """
-        return self.views
-
-    def _export_view(self):
-        """
-        The renderable view to save when exporting to HTML. Overridden by
-        subclasses that support selecting which parts to export.
-        """
-        return self._view
-
-    def to_notebook(self):
-        """
-        Returns the notebook representation of the tasks.
-        """
-        if len(self) and not self.status == "success":
-            raise RuntimeError(
-                "Report has not been executed, run report before exporting to_notebook."
-            )
-        cells, extensions = [], ['tabulator']
-        pending_headers = []
-        for out in self._export_views:
-            cell = ext = None
-            if isinstance(out, Typography):
-                # Buffer headers; only emit them if followed by exportable content
-                level = int(out.variant[1:]) if out.variant and out.variant.startswith('h') else 0
-                prefix = f"{'#'*level} " if level else ''
-                pending_headers.append(make_md_cell(f"{prefix}{out.object}"))
-                continue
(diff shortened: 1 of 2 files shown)


ahuang11:
Thanks! A few thoughts:
1. is left / right side a better location for check box?
2. should we allow reordering?
3. maybe we can not keep the Report section as is, but append a story section and utilize https://panel.holoviz.org/reference/widgets/JSONEditor.html separately for the story portion... imagine you want multiple, nested headings, and reordering the tables based on what you want to present first. JSONEditor is the only widget I know that allows you to reorder

ghostiee-11:
Thanks for the review Andrew,

1. As a user, I would like to see the checkbox on the left as it is easily visible with the section properly 

2. reordering would be nice, but Accordion doesn't do it natively so it needs custom drag. it also overlaps with the dashboard export i'm adding where EditableTemplate already gives drag reorder. want it in the report view too, or is the dashboard enough?

3.  checked out JSONEditor and yeah it fits, tree mode lets you nest headings and reorder (drag + the sort/transform menu), and value is bidirectional so we can read the arranged structure back and render it. So the idea is a separate story section arranged in a JSONEditor, replacing the inline story ( As done in #1923)?

ghostiee-11:
I guess this pr is only about the checkbox selction thing, if everything clean can merge this @ahuang11, the #1923 PR where i addressed the reordering and JSON thing

lumen/ai/report.py
```diff
@@ -625,6 +625,21 @@ def reset(self, start: int = 0):
...
+        """
+        return self.views
+
+    def _export_view(self):
+        """
+        The renderable view to save when exporting to HTML. Overridden by
+        subclasses that support selecting which parts to export.
+        """
+        return self._view
+
```
ahuang11: Are these necessary? Seems like unnecessary wrappers?

ahuang11:
I guess if you have the JSONEditor, is the checkboxes still necessary because you can remove nodes

image

ghostiee-11:
Yes, using the JSON editor, we can remove the nodes. While making the video, I also considered this, and for the ease of the UI/UX, we should have checkboxes imo.

ahuang11:
Really close to merge! While testing, I noticed nested sections are not selectable.

image

ghostiee-11:
Screenshot 2026-07-17 at 9 35 39 PM

ghostiee-11:
@ahuang11, you can review it..

ahuang11 (approved):

PR #1925, opened 2026-07-10
Allow Analysis to provide context dynamically

ahuang11:
Analysis returns a view and its out-context is captured once, meaning interactive analyses (e.g. a manifold-map lasso) had no supported way to push updated `source`/`pipeline`/`data` to later steps.

This PR adds dynamic provides, where AnalysisOutput watches it and merges into render context, triggering context_updated so out_context is updated, without affecting spec.

Lastly, `_update_component` guards against a `None` spec.

### DIFF
--- a/lumen/ai/analysis.py
+++ b/lumen/ai/analysis.py
@@ -40,6 +40,14 @@ class Analysis(param.ParameterizedFunction):
 
     _run_button = param.Parameter(default=None)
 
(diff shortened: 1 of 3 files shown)


philippjfr:
I played with some other approaches, but in practice this is probably cleanest.

philippjfr (approved):

PR #1926, opened 2026-07-17
Add generic try_import helper and adopt it for optional deps

ghostiee-11:
Adds a single generic `try_import(module_name, load=True)` and adopts it across the optional-dependency helpers, so we do not grow a separate `try_import_<dep>` per PR. `load=False` returns the module only if it is already imported (a cheap hot-path check that never triggers an import); `load=True` imports on demand. The repeated geopandas test guard is also consolidated into a shared `lumen/tests/utils.py`.


lumen/tests/sources/test_optional_imports.py
ahuang11, line 141: Do we still need try_import_xarray?

ahuang11 (approved):

PR #1927, opened 2026-07-19
feat: generate multiple charts per query in VegaLiteAgent

### LINKED ISSUE #1916: Allow generating multiple charts at once in one query
Currently VegaLiteAgent only creates a single plot. Ideally, it can create multiple separate subplots and/or multiple separate tabs in one query.

ghostiee-11:
Fixes #1916.
`VegaLiteAgent` could only ever produce a single plot. It now returns one or more charts from a single request: each chart opens in its own editable tab (vega-lite code beside the plot), plus a view-only **"All"** tab that stacks every  plot when there are several. A chart whose spec fails to parse is skipped, so the others still render.

https://github.com/user-attachments/assets/284f170b-7854-44ff-ac84-4e5384af9929

### DIFF
--- a/lumen/ai/editors.py
+++ b/lumen/ai/editors.py
@@ -726,6 +726,22 @@ class SQLEditor(LumenEditor):
         return f"{self.__class__.__name__}:\n```sql\n{self.spec}\n```"
 
 
+class MultiChartEditor(LumenEditor):
+    """
+    Display-only editor for a composite of several charts (e.g. an "All" tab
+    that stacks every plot as an overview).
+
+    The composite is not a single editable spec, so no code editor is shown
+    (mirrors DocumentEditor); the charts still render through the base
+    LumenEditor view path.
+    """
+
+    _label = "Charts"
+
+    def _render_editor(self):
+        return None
+
+
 class DocumentEditor(LumenEditor):
     """
     Displays relevant documents in tabs with type-aware rendering.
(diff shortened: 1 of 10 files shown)


lumen/ai/editors.py
```diff
@@ -726,6 +726,22 @@ def __str__(self):
...
+    The composite is not a single editable spec, so no code editor is shown
+    (mirrors DocumentEditor); the charts still render through the base
+    LumenEditor view path.
+    """
+
+    _label = "Charts"
+
+    def _render_editor(self):
+        return None
+
```
ahuang11, line 775: I think we should support multi-spec; maybe as sub-tabs?

ahuang11:
Awesome! I think only thing is the spec should be editable.

lumen/ai/editors.py
ghostiee-11: Yupp...

ghostiee-11:
Screenshot 2026-07-22 at 4 04 24 PM

ahuang11:
Thanks! 
image

Do you know what's with the left margin?

also top margins

image

ghostiee-11:
left margin was doubled, the sub-tabs wrapper had its own (0,10) on top of each editor's margin. dropped it so it lines up with a normal tab now. the top one's just the MUI sub-tab header bar for the per-chart specs.

ahuang11 (approved):

PR #1929, opened 2026-07-20
Fix the view yaml auto-retry so it can actually recover

ghostiee-11:
The view yaml auto-retry could never recover. It assumed a `yaml_spec` field that hvPlot (its only user) does not emit, so it raised `KeyError` before `revise` was even reached; it also rebound `context` to a string, passed the spec positionally so it landed on `revise()`'s `view` parameter, discarded the revision instead of retrying with it, and then raised on a stale `error` flag even when a retry had succeeded.

It now reports the spec it actually has, keeps `context` intact, passes `spec=` by keyword, feeds the revision into the next attempt, and clears `error` on recovery. Added regression tests that fail without the change.

### DIFF
--- a/lumen/ai/agents/base_view.py
+++ b/lumen/ai/agents/base_view.py
@@ -113,6 +113,10 @@ class BaseViewAgent(BaseLumenAgent):
             for i in range(3):
                 try:
                     spec = await self._extract_spec(context, spec)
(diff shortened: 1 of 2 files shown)


ahuang11 (approved):
Thanks for the fix!

PR #1935, opened 2026-07-20
fix: build hvPlotAgent response model explicitly and add colormap options

### LINKED ISSUE #1931: hvPlotAgent cannot express any color options, so hvPlot's colormap defaults are unreachable
`hvPlotAgent` builds its response model from the declared params of `hvPlotUIView` via `param_to_pydantic` (`lumen/ai/agents/hvplot.py:51`). `hvPlotBaseView` declares only `kind`, `x`, `y`, `by`, `groupby` and `geo` (`lumen/views/base.py:925-970`), so the generated model has no color field at all. The agent cannot pick a colormap even when the user explicitly asks for one.

That matters more than it first looks, because hvPlot already does a lot here on its own:

- `_default_cmaps` (`hvplot/converter.py:840-845`) maps `linear` to `kbc_r`, `diverging` to `coolwarm`, `categorical` to `glasbey_category10` and `cyclic` to `colorwheel`
- `_process_symmetric` (`hvplot/converter.py:1302-1332`) auto-detects diverging data from the 5th/95th percentiles and then selects the diverging colormap (`converter.py:1227-1233`)
- the strings `linear`, `diverging`, `categorical` and `cyclic` are accepted as semantic cmap aliases (`hvplot/converter.py:1838-1839`)

None of that is reachable from Lumen today.

The one color-adjacent value Lumen does inject is `cnorm="log"` for large datasets (`lumen/ai/agents/hvplot.py:72-74`), which sets a log color norm without any colormap alongside it.

Declaring `cmap` plus a few related options would let the agent use the semantic aliases, which seems like a better fit for an LLM than asking it to pick from the full colormap list.

One thing worth being careful about if anyone picks this up: `symmetric`, `colorbar`, `cnorm`, `robust` and `rescale_discrete_levels` are all `None`-defaulted in hvPlot's converter so it can distinguish "unset" from "explicitly False", and each is forwarded only under `if x is not None`. Declaring `symmetric` as `param.Boolean(default=False)` would silently disable the auto-detection above, so these want `default=None` with `allow_None=True`.

ghostiee-11:
`hvPlotAgent` could not build its structured-output model at all: `param_to_pydantic` walks parent classes and every subclass transitively, so from `hvPlotUIView` it reached Panel's whole class graph and failed differently on each run. `_get_model` is called on every response, so the agent raised rather than degraded.

This builds the model with `create_model` instead, listing fields explicitly and deriving `kind` from the view's own Selector. Since the model was being rebuilt anyway it also adds `cmap`, `cnorm`, `colorbar` and `color`, with the colormap offered as linear/diverging/categorical/cyclic and resolved to a concrete name before it reaches the view.

Fixes #1931

### DIFF
--- a/lumen/ai/agents/hvplot.py
+++ b/lumen/ai/agents/hvplot.py
@@ -2,6 +2,7 @@ from typing import Any, Literal
 
 import param
 
+from hvplot.ui import CMAPS, DEFAULT_CMAPS, Colormapping
 from pydantic import BaseModel, Field, create_model
 
 from ...views import hvPlotUIView
@@ -10,19 +11,21 @@ from ..context import TContext
 from ..utils import get_data
 from .base_view import BaseViewAgent
 
+# Taken from hvPlot rather than restated, so the options cannot drift apart.
+CMAP_KINDS = tuple(DEFAULT_CMAPS)
+CNORMS = tuple(Colormapping.param["cnorm"].objects)
+
 
 def resolve_cmap(cmap: str) -> str:
     """
     Map a semantic colormap name onto a concrete one.
 
-    hvPlot understands 'linear', 'diverging', 'categorical' and 'cyclic' in its
-    converter, but `hvplot.ui.Colormapping.cmap` is a Selector over concrete
-    colormap names and rejects them, so they have to be resolved before the
-    spec reaches the view. The categorical default resolves to a list of
-    colours, which the Selector also rejects, hence the named fallback.
+    hvPlot understands these names in its converter, but
+    `hvplot.ui.Colormapping.cmap` is a Selector over concrete colormap names
+    and rejects them, so they have to be resolved before the spec reaches the
+    view. The categorical default resolves to a list of colors, which the
+    Selector also rejects, hence the named fallback.
     """
-    from hvplot.ui import CMAPS, DEFAULT_CMAPS
-
     resolved = DEFAULT_CMAPS.get(cmap, cmap)
     if not isinstance(resolved, str) or resolved not in CMAPS:
         return "glasbey"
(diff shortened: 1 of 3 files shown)


lumen/ai/agents/hvplot.py
```diff
@@ -1,18 +1,122 @@
...
+    """
+    Map a semantic colormap name onto a concrete one.
+
+    hvPlot understands 'linear', 'diverging', 'categorical' and 'cyclic' in its
+    converter, but `hvplot.ui.Colormapping.cmap` is a Selector over concrete
+    colormap names and rejects them, so they have to be resolved before the
+    spec reaches the view. The categorical default resolves to a list of
+    colours, which the Selector also rejects, hence the named fallback.
+    """
+    from hvplot.ui import CMAPS, DEFAULT_CMAPS
```
ahuang11: Can you make sure your PRs have imports up top unless they are intended to be imported within the function? (I don't think this is the case since 
from ...views import hvPlotUIView is imported up top)

lumen/ai/agents/hvplot.py
```diff
@@ -1,18 +1,122 @@
...
+        x=(
+            column | None,
+            Field(default=None, description="Column to plot on the x-axis."),
+        ),
+        y=(
+            column | None,
+            Field(default=None, description="Column to plot on the y-axis."),
+        ),
+        __base__=hvPlotSpec,
+    )
```
ahuang11, line 120: Is there a better way? Rather than duplicating the args, can we read the signature from https://github.com/holoviz/hvplot/blob/451910e7a17ba9e9ecd4115b30d529cc2cb9dfbb/hvplot/converter.py#L866

It's missing typing at this moment, but maybe could add typing?

ahuang11:
Do you have a solution for https://github.com/holoviz/lumen/pull/1935#discussion_r3626209598

Issue #1932, opened 2026-07-20
Generated plots are not theme aware and have no shared color defaults


ghostiee-11:
Two related gaps I keep running into. Raising them together because the fix for both probably wants the same hook, and I would rather agree on placement before writing anything.

### Plots ignore the theme

The Vega spec never changes with the Lumen theme. Theming stops at the Panel template layer (`lumen/config.py:128-129`) and never reaches plots.

Meanwhile the VegaLiteAgent prompts hardcode light-mode values that get baked into every generated spec, for example `#eeeeee` gridlines and `#666666` subtitle text (`lumen/ai/prompts/VegaLiteAgent/main.jinja2:93`, `:138`), plus assorted hex colors at `:33`, `:152` and `:174`. In dark mode these render close to invisible.

`pn.pane.Vega` does have a `theme` parameter, but Lumen never sets it, and it would lose to a `config` block the model wrote itself.

### No shared color defaults

Every chart picks its own colors, so two views of the same data in one report can end up on different palettes. There is no colorblind-safe default, and high cardinality categoricals run off the end of `category10` and start recycling colors.

There is a natural source for this already in the dependency tree. `holoviews.plotting.util` exposes `list_cmaps` and `process_cmap` over a curated registry spanning matplotlib, bokeh and colorcet, with a `category` taxonomy that marks perceptually uniform maps and a `bg` field recording whether a map suits a light or dark background. `process_cmap(name, ncolors=n)` returns a plain hex list, which is the shape Vega's `config.range` wants.

### Questions before writing anything

1. Should defaults be merged into the spec, so the user sees them in the editor and they survive export, or applied at render time in `VegaLiteView._get_params`, giving a cleaner spec and letting a theme switch re-render without rewriting spec text?
2. Is a `config` merge inside `normalize_vegalite_spec` acceptable, or is that too much post processing?
3. Should the prompts stop emitting hardcoded hex values, given that an explicit `config` from the model overrides anything injected later?

Worth noting the hvPlot path already solves part of this: `_process_symmetric` picks a diverging colormap automatically, and `_default_cmaps` gives it sensible perceptual defaults. The Vega path has nothing equivalent.

ghostiee-11:
I dug into this and two things make the obvious approach not work, so recording them here before anyone spends time on it.

**Injecting a config is not enough on its own.** Vega-Lite treats `config` as the lowest priority defaults layer, so a palette only applies where the model left colour unspecified:

```
[...]

PR #1936, opened 2026-07-20
feat: give generated charts a shared categorical palette

ghostiee-11:
Charts picked their own colors, so two views of the same data in one report could disagree, and anything past ten categories fell off the end of Vega's `category10` default and started repeating.

This supplies a glasbey palette seeded on category10 as a `config` default the model can still override. Its first ten entries are visually identical to category10 (largest total RGB difference is 3 out of 765), so charts with ten categories or fewer render unchanged and only the repeating case differs. It is added only where a chart colors by category, and only on the way in, so annotating or polishing does not restore a palette someone removed.

The first commit is a separate fix: `_polish_plot` used `_extract_spec` purely as a validator and wrote back the un-normalized spec, dropping the projection, basemap layer and zoom params when a polish step introduced a lat/long encoding.

Refs #1932

### DIFF
--- a/lumen/ai/agents/vega_lite.py
+++ b/lumen/ai/agents/vega_lite.py
@@ -581,13 +581,13 @@ class VegaLiteAgent(BaseCodeAgent):
                     step_name, step_desc, out.spec, step_name, messages, context, doc=doc, out=out
                 )
                 try:
-                    # Validate merged spec
+                    # Validate merged spec, and keep what normalization added to it
                     merged_spec = self._deep_merge_dicts(out._spec_dict["spec"], update_dict)
-                    await self._extract_spec(context, {"yaml_spec": dump_yaml(merged_spec)})
+                    normalized = await self._extract_spec(context, {"yaml_spec": dump_yaml(merged_spec)})
                 except Exception as e:
                     log_debug(f"Skipping invalid {step_name} update due to error: {e}")
                     continue
-                out.spec = dump_yaml(merged_spec)
+                out.spec = dump_yaml(normalized["spec"])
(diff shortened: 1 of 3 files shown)


ghostiee-11:
Screenshot 2026-07-20 at 5 26 31 PM

Screenshot 2026-07-20 at 4 49 08 PM

lumen/ai/agents/vega_lite.py
```diff
@@ -93,6 +94,38 @@ class AltairSpec(BaseModel):
...
+    walk rather than a single lookup.
+    """
+    if isinstance(spec, dict):
+        color = spec.get("encoding", {}).get("color")
+        if isinstance(color, dict) and "field" in color and color.get("type") in ("nominal", "ordinal"):
+            return True
+        return any(has_categorical_color(value) for value in spec.values())
+    if isinstance(spec, list):
+        return any(has_categorical_color(item) for item in spec)
+    return False
```
ahuang11: Should this only be part of vega-lite or should it be part of hvplot too? i.e. utils?

ghostiee-11: moved both to `ai/utils` next to `normalize_vegalite_spec` so it's not agent-local.

didn't wire it into `hvPlotAgent` here because hvplot already picks `glasbey_category10` itself when it chooses a cmap, so the two line up. its by overlays are a separate thing though, those still use the 12 colour Cycle and recycle past 12. can do that in #1935 if you want it.

ahuang11 (approved):

ahuang11:
Awesome thanks!

PR #1959, opened 2026-07-24
Fix VegaLite specs without schemas

### LINKED ISSUE #1941: VegaLiteView fails on any spec without $schema
`VegaLiteView.get_panel` hands the spec straight to `pn.pane.Vega` (`lumen/views/base.py:1335-1337`). Panel only recognises a plain dict as a Vega spec when it carries a schema URL: `Vega.applies` returns `True` for a dict only if `'vega' in obj.get('$schema', '').lower()` (`panel/pane/vega.py:276`). A valid Vega-Lite spec without that key therefore fails with an error that says nothing about the missing key:

```python
from lumen.views.base import View

spec = {
    "mark": "bar",
    "encoding": {
        "x": {"field": "cat", "type": "nominal"},
        "y": {"field": "y", "type": "quantitative"},
    },
}
view = View.from_spec({"type": "vegalite", "spec": spec}, source=src, pipeline=pipeline)
view.get_panel()
# ValueError: Vega pane does not support objects of type 'dict'.
```

Adding `"$schema": "https://vega.github.io/schema/vega-lite/v5.json"` to the same spec renders it. I checked the 17 Vega-Lite primitive marks plus `layer`, `hconcat`, `vconcat`, `facet`, `repeat`, transforms and selections: every one of them renders through `VegaLiteView` once the key is present, and every one fails with the message above without it.

The AI path never hits this, which is why it has gone unnoticed. `normalize_vegalite_spec` injects the schema before the view is built (`lumen/ai/utils.py:1669-1671`), and the agent prompt explicitly tells the model to leave it out (`lumen/ai/agents/vega_lite.py:51`). Someone writing a dashboard YAML by hand gets the raw panel error instead, and `$schema` is optional in Vega-Lite itself so there is no reason to expect it to be required.

`VegaLiteView` is a Vega-Lite view by definition, so it can default the key in `_get_params` with the same value `normalize_vegalite_spec` already uses, rather than making every caller repeat it.

Sanjays2402:
Closes #1941

`VegaLiteView` now supplies the same Vega-Lite v5 schema used by the existing normalization path when a dashboard spec omits the optional `$schema` field. Explicit caller-provided schemas remain unchanged.

Regression coverage verifies schema-less specs render through Panel; all 20 focused view tests pass with 3 optional-dependency skips.


lumen/views/base.py
```diff
@@ -1466,15 +1466,19 @@ class VegaLiteView(View):
     def _get_params(self) -> dict[str, Any]:
         df = self.get_data()
         spec_data = self.spec.get('data', {})
+        spec = {
+            '$schema': 'https://vega.github.io/schema/vega-lite/v5.json',
+            **self.spec,
+        }
```
philippjfr:
```suggestion
        spec = dict(self.spec)
        if "$schema" not in spec:
            spec["$schema"] = "https://vega.github.io/schema/vega-lite/v5.json"
```

ahuang11 (approved):

PR #1965, opened 2026-07-25
Better tool handling

ahuang11:
1. Small model may not provide title all the time, and it's unnecessary, so default a title if unprovided so it doesn't crash.

2. Some models also tries to call agents as tools; this prompts them what tools are available.

3. Handle:
```

instructor.v2.core.errors.InstructorRetryException: <failed_attempts>
<generation number="1">
<exception>
    No tool calls or function call found in response (mode: TOOLS)
</exception>
<completion>
```

Gracefully

### DIFF
--- a/lumen/ai/prompts/Planner/main.jinja2
+++ b/lumen/ai/prompts/Planner/main.jinja2
@@ -102,14 +102,11 @@ Tools the LLM can call directly inside its own generation loop (tool-calling API
 These are available to you right now, as the planner, and to any agent whose
 loop includes them.
 
-{% if llm_tools -%}
-The ONLY names you may emit as a function/tool call are:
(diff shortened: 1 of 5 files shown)


ahuang11:
Tests is actually from https://github.com/holoviz/lumen/pull/1964 so once that's merged, this should be fixed

lumen/ai/prompts/Planner/main.jinja2
```diff
@@ -102,6 +102,15 @@ Tools the LLM can call directly inside its own generation loop (tool-calling API
 These are available to you right now, as the planner, and to any agent whose
 loop includes them.
 
+{% if llm_tools -%}
+The ONLY names you may emit as a function/tool call are:
```
philippjfr: This seems quite duplicative. I'm surprised it improves anything since the tool list is already provided via the tool specs.

ghostiee-11 (approved):

Issue #1968, opened 2026-07-25
No extra.js file in assets, but it is referenced in 61 HTML files. And extra.css is in a non-standard location


Coderambling:
The location and (non-)existence of these files is quite confusing. Is this due to migration @ghostiee-11 and @ahuang11 ? Any guidance and/or context would be much  appreciated.

In https://github.com/holoviz-dev/lumen/tree/gh-pages/assets there is an extra.css file, but no extra.js file in that folder or anywhere else in the repository. 

According to [Zensical documentation ](https://zensical.org/docs/customization/?h=extra.js#additional-javascript) custom js should be in docs/javascripts as extra.js . And extra.css should be in docs/stylesheets as extra.css .

In holoviz-dev lumen in assets there is a minified .js file. Should that be referenced from / renamed to extra.js, so only one script line will be needed in the HTML files? That file, nor extra.js, are present anywhere in the holoviz/lume repository btw. Why?

https://github.com/holoviz-dev/lumen/blob/gh-pages/assets/javascripts/bundle.8ffeb9c9.min.js

A search for the string "extra.js" shows occurences in 61 HTML files, just preceding the /HTML closing tag.

https://github.com/search?q=repo%3Aholoviz-dev%2Flumen+extra.js&type=code

Line: <script src="assets/extra.js"></script>

Line is inserted in the file due to this line in zensical.toml:

https://github.com/holoviz/lumen/blob/562936a925cfab12db0a6792be86df7390aba359/zensical.toml#L10-L11

Example from homepage: https://github.com/holoviz-dev/lumen/blob/gh-pages/index.html

https://github.com/holoviz-dev/lumen/blob/43d206ee75d7f32ad27ed177edb65d3460c713ce/index.html#L3113-L3122

```
[...]

PR #1969, opened 2026-07-27
docs: remove stale extra.js reference

w3lld1:
## Summary

- I removed the stale `extra_javascript` entry from the Zensical configuration.
- Generated documentation pages no longer request the missing `assets/extra.js` file.

## Validation

- `uvx --from 'zensical<0.0.17' --with mkdocstrings-python --with mkdocs-gen-files --with mkdocs-literate-nav zensical build --clean`
- Python 3 asset-existence check for configured JavaScript files
- `git diff --check upstream/main...HEAD`
- Confirmed generated HTML contains no `assets/extra.js` references

Fixes #1968

PR #1971, opened 2026-07-27
Comment out extra_javascript line in zensical.toml

Coderambling:
Comment out the extra_javascript entry in zensical.toml, as it currently injects the line belw in all HTML files of doc site, while there is no extra.js currently in use.

```HTML
<script src="assets/extra.js"></script>
```

See #1968 and explanation in https://github.com/holoviz/lumen/issues/1968#issuecomment-5078031555

alternatives considered were:

- Commenting out the line.

- Changing the line to:

`extra_javascript = []
`

- Removing the line altogether.

Chose the first alternative.


ghostiee-11:
Superseded with #1969

PR #1984, opened 2026-07-29
Add kilo as a provider

ahuang11:
Adds https://kilo.ai as an LLM provider

`lumen-ai serve data/olympic* --provider kilo --model nvidia/nemotron-3-super-120b-a12b:free --log-level debug --logfire-tags kilo`

### DIFF
--- a/docs/configuration/llm_providers.md
+++ b/docs/configuration/llm_providers.md
@@ -23,7 +23,7 @@ To use a specific provider, pass the `--provider` flag:
 ``` bash
 lumen-ai serve penguins.csv --provider anthropic
 lumen-ai serve penguins.csv --provider google --model 'gemini-2.5-flash'
-lumen-ai serve penguins.csv --provider openrouter --model 'openai/gpt-4o-mini'
+lumen-ai serve penguins.csv --provider kilo --model 'openai/gpt-4o-mini'
 lumen-ai serve penguins.csv --provider ollama --model 'qwen3:32b'
 lumen-ai serve penguins.csv --provider mlx
 ```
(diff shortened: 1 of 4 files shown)


lumen/ai/llm.py
```diff
@@ -2979,3 +2980,40 @@ def models(self) -> set[str]:
...
+
+class Kilo(OpenAI):
+    """
+    An LLM implementation using the Kilo API.
+
+    Kilo provides an OpenAI-compatible endpoint that routes requests to
+    models from multiple providers. Kilo has a free tier, but for other models,
+    optionally set the ``Kilo_API_KEY`` environment variable or pass ``api_key``
+    directly. The provider is auto-detected when
+    ``Kilo_API_KEY`` is present, or can be selected explicitly with
```
ghostiee-11: `Kilo_API_KEY` should be `KILO_API_KEY`, and the key isn't optional: with neither set, `Kilo()._instantiate_client()` raises `OpenAIError: The api_key client option must be set`, and if `OPENAI_API_KEY` happens to be set the SDK sends that key to api.kilo.ai instead. `api_key = param.String(default="kilo")` like Ollama fixes both.

docs/configuration/llm_providers.md
```diff
@@ -23,7 +23,7 @@ To use a specific provider, pass the `--provider` flag:
 ``` bash
 lumen-ai serve penguins.csv --provider anthropic
 lumen-ai serve penguins.csv --provider google --model 'gemini-2.5-flash'
-lumen-ai serve penguins.csv --provider openrouter --model 'openai/gpt-4o-mini'
+lumen-ai serve penguins.csv --provider kilo --model 'openai/gpt-4o-mini'
```
ghostiee-11, line 26: This swapped the openrouter example for kilo instead of adding a line, so openrouter has no CLI example left.

lumen/ai/llm.py
```diff
@@ -2979,3 +2980,40 @@ def models(self) -> set[str]:
...
+    model_kwargs = param.Dict(default={
+        "default": {"model": "kilo-auto/free"},
+    })
+
+    select_models = param.List(default=[
+        "kilo-auto/free",
+        "kilo-auto/balanced",
+        "kilo-auto/efficient",
+        "kilo-auto/frontier",
+    ], constant=True, doc="Available Kilo models for selection dropdowns.")
```
ghostiee-11, line 3019: No `models()` override, so it inherits `OpenAI.models()` and queries api.openai.com. `{endpoint}/models` returns 200 with the same `data[].id` shape, so OpenRouter's copies over as is. Low priority since nothing in lumen calls `models()`.

docs/configuration/llm_providers.md
ahuang11: This is just a preview of the providers; doesn't list all.

lumen/ai/llm.py
ahuang11: Don't think api_key is required:
```
import os

os.environ.pop("KILO_API_KEY")

from lumen.ai.llm import Kilo

await Kilo().invoke([{"role": "user", "content": "Hello"}])
```
works

ghostiee-11 (approved):

PR #1986, opened 2026-07-30
Clean up prompts and reduce token usage

ahuang11:
Make prompts more consistent and reduce token usage


ghostiee-11:
Its great... LGTM

ghostiee-11 (approved):

PR #2001, opened 2026-08-06
fix: blank choropleth when the lookup names an unregistered dataset

ghostiee-11:
A choropleth joins the table onto map outlines, so the table travels as a named dataset and the `lookup` transform has to name it exactly; nothing validates that name, so a spec naming a dataset that was never registered passes every server-side check and then renders a blank map with no error anywhere. `VegaLiteView` now repoints a dangling lookup at the one dataset it actually registered, and the choropleth example in the prompt shows the real table name instead of a `<TABLE_NAME>` placeholder.

Also copies the spec's `datasets` dict before writing to it, which was leaving a DataFrame on the caller's spec on every render.

### DIFF
--- a/lumen/ai/agents/deck_gl.py
+++ b/lumen/ai/agents/deck_gl.py
@@ -98,7 +98,7 @@ class DeckGLAgent(BaseCodeAgent):
             "Use for large-scale geospatial data with latitude/longitude coordinates",
             "Use for hexbin aggregations, heatmaps, or 3D extruded visualizations on maps",
             "Can render a geometry column (GeoDataFrame polygons/lines/points) on a basemap, suited to 3D/extruded maps or geometry whose schema is 'geographic' (a lat/lon CRS) so the basemap gives real-world context",
-            "NOT when the table only names places (a state or country column) without geometry or lat/lon, since deck.gl has nothing to draw and cannot fetch boundaries; that choropleth belongs to VegaLiteAgent",
+            "NOT when the table only names places (a state or country column) without geometry or lat/lon, since deck.gl has nothing to draw and cannot fetch boundaries",
         ]
     )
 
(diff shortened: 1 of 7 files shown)


lumen/ai/prompts/VegaLiteAgent/main.jinja2
```diff
@@ -261,30 +264,51 @@ encoding:
...
+rather than the table; this is the one chart where `data: {name: ...}` is wrong, because a
+table has no geometry and the map would render empty. The table is named inside
+`transform.lookup.from.data`, and the join reads the boundaries first, so `lookup` is the
+field on the boundaries and `key` is the matching column in the table.
+
+Draw the boundaries twice: an unfilled base layer so regions the table does not cover
+still appear, then the joined layer on top. Without the base layer a region with no
+matching row is dropped rather than drawn, and a table covering 15 states leaves the
+other 35 as holes in the map. The base layer is outlined rather than filled so it reads
+on either a light or a dark theme.
```
ahuang11: Should this be instruction? Also I'm afraid this is slightly overfitting?

lumen/ai/editors.py
```diff
@@ -352,6 +352,49 @@ def _deserialize_component(
             spec_dict.pop('pipeline')
         return type(component).from_spec(spec_dict, pipeline=pipeline)
 
+    # Data formats whose features carry geometry a geoshape mark can draw.
+    _geometry_formats = frozenset({"topojson", "geojson"})
```
ahuang11: should this be located here?

lumen/ai/agents/deck_gl.py
```diff
@@ -98,6 +98,7 @@ class DeckGLAgent(BaseCodeAgent):
             "Use for large-scale geospatial data with latitude/longitude coordinates",
             "Use for hexbin aggregations, heatmaps, or 3D extruded visualizations on maps",
             "Can render a geometry column (GeoDataFrame polygons/lines/points) on a basemap, suited to 3D/extruded maps or geometry whose schema is 'geographic' (a lat/lon CRS) so the basemap gives real-world context",
+            "NOT when the table only names places (a state or country column) without geometry or lat/lon, since deck.gl has nothing to draw and cannot fetch boundaries; that choropleth belongs to VegaLiteAgent",
```
ahuang11: Should we mention VegaLiteAgent? I think they're we should keep them isolated.


```suggestion
            "NOT when the table only names places (a state or country column) without geometry or lat/lon, since deck.gl has nothing to draw and cannot fetch boundaries",
```

ahuang11 (approved):

PR #2006, opened 2026-08-11
Add MCP server for driving Lumen from MCP clients

ghostiee-11:
Adds `mcp/`, a FastMCP server that exposes Lumen's source, SQL and Vega-Lite machinery to any MCP client, so Lumen can be driven as a service rather than only through its UI.


ahuang11:
Glanced at the changes and a couple initial thoughts:

1. Do you have a simple diagram of how this works?
2. Rather than saving to html, let's leverage panel-live-server as a dep to prevent duplication

ghostiee-11:
Screenshot 2026-08-12 at 12 17 36 PM


I agree with the Panel live server usage here.. Even i was thinking the same

philippjfr:
Reviewing now, but yes, I'd really love a high-level description of the intended use case here.

PR #2008, opened 2026-08-15
Support polars and pyarrow frames via narwhals

### LINKED ISSUE #1788: Replace pandas with narwhals
In https://github.com/holoviz/holoviews/pull/6567 we replaced the backend with Narwhals. I think we should do the same in Lumen to better support ingestion of polars/pandas/duckdb/other tabular formats.

ghostiee-11:
Adds narwhals alongside pandas so a Source can return polars or pyarrow: schema generation, `Filter` and 9 transforms run natively, `Pipeline.data` accepts non-pandas frames with a `dataframe_backend` param to pin the output type, and views and the AI layer convert back to pandas at one place each.

Anything narwhals cannot express, including `Aggregate` with the default `with_index=True`, converts to pandas and logs why, so pandas behaviour is unchanged throughout.

Closes #1788

## Notes
- Most of the new tests are a differential matrix that runs every ported transform on pandas, polars and pyarrow and compares each against the pandas result for the same spec, plus a check that the narwhals path was actually taken so the comparison cannot silently go vacuous.
- I kept the holoviews pin at `>=1.17` and convert to pandas for the `ParamFilter` `Dataset` call rather than bumping to `>=1.22`. Happy to bump instead if you prefer.
- Worth knowing before review: `Aggregate` defaults to `with_index=True`, which needs an index narwhals does not have, so the common filter into aggregate into view pipeline converts anyway. The gain shows up with `with_index=False` or no aggregate in the chain.


lumen/sources/base.py
```diff
@@ -20,6 +20,7 @@
 )
 from urllib.parse import quote, urlparse
 
+import narwhals.stable.v2 as nw
 import numpy as np
 import pandas as pd
```
ahuang11, line 25: If we have narwhals, should we still import pandas or can we remove/rewrite

ghostiee-11: pandas stays either way since hvplot, holoviews and panel all hard require it, and most of the 32 uses here are scalars or excel/json readers narwhals has no equivalent for. `pd.read_parque`t in the cache and the `pd.DataFrame(records)` constructions could move to `nw.from_dict(backend=...)`, but that needs a default backend decision.

ghostiee-11:
@philippjfr, can you please have a look at this PR.

philippjfr:
Yes will do! Very excited about this.

ghostiee-11:
Yupp.. Thanks ☺️

lumen/sources/base.py
ahuang11: I think hv migrated to narwhals to lower startup time. I forget if hvplot migrated yet

ghostiee-11: hvplot didnt migrated yet and yes hv did to narwhals

lumen/ai/agents/sql.py
```diff
@@ -429,7 +430,7 @@ async def apply_filter(field: str, value: t.Any) -> str:
             pipeline.add_filter(filt)
         except Exception as exc:
             return f"Could not apply filter on {field!r} with value {value!r}: {exc}"
-        return f"Applied filter on {field!r} ({value!r}); {len(pipeline.data)} rows match."
+        return f"Applied filter on {field!r} ({value!r}); {len(as_pandas(pipeline.data))} rows match."
```
philippjfr: Necessary to convert here? I know for a LazyFrame it probably is, but in that case I'd probably avoid getting the length here entirely.

lumen/ai/actions.py
```diff
@@ -112,7 +113,7 @@ async def _execute(self, context: TContext, **kwargs) -> tuple[list[Any], SQLQue
         out_context = {
             "source": source,
             "pipeline": pipeline,
-            "data": await describe_data(pipeline.data),
+            "data": await describe_data(as_pandas(pipeline.data)),
```
philippjfr: How hard would it be to let `describe_data` operate directly on any Narwhals compatible frame? At minimum I'd do the conversion inside `describe_data` so that we can update it in one place at a later stage.

philippjfr (commented):
Great start. I left a few comments but to be clear I'm fine with a minimal approach for now and then convert more of the machinery to operate on the Narwhals wrapper later. That said I think we should allow functions that compute descriptions and schemas to accept other data types, and do the conversion internally for now, so we can easily add the support later.

ghostiee-11:
makes sense. `get_dataframe_schema` already takes any narwhals frame and converts internally, so i'll do the same for `describe_data` and drop the call site conversions.

lumen/ai/actions.py
ghostiee-11: will move it inside `describe_data`. that also removes the same conversion from `agents/source.py` and `editors.py`, so it's one place to change when the summariser itself moves to narwhals.

lumen/ai/agents/sql.py
ghostiee-11: fair, will drop the count from the message. only other option is `select(nw.len()).collect()` so it works on a LazyFrame too, but the string is just a confirmation to the agent so not worth the query.

philippjfr (approved):

ghostiee-11:
Thanks Phillip for reviewing and approving :)

lumen/ai/utils.py
```diff
@@ -935,7 +940,7 @@ def describe_data_sync(
 
 
 async def describe_data(
-    df: pd.DataFrame,
+    df: Any,
```
ahuang11: Does Narwhals have a Type we can use?

ghostiee-11: yes, `IntoFrame | Frame` from `narwhals.stable.v2.typing`. `IntoFrame` is the native frames, `Frame` covers an already wrapped one since `as_pandas` takes those too. put it under `TYPE_CHECKING`, narwhals pulls `.typing` in itself so it costs nothing at import.

Issue #2004, opened 2026-08-11
Add a routing layer to llm.py


ahuang11:
In https://github.com/holoviz/lumen/blob/main/lumen/ai/llm.py#L2825-L2829 we hardcode:

```
    model_kwargs = param.Dict(default={
        "default": {"model": "gpt-5.4-mini"},
        "edit": {"model": "anthropic/claude-sonnet-4-5"},
        "sql": {"model": "gpt-5.4-mini"},
    }, doc="""
```

But perhaps this is the incorrect way to do it. Perhaps we should allow users to edit `select_models` (maybe make it a dict, or some way to provide description about the model) and if "routing": {"model": "some_routing_model_like_nemotron_switchyard"} is defined in model_kwargs, it allows it to auto-select based on task complexity.

Vamshi1404:
Hi @ahuang11 — traced how model_spec flows through the codebase before proposing an approach; want to align on design before writing code.

The ask: Currently each task (default/edit/sql) is hardcoded to a model via model_kwargs. Goal: make it opt-in dynamic — if a "routing" entry (e.g. {"model": "nemotron_switchyard"}) is present, a routing model auto-selects based on task complexity instead of the fixed mapping. No "routing" key → unchanged behavior.

Hook point: _get_model_kwargs (llm.py:232) is synchronous with no access to prompt content, so routing can't live there. I'd resolve it at the top of invoke()/stream() instead — once per call, before the tool loop.

Avoiding recursion: The routing model is invoked via a dict model_spec ({"model": <routing model>}), and dict specs bypass the model_kwargs lookup entirely — so routing only triggers on string specs, never on the routing model's own call. I'd document this explicitly as an invariant, not leave it implicit.

Fallback: try/except around the routing call — any failure or invalid response falls back to "default", same as today's behavior.

Open question: Should routing apply to every string model_spec, or only heavier tasks? Routing everything adds a round-trip even to lightweight calls like the "ui" ping in initialize() (llm.py:824) — but exempting specific keys means deciding which ones, which feels like your call to make.

Happy to open a draft PR early if you'd rather discuss there. Let me know which way you'd like to go on the open question and I'll start implementing.

ghostiee-11:
Fixed with #2015

PR #2015, opened 2026-08-16
Add opt-in model routing (#2004)

Vamshi1404:
## Description

I added opt-in model routing to `Llm.invoke()`, so a task type whose `model_kwargs` entry declares a `"routing"` key now invokes a routing model first to pick which entry to use for that call based on task complexity, instead of always resolving to the fixed config. Routing is opt-in per entry and resolved at a single hook point, so entries without a `"routing"` key behave byte-for-byte as before and `stream()` is covered automatically since every code path in it delegates to `invoke()`.

Relates to #2004. Draft pending design sign-off from @ahuang11 and @ghostiee-11, not ready to merge.

```python
model_kwargs = {
    "default": {"model": "gpt-5.4-mini"},
    "edit": {"model": "gpt-5.2", "routing": {"model": "nemotron-switchyard"}},
}
```

Changes:

- Routing is opt-in per `model_kwargs` entry and resolved once at the top of `Llm.invoke()` right after `_add_system_message`, so a `"routing"` key is the only signal that changes behavior.
- The routing model is always invoked with a dict `model_spec`, which `_get_model_kwargs()` returns as-is, so the routing call bypasses the `model_kwargs` lookup and can never route itself, which removes the need for a recursion guard.
- The routing response is a pydantic model whose `model_spec` field is a `Literal` rebuilt at call time from the current `model_kwargs` keys via `pydantic.create_model`, so a hallucinated key fails validation instead of silently falling back to `"default"`.
- The routing call is wrapped in try/except and logs through `log_debug(..., prefix="[LLM routing]")`, falling back to the original `model_spec` unchanged if the router raises.

No breaking change, new dependency, or migration needed. Tests added in `lumen/tests/ai/test_llm.py`: `test_route_spec_model_constrains_to_current_model_kwargs_keys`, `test_no_routing_key_passes_through`, `test_routing_invoked_with_dict_spec_and_decision_used`, `test_routing_exception_falls_back`, `test_dict_model_spec_never_routed`, and `test_invoke_uses_routed_model_spec`.

## AI Disclosure

Tool & Model: opencode (big-pickle model)

Usage: Drafted the implementation and unit tests for the routing feature, and wrote this description.

## Checklist

- [x] Added documentation

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -143,11 +143,14 @@ class Llm(param.Parameterized):
         'default', 'reasoning' and 'sql'. Agents may pick which model to
         invoke for different reasons.
 
-        An entry may declare a 'routing' key to opt in to model routing for
-        that type: the routing model is invoked first to pick which entry to
-        actually use for each call, e.g.
+        An entry may declare an optional 'description' (shown to the routing
+        model so it can reason about what each option is for) and an optional
+        'routing' key to opt in to model routing for that type: the routing
+        model is invoked first to pick which entry to actually use for each
+        call, e.g.
         {"default": {"model": "gpt-5.4-mini"},
          "edit": {"model": "gpt-5.2",
+                  "description": "Best for editing tables and visualizations",
                   "routing": {"model": "nemotron-switchyard"}}}""")
 
     tools = param.List(default=[], doc="""
@@ -226,6 +229,17 @@ class Llm(param.Parameterized):
                 f"Please specify a 'default' model in the model_kwargs "
                 f"parameter for {self.__class__.__name__}."
             )
+        for spec_name, config in self.model_kwargs.items():
+            if (
+                isinstance(config, dict)
+                and "routing" in config
+                and not isinstance(config["routing"], dict)
+            ):
+                raise ValueError(
+                    f"Invalid 'routing' entry for model spec {spec_name!r} in "
+                    f"model_kwargs: expected a dict, got "
+                    f"{type(config['routing']).__name__}."
+                )
 
     @param.depends("logfire_tags", watch=True)
     def _update_logfire_tags(self):
@@ -240,12 +254,20 @@ class Llm(param.Parameterized):
         """
         Can specify model kwargs as a dict or as a string that is a key in the model_kwargs
         or as a string that is a model type; else the actual name of the model.
+
+        The ``routing`` and ``description`` keys are routing-only directives and are
+        stripped here so they can never reach a provider's ``get_client`` and from
+        there the SDK constructor. ``_get_model_kwargs`` is the single place every
+        provider resolves its model config, so stripping here protects them all.
         """
         if isinstance(model_spec, dict):
-            return model_spec
-
-        model_kwargs = self.model_kwargs.get(model_spec) or self.model_kwargs["default"]
-        return dict(model_kwargs)
+            model_kwargs = dict(model_spec)
+        else:
+            model_kwargs = self.model_kwargs.get(model_spec) or self.model_kwargs["default"]
+            model_kwargs = dict(model_kwargs)
+        model_kwargs.pop("routing", None)
+        model_kwargs.pop("description", None)
+        return model_kwargs
 
     def _get_create_kwargs(self, response_model: type[BaseModel] | None) -> dict[str, Any]:
         kwargs = dict(self.create_kwargs)
@@ -412,6 +434,26 @@ class Llm(param.Parameterized):
             model_spec=(Literal[*tuple(self.model_kwargs)], ...),
         )
 
+    def _routing_system_prompt(self) -> str:
+        """Build the dedicated system prompt for the routing call.
+
+        The router is told it is choosing between the configured ``model_kwargs``
+        entries and is given each entry's optional ``description`` so it can
+        reason about what each option is for, rather than bare key names.
+        """
+        options = "\n".join(
+            f"- {key}: {config.get('description', 'No description provided.')}"
+            for key, config in self.model_kwargs.items()
+            if isinstance(config, dict)
+        )
+        return (
+            "You are a routing model. Based on the user's request, choose the "
+            "configured model type that is best suited to handle it.\n"
+            "Available model types:\n"
+            f"{options}\n"
+            "Reply with exactly one of the option names above."
+        )
+
     async def _resolve_routing(
         self,
         model_spec: str | dict,
@@ -419,27 +461,35 @@ class Llm(param.Parameterized):
     ) -> str | dict:
         """
         Pick which ``model_kwargs`` entry to use via a routing model when the
-        resolved config for ``model_spec`` declares one.
+        entry named by ``model_spec`` declares a ``routing`` config.
 
-        Only string ``model_spec`` values are routed. A dict ``model_spec``
-        bypasses the ``model_kwargs`` lookup entirely in ``_get_model_kwargs``,
-        which is exactly what the routing call below relies on: it always uses
-        a dict spec, so a routing call can never trigger routing for itself
-        (no recursion guard is needed).
+        Only string ``model_spec`` values that name a key in ``model_kwargs`` are
+        routed: dict specs bypass the ``model_kwargs`` lookup entirely and
+        repo-id strings (e.g. LlamaCpp) are not keys, so both are never routed.
+
+        The routing call itself uses a dict spec and ``tools=[]``, so it can
+        never trigger routing for itself nor run the real tool loop. When
+        routing picks an entry, that entry's resolved config is returned as a
+        dict: dict specs bypass ``_resolve_routing`` downstream, so a routed
+        spec threads through the tool loop and ``stream()`` recursion without
+        being re-resolved (and without paying another routing call per round).
         """
         if isinstance(model_spec, dict):
             return model_spec
 
-        config = self.model_kwargs.get(model_spec) or self.model_kwargs["default"]
-        routing_spec = config.get("routing")
+        routing_spec = self.model_kwargs.get(model_spec, {}).get("routing")
         if not routing_spec:
             return model_spec
 
+        routing_spec = dict(routing_spec)
+        routing_prompt = self._routing_system_prompt()
         try:
             route = await self.invoke(
-                messages=messages,
-                model_spec=dict(routing_spec),
+                messages=[m for m in messages if m.get("role") != "system"],
+                system=routing_prompt,
+                model_spec=routing_spec,
                 response_model=self._route_spec_model(),
+                tools=[],
             )
         except Exception:
             log_debug(
@@ -452,7 +502,7 @@ class Llm(param.Parameterized):
             )
             return model_spec
         log_debug(f"Routing {model_spec!r} -> {route.model_spec!r}", prefix="[LLM routing]")
-        return route.model_spec
+        return dict(self._get_model_kwargs(route.model_spec))
 
     async def invoke(
         self,
@@ -609,12 +659,19 @@ class Llm(param.Parameterized):
         self,
         tools: list[dict[str, Any] | FunctionTool | MCPTool] | None,
     ) -> list[dict[str, Any] | FunctionTool | MCPTool] | None:
-        """Combine instance-level ``self.tools`` with per-call *tools*."""
-        if self.tools and tools:
+        """Combine instance-level ``self.tools`` with per-call *tools*.
+
+        ``tools=None`` means "use instance tools"; an explicit empty list opts
+        out of instance tools entirely (used by the routing call so the routing
+        model never runs the real tool loop).
+        """
+        if tools is None:
+            return list(self.tools) if self.tools else None
+        if not tools:
+            return []
+        if self.tools:
             return list(self.tools) + list(tools)
-        elif self.tools:
-            return list(self.tools)
-        return tools
+        return list(tools)
 
     @classmethod
     def _normalize_tools(
@@ -925,6 +982,7 @@ class Llm(param.Parameterized):
         """
         combined_tools = self._combine_tools(tools)
         _, tool_instances, tool_contexts = self._normalize_tools(combined_tools)
+        model_spec = await self._resolve_routing(model_spec, messages)
         messages, contains_image = self._check_for_image(messages)
         if self.logfire_tags is not None or contains_image:
             output = await self.invoke(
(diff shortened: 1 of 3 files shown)


ghostiee-11:
Hey @Vamshi1404, Thanks for the contribution, can you please use this https://github.com/holoviz-dev/holoviz-skills/blob/main/contributing-to-holoviz/skills/pr-description/SKILL.md for writing PR description

Vamshi1404:
Thanks @ghostiee-11 — updated the description using that skill (fixed the title to conventional-commit style, corrected the AI disclosure, and added a runnable example). Let me know if anything else needs adjusting.

lumen/ai/llm.py
```diff
@@ -393,6 +400,60 @@ def warmup(cls, model_kwargs: dict | None):
...
+        Only string ``model_spec`` values are routed. A dict ``model_spec``
+        bypasses the ``model_kwargs`` lookup entirely in ``_get_model_kwargs``,
+        which is exactly what the routing call below relies on: it always uses
+        a dict spec, so a routing call can never trigger routing for itself
+        (no recursion guard is needed).
+        """
+        if isinstance(model_spec, dict):
+            return model_spec
+
+        config = self.model_kwargs.get(model_spec) or self.model_kwargs["default"]
```
ghostiee-11: routing never gets stripped so it lands in the sdk constructor

```python
await OpenAI(api_key="sk-test", model_kwargs={
    "default": {"model": "base"},
    "edit": {"model": "edit-model", "routing": {"model": "router"}},
}).get_client("edit")
# TypeError: AsyncOpenAI.__init__() got an unexpected keyword argument 'routing'
```

same on anthropic, litellm sends it as a request param. and it hits both paths, fallback returns "edit" which is the entry holding routing, and the Literal has "edit" in it too

can you pop it in the base `_get_model_kwargs` instead, fixes every provider at once. then `_resolve_routing` can just call `_get_model_kwargs` and not repeat the lookup, that also stops it clobbering llamacpp repo-id specs

lumen/ai/llm.py
```diff
@@ -393,6 +400,60 @@ def warmup(cls, model_kwargs: dict | None):
...
+        routing_spec = config.get("routing")
+        if not routing_spec:
+            return model_spec
+
+        try:
+            route = await self.invoke(
```
ghostiee-11, line 562: router isnt told its routing, no system prompt no field description nothing about what the keys mean. on openai it inherits the task system prompt, on anthropic it gets none

```
OpenAI    -> [{'role': 'system', 'content': 'You are a SQL agent...'}, {'role': 'user', ...}]
Anthropic -> [{'role': 'user', ...}]
```

can we give it a routing prompt and per entry descriptions

lumen/ai/llm.py
```diff
@@ -393,6 +400,60 @@ def warmup(cls, model_kwargs: dict | None):
...
+                messages=messages,
```
ghostiee-11: no tools passed so `_combine_tools(None)` returns self.tools and the router runs the real tool loop, side effects fire twice per turn. tools=[] here

lumen/ai/llm.py
```diff
@@ -393,6 +400,60 @@ def warmup(cls, model_kwargs: dict | None):
...
+                model_spec=dict(routing_spec),
```
ghostiee-11: dict(routing_spec) is inside the try, so `"routing": "router-model"` raises ValueError, gets swallowed and routing just silently never runs. move it above the try and validate in `__init__` next to the "default" check

lumen/ai/llm.py
```diff
@@ -431,6 +492,7 @@ async def invoke(
         """
         system = system.strip().replace("\n\n", "\n")
         messages, input_kwargs = self._add_system_message(messages, system, input_kwargs)
+        model_spec = await self._resolve_routing(model_spec, messages)
```
ghostiee-11, line 621: stream() recurses into invoke() per tool round so this re-routes every round and the model can flip mid response. also its a blocking call before first token on self.timeout which is 120s. can we resolve once per turn

lumen/tests/ai/test_llm.py
```diff
@@ -870,3 +870,110 @@ def test_api_key_env_var_does_not_clobber_provider_default(monkeypatch):
...
+    run_client_specs = []
+
+    async def fake_route(model_spec, messages):
+        return "sql"
+
+    async def fake_run_client(model_spec, messages, **kwargs):
+        run_client_specs.append(model_spec)
+        return "done"
+
+    monkeypatch.setattr(llm, "_resolve_routing", fake_route)
```
ghostiee-11: these patch out the thing they test so nothing reaches get_client, all 7 pass while the feature cant complete a call. one test asserting get_client works with a routing entry would have caught it

`_routing_llm()` also builds the abstract Llm directly, can it use `_make()` like the rest of the file

ghostiee-11 (commented):
went through this locally against 826042b. the routing idea reads well but it can't complete a call as written, details inline. main one is the routing key never getting stripped before the kwargs hit the sdk constructor.

lumen/ai/llm.py
```diff
@@ -863,6 +982,7 @@ async def stream(
         """
         combined_tools = self._combine_tools(tools)
         _, tool_instances, tool_contexts = self._normalize_tools(combined_tools)
+        model_spec = await self._resolve_routing(model_spec, messages)
```
ghostiee-11: when the router is down stream() routes twice (invoke 1 attempt, stream 2), each on self.timeout, so up to 4 min before the fallback. can stream() skip the re-resolve?

lumen/ai/llm.py
```diff
@@ -393,6 +422,88 @@ def warmup(cls, model_kwargs: dict | None):
...
+
+        routing_spec = dict(routing_spec)
+        routing_prompt = self._routing_system_prompt()
+        try:
+            route = await self.invoke(
+                messages=[m for m in messages if m.get("role") != "system"],
+                system=routing_prompt,
+                model_spec=routing_spec,
+                response_model=self._route_spec_model(),
+                tools=[],
```
ghostiee-11, line 567: `_normalize_tools([])` returns `[]` not None so `tools: []` goes to the provider, instructor only overwrites it in tools mode not the json modes. return None for empty instead?

lumen/ai/llm.py
```diff
@@ -2556,10 +2675,7 @@ def __init__(self, **params):
             self._instructor_wrapper = "openai"
 
     def _get_model_kwargs(self, model_spec: str | dict) -> dict[str, Any]:
-        if isinstance(model_spec, dict):
-            return model_spec
-        model_kwargs = self.model_kwargs.get(model_spec) or self.model_kwargs["default"]
-        return dict(model_kwargs)
+        return super()._get_model_kwargs(model_spec)
```
ghostiee-11: this is just `return super()._get_model_kwargs(model_spec)` now, can it go?

lumen/tests/ai/test_llm.py
```diff
@@ -870,3 +871,234 @@ def test_api_key_env_var_does_not_clobber_provider_default(monkeypatch):
 
     monkeypatch.setenv("OPENAI_API_KEY", "sk-from-env")
     assert OpenAI().api_key == "sk-from-env"
+
+
+# ---------------------------------------------------------------------------
+# Model routing
+# ---------------------------------------------------------------------------
+
+ROUTED_MODEL_KWARGS = {
```
ghostiee-11: no description key in the fixture so the description strip isnt covered, worth adding one

lumen/ai/llm.py
Vamshi1404: _resolve_routing fallback now returns dict(self._get_model_kwargs(model_spec)) instead of the bare string. Both success and fallback paths return dicts, so the dict-bypass prevents re-resolution downstream. Exactly one routing attempt per turn regardless of router state.

lumen/ai/llm.py
Vamshi1404: Changed if tools is None: to if not tools: in _normalize_tools. Empty list now returns None → tools: key is never set on kwargs → instructor/SDK doesn't see tools: [] in structured-output mode

lumen/ai/llm.py
Vamshi1404: Deleted — it was a pure super() delegation.

lumen/tests/ai/test_llm.py
Vamshi1404: "description": "Best for editing tables and visualizations" added to the "edit" entry in ROUTED_MODEL_KWARGS. New test test_get_client_strips_description_key asserts description never reaches the SDK constructor.

lumen/ai/llm.py
```diff
@@ -431,6 +542,7 @@ async def invoke(
         """
         system = system.strip().replace("\n\n", "\n")
         messages, input_kwargs = self._add_system_message(messages, system, input_kwargs)
+        model_spec = await self._resolve_routing(model_spec, messages)
```
ghostiee-11, line 621: this runs before `_check_for_image`, so the router gets the raw image and the nested invoke base64s it and uploads the whole thing to a model picked for being small and text-only.

lumen/ai/llm.py
```diff
@@ -393,6 +422,88 @@ def warmup(cls, model_kwargs: dict | None):
...
+
+        routing_spec = self.model_kwargs.get(model_spec, {}).get("routing")
+        if not routing_spec:
+            return model_spec
+
+        routing_spec = dict(routing_spec)
+        routing_prompt = self._routing_system_prompt()
+        try:
+            route = await self.invoke(
+                messages=[m for m in messages if m.get("role") != "system"],
```
ghostiee-11: router gets the entire history, 21 messages / 4385 chars on my run, same as the real call. last user message would do?

lumen/ai/llm.py
```diff
@@ -393,6 +422,88 @@ def warmup(cls, model_kwargs: dict | None):
...
+
+    def _routing_system_prompt(self) -> str:
+        """Build the dedicated system prompt for the routing call.
+
+        The router is told it is choosing between the configured ``model_kwargs``
+        entries and is given each entry's optional ``description`` so it can
+        reason about what each option is for, rather than bare key names.
+        """
+        options = "\n".join(
+            f"- {key}: {config.get('description', 'No description provided.')}"
```
ghostiee-11: out of the box every line is `- default: No description provided.` since no provider default ships a description. `LLMDialog._get_all_agent_types()` already builds spec key -> description, could reuse that.

lumen/ai/llm.py
```diff
@@ -393,6 +422,88 @@ def warmup(cls, model_kwargs: dict | None):
...
+        never trigger routing for itself nor run the real tool loop. Both
+        the success and fallback paths return a dict config: dict specs
+        bypass ``_resolve_routing`` downstream, so a resolved spec threads
+        through the tool loop and ``stream()`` recursion without being
+        re-resolved (and without paying another routing call per round).
+        """
+        if isinstance(model_spec, dict):
+            return model_spec
+
+        routing_spec = self.model_kwargs.get(model_spec, {}).get("routing")
```
ghostiee-11: agents pass their own spec key (sql, chat, vega_lite, analysis) and none of those exist in any provider default, so with routing on "default" the router only fires for model_spec="default" which lumen itself never sends:

```
model_spec='sql'       -> ['gpt-5.6-luna']            routed=False
model_spec='vega_lite' -> ['gpt-5.6-luna']            routed=False
model_spec='default'   -> ['router', 'gpt-5.4-nano']  routed=True
```

can the fallback entry's routing apply too, guarded on the repo-id case instead of key presence?

lumen/ai/llm.py
Vamshi1404: Fixed: _check_for_image now runs before _resolve_routing in both invoke() and stream(), and the routing call strips any image content from messages regardless (new _strip_for_routing), so the router never receives raw or base64'd image data.

lumen/ai/llm.py
Vamshi1404: Trimmed: the routing call now receives only the last user message via _strip_for_routing, not the full conversation history. Added a test confirming this.

lumen/ai/llm.py
Vamshi1404: Reused LLMDialog._get_all_agent_types() as you suggested — added a spec_descriptions fallback so descriptions populate out of the box, with an explicit description in model_kwargs still taking priority if set.

lumen/ai/llm.py
Vamshi1404: You're right, and it's worse than the one example ; traced every model_spec= call site: 13 real agent-derived keys, none shipped in any provider's model_kwargs defaults, so routing could only ever fire on edit or a manually-added key. Wrote up three options with trade-offs and posted them as a top-level comment below since this is a design decision bigger than this thread ;leaning toward auto-populating agent spec keys at Llm init, and already confirmed that's not blocked by a circular import (lazy import inside __init__ works, same pattern llm_dialog.py already uses). Holding off on implementing until you weigh in on which direction you'd want.

lumen/tests/ai/test_llm.py
```diff
@@ -870,3 +871,519 @@ def test_api_key_env_var_does_not_clobber_provider_default(monkeypatch):
...
+    "default": {"model": "base-model"},
+    "edit": {"model": "edit-model", "routing": {"model": "router-model"}, "description": "Best for editing tables and visualizations"},
+    "sql": {"model": "sql-model"},
+}
```
ahuang11: Constants should be pulled up top.

lumen/tests/ai/test_llm.py
```diff
@@ -870,3 +871,519 @@ def test_api_key_env_var_does_not_clobber_provider_default(monkeypatch):
...
+
+
+def _routing_llm():
+    llm = _make(OpenAI, {})
+    llm.model_kwargs = {k: dict(v) for k, v in ROUTED_MODEL_KWARGS.items()}
+    return llm
```
ahuang11: Should this be a pytest fixture

ahuang11:
Thanks for contributing! I think this is great. If possible I'd love to see a video of this working, which brings me to my next idea: can we add the model used in the ChatMessage sent footer?

Here's my test version (I realized footer objects and header objects are located weirdly; maybe can be an upstream fix so I hijacked timestamp format)

```
import panel_material_ui as pmui
import panel as pn

pn.extension()

model = "gpt-5.6-luna"
pmui.ChatMessage(
    "Test", timestamp_format=f"%H:%M %p (used {model})"
)
```

image

ghostiee-11:
I am opening an issue for this @ahuang11, good idea!!

ghostiee-11 (approved):

ghostiee-11:
Thanks @Vamshi1404, for taking this up!!

PR #2018, opened 2026-08-17
Fix Filter warning formatting

### LINKED ISSUE #2013: Filter warning is missing its f prefix and references an out of scope name
The "condition not understood" warning in `Filter.apply` (`transforms/base.py`) is missing its `f` prefix, so it prints the braces literally:

```
Condition {val!r} on {col!r} column not understood. Filter query will not be applied.
```

`col` is also not a name in that scope, the loop variable is `k`, so adding `f` alone would raise `NameError`.

yiboban:
## Summary

- format invalid `Filter` condition warnings with the actual condition value
- reference the in-scope column key instead of the undefined `col` name
- add a regression test for the warning and unchanged DataFrame

## Testing

- `python -m pytest lumen/tests/transforms/test_base.py -q`

Fixes #2013

### DIFF
--- a/lumen/tests/transforms/test_base.py
+++ b/lumen/tests/transforms/test_base.py
@@ -5,7 +5,7 @@ import pandas as pd
 import param  # type: ignore
 
 from lumen.transforms.base import (
-    Count, DropNA, Eval, Sum, Transform, project_lnglat,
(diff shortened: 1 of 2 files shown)


ghostiee-11 (approved):

ghostiee-11:
Thanks, @yiboban for the contribution 😊

PR #2020, opened 2026-08-17
chores: lint imports that are not at the top of a file

ghostiee-11:
Turns on ruff's PLC0415 so imports buried inside functions get caught by lint instead of in review. Imports that were deferred for no reason are moved to the top; the noqa is left only where the deferral is real, meaning an optional dependency, a circular import, or a module heavy enough to keep off the import path.

### DIFF
--- a/lumen/ai/agents/source.py
+++ b/lumen/ai/agents/source.py
@@ -29,6 +29,7 @@ from ..controls.ingest.result import SourceResult
 from ..llm import Message
 from ..schemas import Metaset, get_metaset
 from ..tools import FunctionTool
+from ..tools.source_lookup import _control_hash
 from ..translate import doc_descriptions
 from ..utils import (
     describe_data, get_pipeline, log_debug, result_to_dataframe,
@@ -261,9 +262,6 @@ class SourceAgent(Agent):
         ctrl_map: dict[str, BaseSourceControls] = {}
         for c in controls:
             if c._supports_tools:
-                from ..tools.source_lookup import (  # noqa: PLC0415
-                    _control_hash,
-                )
                 ctrl_map[_control_hash(c)] = c
 
         tools: list[FunctionTool] = []
--- a/lumen/ai/agents/source.py
+++ b/lumen/ai/agents/source.py
@@ -261,7 +261,9 @@ class SourceAgent(Agent):
         ctrl_map: dict[str, BaseSourceControls] = {}
         for c in controls:
             if c._supports_tools:
-                from ..tools.source_lookup import _control_hash
+                from ..tools.source_lookup import (  # noqa: PLC0415
+                    _control_hash,
+                )
                 ctrl_map[_control_hash(c)] = c
 
         tools: list[FunctionTool] = []
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -765,6 +765,7 @@ class Llm(param.Parameterized):
             else:
                 tool_context = None
             if callable(tool) and hasattr(tool, "__lumen_tool_annotations__"):
+                # Deferred: .tools -> .tools.base -> .actor -> .llm is a cycle.
                 from .tools import FunctionTool  # noqa: PLC0415
                 tool = FunctionTool(tool)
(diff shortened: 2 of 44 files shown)


lumen/ai/agents/source.py
```diff
@@ -261,7 +261,9 @@ def _build_tools(
         ctrl_map: dict[str, BaseSourceControls] = {}
         for c in controls:
             if c._supports_tools:
-                from ..tools.source_lookup import _control_hash
+                from ..tools.source_lookup import (  # noqa: PLC0415
```
ahuang11: I don't think this noqa is necessary

lumen/ai/controls/ingest/file.py
```diff
@@ -261,7 +261,7 @@ def _extract_metadata_content(self, file_obj: io.BytesIO, extension: str) -> str
                 return json.dumps(json.loads(content), indent=2)
             except json.JSONDecodeError:
                 return content
-        from markitdown import MarkItDown
+        from markitdown import MarkItDown  # noqa: PLC0415
```
ahuang11: I thought we migrated this to utils?

ahuang11 (changes requested):
Lots of noqas are unnecessary; please review them carefully. For example if they are `.tools.base import FunctionTool` it's already imported elsewhere afaik so nesting them under a function won't be lazily loaded.

lumen/ai/llm.py
```diff
@@ -765,7 +765,7 @@ def _normalize_tools(
             else:
                 tool_context = None
             if callable(tool) and hasattr(tool, "__lumen_tool_annotations__"):
-                from .tools import FunctionTool
+                from .tools import FunctionTool  # noqa: PLC0415
```
ahuang11, line 769: Do we need this nested?

lumen/ai/llm.py
```diff
@@ -941,7 +941,7 @@ async def _run_tool_calls(
         tool_contexts: dict[str, Any],
         messages: list[Message],
     ) -> list[Message]:
-        from .tools import FunctionTool, MCPTool
+        from .tools import FunctionTool, MCPTool  # noqa: PLC0415
```
ahuang11, line 947: Same here; do we need this nested? I imagine lumen automatically importing all.

lumen/ai/vector_store.py
```diff
@@ -584,7 +584,9 @@ async def add_file(
         -------
         List of assigned IDs for the added items.
         """
-        from markitdown import FileConversionException, StreamInfo
+        from markitdown import (  # noqa: PLC0415
+            FileConversionException, StreamInfo,
+        )
```
ahuang11, line 591: Do we need this nested? pretty sure utils already imports this

ghostiee-11:
These three are all real. `lumen.ai.tools` imports `lumen.ai.tools.base`, which imports Actor from `lumen.ai.actor`, which imports `lumen.ai.llm` hoisting either one gives `ImportError: cannot import name 'Actor' from partially initialized module 'lumen.ai.actor'`. For markitdown, `get_markitdown` in config.py only defers the import, it never runs it at module scope, so `markitdown` is not in `sys.modules` after `import lumen.ai` and pulling it in costs 2s. I went back through all 85 noqas the same way and none of them turned out to be removable, so I added a one-line reason at each site that isn't obvious from the module name.

ahuang11:
Thanks for tracing it. I'd love to see inline comments for the ones that are unintuitive

ghostiee-11:
Sure

ghostiee-11:
Added them. state.py has twelve, so each one names the module it pulls in (`# Deferred: .variables imports state at module scope.`); same for the pipeline/duckdb, tiktoken and hvplot accessor sites. Left the plain optional-dependency ones bare since the module name is the reason.

One correction to my last comment: I said vl_convert was covered by a try/except, but that except only catches ValueError, so a missing wheel already raised. That deferral bought nothing, so it's a top-level import now.

ahuang11 (approved):

ghostiee-11:
Thanks for approving and reviewing @ahuang11 :)

PR #2022, opened 2026-08-18
fix: update ingestion controls docs

### LINKED ISSUE #1974: Update ingestion controls docs
I think https://lumen.holoviz.org/configuration/controls/ is outdated vs https://github.com/holoviz/lumen/tree/main/lumen/ai/controls/ingest

Mr-Neutr0n:
Fixes #1974

Updated the ingestion controls docs (docs/configuration/controls.md) to match the current `lumen/ai/controls/ingest` implementation: fixed outdated class names (`UploadControls`→`UploadSourceControls`, `DownloadControls`→`DownloadSourceControls`), corrected the upload button label to "Upload file(s)", documented the `FileSourceControls` data/metadata file classification, and added missing documentation for `CatalogSourceControls`, `RESTAPISourceControls`, and `OpenAPISourceControls` plus updated the "See also" tutorial links.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/docs/configuration/controls.md
+++ b/docs/configuration/controls.md
@@ -63,7 +63,7 @@ Requires the `kagglehub` package:
 pip install kagglehub
 ```
 
-If `kagglehub` is not installed, Kaggle URL support is silently disabled.
+If `kagglehub` is not installed, the input placeholder stops advertising Kaggle URLs, and pasting one returns a message telling you to install the package.
 
 ### Creating custom controls
 
@@ -429,7 +429,7 @@ class MesonetDailyControls(URLSourceControls):
 
 ### RESTAPISourceControls
 
-`RESTAPISourceControls` exposes manually-defined REST API endpoints. Each endpoint is described with `method`, `path`, `summary`, `description`, and `parameters`, and the control builds a typed callable (rendered as widgets) for each one.
+`RESTAPISourceControls` exposes manually-defined REST API endpoints. Each endpoint is described with `method`, `path`, `summary`, `description`, and `parameters`, and the control builds a callable (rendered as widgets) for each one.
 
 ```python
 from lumen.ai.controls import RESTAPISourceControls
--- a/docs/configuration/controls.md
+++ b/docs/configuration/controls.md
@@ -26,16 +26,30 @@ Common use cases include:
 
 ### Built-in controls
 
-| Control | Use for |
-|---------|---------|
-| `UploadControls` | Uploading local files (CSV, Excel, etc.) |
-| `DownloadControls` | Fetching data from URLs (including Kaggle datasets) |
+Lumen ships with a set of built-in source controls:
 
-`UploadControls` stages selected files before processing. After selecting files, click `Confirm file(s)` to process them, or `Clear selected` to reset the staged selection.
+| Control | Type | Use for |
+|---------|------|---------|
+| `UploadSourceControls` | Concrete | Uploading local files (CSV, Excel, etc.) |
+| `DownloadSourceControls` | Concrete | Fetching data from URLs (including Kaggle datasets) |
+| `CodeSourceControls` | Concrete | Wrapping Python functions or object methods as data sources |
+| `URLSourceControls` | Concrete | Fetching data from URL templates |
+| `RESTAPISourceControls` | Concrete | Defining REST API endpoints manually |
+| `OpenAPISourceControls` | Concrete | Auto-discovering endpoints from an OpenAPI spec |
+| `CatalogSourceControls` | Base class | Browsing and loading from a pre-fetched dataset catalog |
+
+`UploadSourceControls` stages selected files before processing. After selecting files, click `Upload file(s)` to process them, or `Clear selected` to reset the staged selection.
+
+Both `UploadSourceControls` and `DownloadSourceControls` inherit from `FileSourceControls`, which classifies each file as either **data** or **metadata**:
+
+- **Data files** (CSV, Parquet, JSON, Excel, GeoJSON, HTML, scientific formats, etc.) are parsed into DuckDB tables.
+- **Metadata files** (e.g. `md`, `txt`, `yaml`, `yml`, `pdf`, `docx`, `pptx`, ...) or files whose name contains `metadata`, `readme`, or `schema` are added to the document vector store so the agent can retrieve their contents.
+
+For each file you can toggle between `data` and `metadata`, edit its table alias, and (for `.xlsx` files) pick the sheet to import.
 
 #### Kaggle datasets
 
-`DownloadControls` can fetch datasets directly from Kaggle. Paste a Kaggle dataset URL into the URL input field:
+`DownloadSourceControls` can fetch datasets directly from Kaggle. Paste a Kaggle dataset URL into the URL input field:
 
 ```
 https://www.kaggle.com/datasets/sharmajicoder/gen-z-social-media-usage-dataset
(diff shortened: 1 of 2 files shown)


ghostiee-11:
Thanks @Mr-Neutr0n, for your contribution.

docs/configuration/controls.md
```diff
@@ -26,16 +26,30 @@ Common use cases include:
...
+Lumen ships with a set of built-in source controls:
 
-`UploadControls` stages selected files before processing. After selecting files, click `Confirm file(s)` to process them, or `Clear selected` to reset the staged selection.
+| Control | Type | Use for |
+|---------|------|---------|
+| `UploadSourceControls` | Concrete | Uploading local files (CSV, Excel, etc.) |
+| `DownloadSourceControls` | Concrete | Fetching data from URLs (including Kaggle datasets) |
+| `CodeSourceControls` | Concrete | Wrapping Python functions or object methods as data sources |
+| `URLSourceControls` | Concrete | Fetching data from URL templates |
+| `RESTAPISourceControls` | Concrete | Defining REST API endpoints manually |
```
ghostiee-11: nit: these two are really subclass-only. `RESTAPISourceControls`'s docstring starts with "Base class ..." and `url_template` defaults to `""`.

docs/configuration/controls.md
```diff
@@ -26,16 +26,30 @@ Common use cases include:
...
+| `OpenAPISourceControls` | Concrete | Auto-discovering endpoints from an OpenAPI spec |
+| `CatalogSourceControls` | Base class | Browsing and loading from a pre-fetched dataset catalog |
+
+`UploadSourceControls` stages selected files before processing. After selecting files, click `Upload file(s)` to process them, or `Clear selected` to reset the staged selection.
```
ghostiee-11: `Upload file(s)` is only on `UploadSourceControls`; download still says `Confirm file(s)`. could we mention both since the next paragraph covers both classes?

docs/configuration/controls.md
```diff
@@ -26,16 +26,30 @@ Common use cases include:
...
+
+Both `UploadSourceControls` and `DownloadSourceControls` inherit from `FileSourceControls`, which classifies each file as either **data** or **metadata**:
+
+- **Data files** (CSV, Parquet, JSON, Excel, GeoJSON, HTML, scientific formats, etc.) are parsed into DuckDB tables.
```
ghostiee-11: is json meant to be in this list? it's in `METADATA_EXTENSIONS` too and `UploadedFileRow.__init__` checks that one first, so uploading a `.json` gives you metadata, not a table.

ghostiee-11 (commented):
A few of the new claims don't match what the code does, though I left those inline.

docs/configuration/controls.md
```diff
@@ -26,16 +26,30 @@ Common use cases include:
...
+- **Metadata files** (e.g. `md`, `txt`, `yaml`, `yml`, `pdf`, `docx`, `pptx`, ...) or files whose name contains `metadata`, `readme`, or `schema` are added to the document vector store so the agent can retrieve their contents.
```
ghostiee-11: the patterns are `_metadata` / `metadata_`, so plain `metadata.csv` won't match here. also missing `json`, `doc`, `ppt` from the list.

docs/configuration/controls.md
```diff
@@ -221,9 +235,55 @@ def _update_year_options(self):
 - **Cache API responses** when possible to avoid redundant calls
 - **Use `normalize_table_name()`** to ensure DuckDB-compatible table names
 
+## Catalog controls
+
+`CatalogSourceControls` is a base class for controls that browse a pre-fetched catalog of datasets, letting users click a row to ingest that dataset. When a `vector_store` is provided, catalog entries are embedded in the background so the `SourceAgent` can search and load them from a natural-language query.
```
ghostiee-11: think this reads like you need a vector store for agent search. `as_tools()` always exposes the tool and `_search_catalog` keyword-matches `search_columns` when there isn't one, so the vector store just makes it semantic.

docs/configuration/controls.md
```diff
@@ -221,9 +235,55 @@ def _update_year_options(self):
...
+
+Class-level attributes configure the `Tabulator` browser:
+
+| Attribute | Purpose |
+|-----------|---------|
+| `display_columns` | `{col: {"title": ..., "width": ..., "formatter": ...}}` |
+| `filter_columns` | `{col: header_filter config}` for column filtering |
+| `search_columns` | Column names concatenated for vector embedding |
+| `detail_columns` | Column names shown in the expanded row detail view |
+| `vector_store` | Optional `VectorStore`; entries are embedded for agent search |
```
ghostiee-11: `vector_store` doesn't touch the Tabulator, and we pass it per instance everywhere else. maybe pull it out of this table?

docs/configuration/controls.md
```diff
@@ -221,9 +235,55 @@ def _update_year_options(self):
...
+    }
+    search_columns = ["name", "description"]
+    detail_columns = ["description", "url"]
+
+    async def _load_catalog(self) -> pd.DataFrame:
+        return await fetch_catalog()
+
+    async def _fetch_entry(self, entry: pd.Series) -> SourceResult:
+        df = await download(entry["url"])
+        return SourceResult.from_dataframe(df, entry["name"])
```
ghostiee-11: `from_dataframe` drops the name into an unquoted `SELECT * FROM {table_name}`, so anything with a space breaks. can we use `normalize_table_name()` here since we recommend it a few lines up?

docs/configuration/controls.md
```diff
@@ -221,9 +235,55 @@ def _update_year_options(self):
...
+```
+
+Agents can ingest a dataset by calling `load_entry(row_idx)`, or by searching the catalog via the control's tool exposed through `as_tools()`.
```
ghostiee-11: `load_entry` doesn't have any callers, agents go through `as_tools()` / `_load_from_query`. its docstring says "DataFrame index" but `_fetch_entry_by_index` uses `.iloc`, so probably better left out.

docs/configuration/controls.md
```diff
@@ -363,6 +423,64 @@ class MesonetDailyControls(URLSourceControls):
...
+)
+```
+
+| Parameter | Purpose |
+|-----------|---------|
+| `base_url` | Root URL for all API requests |
+| `headers` | Default HTTP headers sent with every request |
+| `endpoints` | `{display_name: endpoint_spec}`; each spec has `method`, `path`, `summary`, `description`, `parameters` |
+
+Endpoint parameters support `type`, `enum` (rendered as a dropdown), `required`, `default`, and `description`. Override `_resolve_param_type()` for richer type mapping.
```
ghostiee-11: `_resolve_param_type` just returns `str` in the base class, so `type` doesn't do anything unless you subclass. worth saying that outright, otherwise `"type": "integer"` looks like it should work.

docs/configuration/controls.md
```diff
@@ -363,6 +423,64 @@ class MesonetDailyControls(URLSourceControls):
...
+
+### OpenAPISourceControls
+
+`OpenAPISourceControls` fetches an OpenAPI 3.x spec at startup, resolves `$ref` references, and registers every path/method pair as an action — no manual endpoint definitions needed.
+
+```python
+from lumen.ai.controls import OpenAPISourceControls
+
+controls = OpenAPISourceControls(
```
ghostiee-11, line 471: both this and the catalog example load in `pn.state.onload`, which doesn't fire outside a served session, so in a script `endpoints` stays None and the layout sits on `loading=True`. maybe "on session load" instead of "at startup"?

docs/configuration/controls.md
```diff
@@ -383,3 +501,5 @@ The agent sees each control's actions as tools and can call them with appropriat
 
 - [Building a Census Data Explorer](../examples/tutorials/census_data_ai_explorer.md) — Complete walkthrough with BaseSourceControls
 - [Building a Weather Data Explorer](../examples/tutorials/mesonet_weather_explorer.md) — URLSourceControls tutorial with preprocessing
+- [Building a Weather API Explorer](../examples/tutorials/weather_openapi_explorer.md) — OpenAPISourceControls tutorial with auto-discovered endpoints
```
ghostiee-11, line 508: these two aren't in the `zensical.toml` nav yet, so they're orphaned and mkdocs warns. mind adding them next to the other tutorials?

ghostiee-11 (approved):

PR #2029, opened 2026-08-18
fix: handle non-default catalog indexes

### LINKED ISSUE #2024: `CatalogSourceControls` mixes label and positional indexing, loading the wrong dataset
`_search_catalog` returns a label from `catalog_df.iterrows()` (catalog.py:363) and `_embed` stores the same label as `_row_idx` (catalog.py:250), but `_load_from_query` looks it up with `.iloc` (catalog.py:291).

Any `_load_catalog()` returning a filtered or concatenated frame has a non-`RangeIndex`, so a match on label 42 loads position 42: either `IndexError` or a completely different dataset with no error. A string index also makes `int(idx)` raise during `_embed`. Using `.loc` on both sides would fix it.

fly1d:
## Description

I fixed `CatalogSourceControls` so catalog search results use row positions end to end: `_embed()` stores JSON-safe integer positions, `_search_catalog()` returns positions, and `_load_from_query()` validates them before using `.iloc`. This prevents filtered, concatenated, duplicate-label, and MultiIndex catalogs from loading the wrong entry or failing, and I route agent-driven loads through `_handle_success()` so `outputs["table"]`, `_last_table`, source deduplication, and events stay synchronized. There are no breaking API changes, new dependencies, or migration steps.

```python
import asyncio

import pandas as pd

from lumen.ai.controls import CatalogSourceControls, SourceResult


class Catalog(CatalogSourceControls):
    search_columns = ["name"]

    def __init__(self, frame):
        self._frame = frame
        super().__init__(context={})

    async def _load_catalog(self):
        return self._frame

    async def _fetch_entry(self, entry):
        return SourceResult(table=entry["name"])


async def main():
    first = pd.DataFrame({"name": ["first", "second"]})
    second = pd.DataFrame({"name": ["third", "target"]})
    controls = Catalog(pd.concat([first, second]))
    await asyncio.sleep(0)
    result = await controls._load_from_query("target")
    assert result.table == "target"


asyncio.run(main())
# Before: the duplicate label makes the old lookup load "second" instead of "target".
# After: the positional lookup loads "target".
```

I ran `pixi run -e test-312 pytest lumen/tests/ai/test_controls/test_catalog_source_controls.py -q` (5 passed), `pixi run -e test-312 pytest lumen/tests/ai/test_controls -q` (195 passed, 2 skipped), `pixi run -e test-312 test-unit` (2040 passed, 42 skipped), and `pixi run -e lint lint` (passed).

Fixes #2024

## AI Disclosure

Tool & Model: OpenAI Codex + GPT-5
Usage: I used Codex to inspect the issue and review feedback, implement the positional lookup and output synchronization changes, add regression tests, run the listed verification, and draft this description. I reviewed the resulting code, test coverage, and description before submission.

## Checklist

### DIFF
--- a/lumen/ai/controls/ingest/catalog.py
+++ b/lumen/ai/controls/ingest/catalog.py
@@ -238,19 +238,16 @@ class CatalogSourceControls(BaseSourceControls):
             return
 
         items = []
-        for idx, row in self.catalog_df.iterrows():
+        for pos, (_, row) in enumerate(self.catalog_df.iterrows()):
             text = self._entry_to_text(row)
             if not text:
                 continue
 
             # Structured metadata for post-semantic filtering
             # VectorStore metadata values must be str/int/float/bool
-            row_idx = idx.item() if hasattr(idx, "item") else idx
-            if not isinstance(row_idx, (str, int, float, bool)):
-                row_idx = str(row_idx)
             metadata = {
                 "type": "catalog_entry",
-                "_row_idx": row_idx,
+                "_row_idx": pos,
                 "_control_id": id(self),
             }
             for col in self.filter_columns:
@@ -286,24 +283,22 @@ class CatalogSourceControls(BaseSourceControls):
             return SourceResult.empty("Catalog not yet loaded.")
 
         match_idx = await self._search_catalog(query)
-        if match_idx is None:
-            return SourceResult.empty(
-                f"No dataset matching '{query}' found in catalog."
-            )
-
-        entry = self.catalog_df.loc[match_idx]
+        no_match = f"No dataset matching '{query}' found in catalog."
         try:
+            if (
+                not isinstance(match_idx, int)
+                or match_idx < 0
+                or match_idx >= len(self.catalog_df)
+            ):
+                return SourceResult.empty(no_match)
+            try:
+                entry = self.catalog_df.iloc[match_idx]
+            except (IndexError, TypeError):
+                return SourceResult.empty(no_match)
             result = await self._fetch_entry(entry)
-            # Register the source on the control so the UI's
-            # _sync_sources watcher fires and the SourceCatalog
-            # ("Available Sources") is updated.  When invoked
-            # via the Tabulator click path this happens inside
-            # _run_load → _handle_success; the SourceAgent path
-            # bypasses _run_load so we do it explicitly here.
-            if result and result.sources:
-                for src in result.sources:
-                    self._register_source_output(src)
-                self.param.trigger("outputs")
+            # The SourceAgent path bypasses _run_load, so apply the same output
+            # registration and event handling as the Tabulator click path.
+            self._handle_success(result)
             return result
         finally:
             # _fetch_entry may update the progress bar (e.g.
@@ -327,8 +322,8 @@ class CatalogSourceControls(BaseSourceControls):
         self._cached_catalog_tools = tools
         return tools
 
-    async def _search_catalog(self, query: str) -> object | None:
-        """Find the best matching catalog row index label for *query*.
+    async def _search_catalog(self, query: str) -> int | None:
+        """Find the best matching catalog row position for *query*.
 
         Uses vector search when a ``vector_store`` is available and
         falls back to keyword matching on ``search_columns``.
@@ -356,11 +351,11 @@ class CatalogSourceControls(BaseSourceControls):
         best_idx = None
         best_score = 0
 
-        for idx, row in self.catalog_df.iterrows():
+        for pos, (_, row) in enumerate(self.catalog_df.iterrows()):
             text = self._entry_to_text(row).lower()
             score = sum(1 for word in query_words if word in text)
             if score > best_score:
                 best_score = score
-                best_idx = idx
+                best_idx = pos
 
         return best_idx if best_score > 0 else None
--- a/lumen/ai/controls/ingest/catalog.py
+++ b/lumen/ai/controls/ingest/catalog.py
@@ -245,9 +245,12 @@ class CatalogSourceControls(BaseSourceControls):
 
             # Structured metadata for post-semantic filtering
             # VectorStore metadata values must be str/int/float/bool
+            row_idx = idx.item() if hasattr(idx, "item") else idx
+            if not isinstance(row_idx, (str, int, float, bool)):
+                row_idx = str(row_idx)
             metadata = {
                 "type": "catalog_entry",
-                "_row_idx": int(idx) if not isinstance(idx, int) else idx,
+                "_row_idx": row_idx,
                 "_control_id": id(self),
             }
             for col in self.filter_columns:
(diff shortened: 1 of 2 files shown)


lumen/ai/controls/ingest/catalog.py
```diff
@@ -245,9 +245,12 @@ async def _embed(self):
 
             # Structured metadata for post-semantic filtering
             # VectorStore metadata values must be str/int/float/bool
+            row_idx = idx.item() if hasattr(idx, "item") else idx
+            if not isinstance(row_idx, (str, int, float, bool)):
+                row_idx = str(row_idx)
```
ghostiee-11: could we store the position here instead of the label? `for pos, (idx, row) in enumerate(self.catalog_df.iterrows())` with `"_row_idx": pos` is always a JSON-safe int, so both normalization lines go away and `.iloc` keeps working for every index type.

as-is a MultiIndex round-trips as the string `"('s', 2)"` and the `.loc` lookup on it raises `KeyError` (checked on this branch). same for a UUID or bytes index; datetime only survives because pandas re-parses the string form.

lumen/ai/controls/ingest/catalog.py
```diff
@@ -288,7 +291,7 @@ async def _load_from_query(self, query: str) -> SourceResult:
                 f"No dataset matching '{query}' found in catalog."
             )
 
-        entry = self.catalog_df.iloc[match_idx]
+        entry = self.catalog_df.loc[match_idx]
```
ghostiee-11: `.loc` breaks on duplicate labels, which is the concatenated case #2024 names. with `pd.concat([df_a, df_b])` the index is `[0, 1, 0, 1]`, so `.loc[1]` hands back a 2-row DataFrame, `entry["name"]` is a Series, and the load fails. keeping `.iloc` with a positional `_row_idx` avoids it.

also this line sits outside the `try`, so a stale index raises `KeyError` out of the tool instead of returning `SourceResult.empty(...)` like the check right above it, and skips the `finally` cleanup.

lumen/ai/controls/ingest/catalog.py
```diff
@@ -298,10 +301,8 @@ async def _load_from_query(self, query: str) -> SourceResult:
             # _run_load → _handle_success; the SourceAgent path
             # bypasses _run_load so we do it explicitly here.
             if result and result.sources:
-                existing = self.context.get("sources", [])
                 for src in result.sources:
-                    if src not in existing:
-                        self._register_source_output(src)
+                    self._register_source_output(src)
                 self.param.trigger("outputs")
```
ghostiee-11: could this just be `self._handle_success(result)`? it already registers every source, and additionally sets `outputs["table"]` / `_last_table` and triggers `upload_successful`.

without those, `ui.py:1599` copies the previous `outputs["table"]` back into the global context, so the UI keeps the old dataset selected next to the newly loaded source.

lumen/tests/ai/test_controls/test_catalog_source_controls.py
```diff
@@ -0,0 +1,100 @@
...
+    async def _load_catalog(self):
+        return pd.DataFrame()
+
+    async def _fetch_entry(self, entry):
+        return SourceResult(table=entry["name"], message=entry["name"])
+
+
+@pytest.mark.asyncio
+async def test_load_from_query_uses_noncontiguous_integer_index_label():
+    controls = _CatalogControls(context={})
```
ghostiee-11: `__init__` schedules `_load_catalog_wrapper` via `pn.state.onload`, so this assignment gets overwritten by `_load_catalog`'s empty frame the moment the loop yields. adding a single `await asyncio.sleep(0)` before the call makes all three tests fail with `assert None == 'target'`.

returning the fixture frame from `_load_catalog` instead of assigning `catalog_df` after construction would hold up. a duplicate-label case would be worth adding here too.

ghostiee-11 (commented):
Thanks for picking this up. Ran the cases from the issue against this branch and the label switch fixes the non-contiguous index but reintroduces a failure on the concatenated frame the issue names, so left a few notes inline.

lumen/tests/ai/test_controls/test_catalog_source_controls.py
fly1d: Updated in 3bbfe69. The test subclass now owns the fixture frame and returns it from `_load_catalog`; `_make_controls` yields once so the scheduled `pn.state.onload` callback completes before each test proceeds. The duplicate-label case is included. The full unit suite passes with 2040 passed and 42 skipped.

lumen/ai/controls/ingest/catalog.py
fly1d: Updated in 3bbfe69. The agent path now calls `self._handle_success(result)`. The regression checks source deduplication plus `outputs["table"]` and `_last_table`, covering the state that the UI copies into the global context.

lumen/ai/controls/ingest/catalog.py
fly1d: Updated in 3bbfe69. Lookup remains positional with `.iloc`; the match is validated inside the `try`/`finally`, and stale or malformed positions return `SourceResult.empty(...)` while progress is cleared. I added both the `pd.concat` duplicate-label case and a stale vector-position regression.

lumen/ai/controls/ingest/catalog.py
fly1d: Updated in 3bbfe69. `_embed()` now enumerates the frame and stores the Python integer row position directly, so label normalization is gone. The vector-path regression now uses a MultiIndex and asserts metadata `[0, 1]` before loading position 1.

ghostiee-11:
@fly1d can you use https://github.com/holoviz-dev/holoviz-skills/blob/main/contributing-to-holoviz/skills/pr-description/SKILL.md, for drafting PR desc

fly1d:
Updated the title and PR description using the linked HoloViz skill and current organization template. The description now uses first-person causal prose, includes a copy-pastable duplicate-index reproducer with Before/After behavior, lists the exact test results, and keeps the complete AI disclosure and responsibility checklist.

lumen/ai/controls/ingest/catalog.py
```diff
@@ -283,26 +283,22 @@ async def _load_from_query(self, query: str) -> SourceResult:
...
+            if (
+                not isinstance(match_idx, int)
+                or match_idx < 0
+                or match_idx >= len(self.catalog_df)
+            ):
+                return SourceResult.empty(no_match)
+            try:
+                entry = self.catalog_df.iloc[match_idx]
+            except (IndexError, TypeError):
+                return SourceResult.empty(no_match)
```
ghostiee-11: nit: the bounds check right above already guarantees `match_idx` is an int in `[0, len)`, so `.iloc` can't raise `IndexError` or `TypeError` here and this inner try never fires. probably worth keeping just the bounds check.

ghostiee-11 (approved):

PR #2033, opened 2026-08-19
fix: allow all tabular hvPlot kinds

### LINKED ISSUE #1940: hvPlotView.kind still blocks four kinds hvPlot supports on DataFrames, including vectorfield
`hvPlotBaseView.kind` hardcodes its list of accepted kinds (`lumen/views/base.py:927-935`), and that list has drifted from what hvPlot actually accepts for a DataFrame. It has already been patched reactively twice: #1930 added `paths` and `polygons` after geometry columns silently fell back to points, and #1823 adds `quadmesh` for gridded data. Four more kinds are still blocked.

`hvPlotTabular.__all__` is the authoritative set for tabular data and holds 24 entries. Setting aside `explorer`, which `hvPlotUIView` covers, these are missing from Lumen's list:

- `vectorfield`
- `density`
- `table`
- `dataset`

All four work on a plain DataFrame today:

```python
df.hvplot(kind="vectorfield", x="x", y="y", angle="u", mag="v")  # OK
df.hvplot(kind="density", y="y")                                 # OK
df.hvplot(kind="table")                                          # OK
df.hvplot(kind="dataset")                                        # OK
```

but each one raises out of Lumen before hvPlot is ever reached:

```python
hvPlotView(kind="vectorfield", pipeline=pipeline)
# ValueError: Selector parameter 'hvPlotBaseView.kind' does not accept 'vectorfield'
```

`vectorfield` is the one real gap. It is a plot type with no equivalent anywhere else in Lumen and it is the standard way to show wind or current data, which is exactly what the xarray sources bring in, so it sits right next to the gridded work in #1823. `density` is an alias for `kde` so it is a convenience only, and `table` and `dataset` overlap with the `Table` view and with `hv.Dataset`, so those three are mostly about not surprising someone who knows hvPlot's API. (`rgb` is also absent, and unlike the four above it needs the gridded conversion #1823 introduces, so it belongs with that work rather than here.)

This constrains the LLM as much as YAML authors. As #1930 spelled out, `param_to_pydantic` maps a `Selector` to `Literal[tuple(parameter.objects)]` (`lumen/ai/translate.py:261-271`), so the list is literally the enum the model chooses from. A user asking for a vector field gets something else, with no way to express the request.

Adding the four strings is the narrow fix. Given this is now the third round of the same problem, the more durable one is deriving the objects list from hvPlot (`hvPlotTabular.__all__`, minus `explorer`) instead of restating it, so it cannot drift again on the next hvPlot release.

fly1d:
## Description

This PR derives `hvPlotView.kind` choices from hvPlot's public `hvPlotTabular.__all__`, while retaining the gridded kinds that Lumen handles through xarray conversion. The Selector now accepts `vectorfield`, `density`, and `table`, so valid DataFrame plots reach hvPlot instead of being rejected by Lumen; `explorer` remains reserved for `hvPlotUIView`, and `dataset` is excluded because it returns a bare `hv.Dataset` with no plotting class for `hvPlotView`.

```python
import pandas as pd
from lumen.sources.base import InMemorySource
from lumen.views.base import hvPlotView

df = pd.DataFrame({
    "x": [0.0, 1.0], "y": [1.0, 2.0],
    "angle": [0.0, 0.5], "mag": [1.0, 2.0],
})
pipeline = Pipeline(source=InMemorySource(tables={"data": df}), table="data")
hvPlotView(
    pipeline=pipeline, kind="vectorfield", x="x", y="y",
    angle="angle", mag="mag",
).get_plot(df)
```

Before this change, constructing the view raises `ValueError: Selector parameter 'hvPlotBaseView.kind' does not accept 'vectorfield'`; after it, the call reaches hvPlot and returns a `VectorField`. `dataset` remains rejected at construction because allowing it would defer the failure to rendering as `SkipRendering: No plotting class for Dataset found.` The change adds no dependencies, migration steps, or breaking behavior.

I ran `pixi run -e test-312 pytest lumen/tests/views/test_hvplot_gridded.py -q` (20 passed), `pixi run -e test-312 test-unit` (2042 passed, 42 skipped), and the targeted pre-commit hooks (all passed). Direct smoke checks confirmed `density`, `table`, and `vectorfield` render through `hvPlotView`, while the regression test verifies that `dataset` and `explorer` remain excluded.

Fixes #1940

## AI Disclosure

Tool & Model: OpenAI Codex + GPT-5
Usage: I used Codex to inspect issue #1940 and the existing hvPlot view, implement the narrow Selector synchronization and regression test, incorporate maintainer review feedback about unsupported `dataset` rendering, run the focused and full test suites, and draft this description. I reviewed the resulting diff, verification, and disclosure before submission.

## Checklist

- [ ] Added documentation

### DIFF
--- a/lumen/tests/views/test_hvplot_gridded.py
+++ b/lumen/tests/views/test_hvplot_gridded.py
@@ -5,10 +5,12 @@ import numpy as np
 import pandas as pd
 import pytest
 
+from hvplot import hvPlotTabular
+
 from lumen.pipeline import Pipeline
 from lumen.sources.base import InMemorySource
 from lumen.views import base as views_base
-from lumen.views.base import hvPlotUIView, hvPlotView
+from lumen.views.base import GRIDDED_KINDS, hvPlotUIView, hvPlotView
 
 # ---- Fixtures ----
 
@@ -58,6 +60,16 @@ def test_hvplot_z_param_accepted():
 def test_hvplot_quadmesh_in_kind_objects():
     """quadmesh is a recognised kind value."""
     assert "quadmesh" in hvPlotView.param.kind.objects
+
+
+def test_hvplot_tabular_kinds_stay_in_sync_with_hvplot():
+    """All tabular hvPlot kinds are accepted except the UI explorer."""
+    expected = {
+        kind for kind in hvPlotTabular.__all__ if kind not in {"explorer", "dataset"}
+    }
+    assert set(hvPlotView.param.kind.objects) == expected | set(GRIDDED_KINDS)
+    assert "explorer" not in hvPlotView.param.kind.objects
+    assert "dataset" not in hvPlotView.param.kind.objects
 
 
 def test_hvplot_heatmap_from_longform_df(gridded_pipeline):
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -81,10 +81,9 @@ REDUCING_KINDS = GRIDDED_KINDS + (
 # Keep the Selector in sync with tabular kinds that produce plots in hvPlot.
 # The explorer is handled by hvPlotUIView, while dataset returns a bare
 # hv.Dataset with no plotting class for hvPlotView to render.
-HVPLOT_TABULAR_KINDS = [
+HVPLOT_KINDS = [
     kind for kind in hvPlotTabular.__all__ if kind not in {"explorer", "dataset"}
-]
-HVPLOT_KINDS = HVPLOT_TABULAR_KINDS + list(GRIDDED_KINDS)
+] + list(GRIDDED_KINDS)
 
 
 class View(MultiTypeComponent, Viewer):
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -21,6 +21,7 @@ import panel as pn
 import param  # type: ignore
 
 from bokeh.models import NumeralTickFormatter  # type: ignore
+from hvplot import hvPlotTabular  # type: ignore
 from panel.io.document import immediate_dispatch
 from panel.pane.base import PaneBase
 from panel.pane.holoviews import HoloViews as HoloViewsPane
@@ -76,6 +77,14 @@ MAX_RENDER_ROWS = 250_000
 REDUCING_KINDS = GRIDDED_KINDS + (
     "heatmap", "hexbin", "hist", "kde", "box", "violin", "bivariate",
 )
+
+# Keep the Selector in sync with tabular kinds that produce plots in hvPlot.
+# The explorer is handled by hvPlotUIView, while dataset returns a bare
+# hv.Dataset with no plotting class for hvPlotView to render.
+HVPLOT_TABULAR_KINDS = [
+    kind for kind in hvPlotTabular.__all__ if kind not in {"explorer", "dataset"}
+]
+HVPLOT_KINDS = HVPLOT_TABULAR_KINDS + list(GRIDDED_KINDS)
 
 
 class View(MultiTypeComponent, Viewer):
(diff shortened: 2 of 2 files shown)


lumen/tests/views/test_hvplot_gridded.py
```diff
@@ -60,6 +60,17 @@ def test_hvplot_quadmesh_in_kind_objects():
     assert "quadmesh" in hvPlotView.param.kind.objects
 
 
+def test_hvplot_tabular_kinds_stay_in_sync_with_hvplot():
+    """All tabular hvPlot kinds are accepted except the UI explorer."""
+    from hvplot import hvPlotTabular
```
ghostiee-11: imports on top

lumen/views/base.py
```diff
@@ -77,6 +78,10 @@
     "heatmap", "hexbin", "hist", "kde", "box", "violin", "bivariate",
 )
 
+# Keep the Selector in sync with the tabular kinds accepted by hvPlot. The
+# explorer is handled by hvPlotUIView rather than hvPlotView.
+HVPLOT_TABULAR_KINDS = [kind for kind in hvPlotTabular.__all__ if kind != "explorer"]
```
ghostiee-11: `dataset` needs excluding alongside `explorer`. It returns a bare `hv.Dataset`, which has no plotting class, so `hvPlotView(pipeline=pl, kind="dataset").get_panel()` raises `SkipRendering: No plotting class for Dataset found.` Right now that's a clear ValueError at construction; after this it's a crash at render, and `param_to_pydantic` offers it to the LLM as a valid choice.

lumen/views/base.py
```diff
@@ -950,12 +955,7 @@ class hvPlotBaseView(View):
 
     kind = param.Selector(
         default=None, doc="The kind of plot, e.g. 'scatter' or 'line'.",
-        objects=[
-            'area', 'bar', 'barh', 'bivariate', 'box', 'contour', 'contourf',
-            'errorbars', 'hist', 'image', 'kde', 'labels',
-            'line', 'scatter', 'heatmap', 'hexbin', 'ohlc', 'paths', 'points',
-            'polygons', 'quadmesh', 'step', 'violin'
-        ]
+        objects=HVPLOT_TABULAR_KINDS + ['contour', 'contourf', 'image', 'quadmesh']
```
ghostiee-11: `GRIDDED_KINDS` a few lines up is already this exact tuple: `HVPLOT_TABULAR_KINDS + list(GRIDDED_KINDS)`? Same in the test's `expected`, which spells them out a third time.

fly1d: Addressed by introducing  as the single combined list from the supported tabular kinds and . The test now reuses  instead of spelling out the gridded names.

lumen/views/base.py
fly1d: Addressed by excluding  alongside .  now keeps the existing construction-time rejection instead of offering a kind that fails later with  because bare  has no plotting class. The regression test covers both exclusions.

lumen/tests/views/test_hvplot_gridded.py
fly1d: Moved  to the module-level imports. Thanks.

lumen/views/base.py
fly1d: Correction: the updated code introduces HVPLOT_KINDS as the combined list of supported tabular kinds and GRIDDED_KINDS. The test reuses GRIDDED_KINDS instead of repeating the gridded names.

lumen/tests/views/test_hvplot_gridded.py
fly1d: Correction: the hvPlotTabular import is now at module scope.

lumen/views/base.py
fly1d: Correction: the updated code excludes dataset and explorer. hvPlotView keeps rejecting dataset during construction because a bare hv.Dataset has no plotting class; the regression test covers both exclusions.

ghostiee-11 (approved):

ghostiee-11:
LGTM Thanks @fly1d for the fix!! :)

PR #2034, opened 2026-08-19
fix: stop awaiting the async generator that streams view specs

ghostiee-11:
`_stream_prompt` is an async generator function, so awaiting its return value raises TypeError and every view the agent generates fails after exhausting its retries. Every other call site iterates it directly; the two retry tests hid this by mocking it as a coroutine, so those mocks now mirror the real signature.


ahuang11 (approved):

ahuang11:
Nice catch!

ghostiee-11:
Hehe Thanks!!

PR #2035, opened 2026-08-19
feat: expose datashader categorical blending on hvPlot views

ghostiee-11:
Coloring a dense scatter by a category is a datashader job, and hvPlot derives the count_cat aggregator from `by` whenever datashade is set, but the keywords were only reachable as loose kwargs and so invisible to the agent's schema; this promotes datashade, dynspread and color_key to params and points the large-frame default at datashade rather than rasterize when `by` is set. It also carries the `param_to_pydantic` fix needed to build the agent's response model at all, which overlaps #1935 and #1931 by a different route (fixing the recursion rather than replacing it with an explicit `create_model`), so those should be reconciled before either lands.

Colors are best with holoviz/hvplot#1752, but degrade to the default palette without it.

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1108,7 +1108,11 @@ class hvPlotUIView(hvPlotBaseView):
             if any(k in control for control in controls)
             and v is not None and k != 'name'
         }
-        if self.color_key is not None:
+        # Only completed once a control has claimed it above: hvPlot gained the
+        # color_key control after this was written, and forcing the keyword in
+        # regardless makes hvPlotExplorer.__init__ reject it outright on an
+        # older hvPlot rather than simply coloring from the default palette.
+        if 'color_key' in params:
             params['color_key'] = self._complete_color_key(data)
         return (data,), dict(params, **self.kwargs)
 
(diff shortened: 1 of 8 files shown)


lumen/views/base.py
```diff
@@ -967,6 +968,20 @@ class hvPlotBaseView(View):
...
 
+    color_key = param.Dict(default=None, doc="""
+        Mapping of the values in `by` to explicit colors, e.g.
+        {'Irish': '#e41a1c', 'Italian': '#377eb8'}. Only meaningful with
+        datashade; without it datashader picks a categorical palette.""")
+
+    datashade = param.Boolean(default=False, doc="""
+        Aggregate the data server-side with datashader and send an image
+        instead of one glyph per row. Combined with `by` this blends the
+        categories present in each pixel, rather than overplotting them.""")
```
ahuang11, line 997: Should we introduce aggregator too https://hvplot.holoviz.org/en/docs/latest/ref/plotting_options/resampling.html#aggregator

ghostiee-11: Yup.. i think it will be good addition :)

Thanks for pointing it out...

ahuang11 (approved):

PR #2039, opened 2026-08-19
Collect a lazy frame before it reaches Pipeline.data

ghostiee-11:
A `Source` that skips the cache can return a lazy frame, and nothing collected it, so `Pipeline.data` ended up holding a `LazyFrame` and every consumer that calls `len`, `.iloc` or `.index` on it broke. It is now collected once after the last transform.

### DIFF
--- a/docs/configuration/spec/customization.md
+++ b/docs/configuration/spec/customization.md
@@ -216,6 +216,12 @@ below run on it natively. Things to know before choosing one:
   Vega only read pandas. Returning polars saves work up to the view, not through
   it.
 
+A `Source` that skips the cache may return a lazy frame, such as a polars
+`LazyFrame` from `scan_parquet`. The Pipeline collects it once after the last
+transform, so `Pipeline.data` is eager as always. The built-in Sources collect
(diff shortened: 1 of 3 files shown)


ghostiee-11:
Numbers on a 1M row, 21 column parquet (20 numeric plus 1 string) read with polars `scan_parquet`, medians over repeated runs:

| case | before | after |
| --- | --- | --- |
| `columns=` 2 of the 21 | 41-69 ms, 197-221 MB | 7.4-7.6 ms, 58-64 MB |
| all 21 columns | 67 ms | 81 ms |
| 20 numeric, no string column | 97 ms | 93 ms |

The win is the column subset, where the old code collected every column to describe the two that were asked for. Describing the whole frame is a wash, and slightly slower when a string column forces the second scan.

Peak RSS for the full frame case is unchanged, 141-185 MB before against 153-208 MB after, which is noise in both directions. Not holding the frame does not lower the peak, because polars reads the data to compute the aggregates either way. The saving is in what gets read, not in what gets held.

ahuang11 (approved):

PR #2044, opened 2026-08-19
Sample the frame before converting it for the AI summary

### LINKED ISSUE #2010: Port the AI schema summary to narwhals to avoid materializing the full frame
`ai/utils.py:634` converts the whole frame to pandas, then `describe_data_sync` samples 5000 rows from it at line 796. A 10M row polars frame is fully materialized to summarize 5000 rows.

Porting the summarizer to narwhals would let it sample first and convert 5000 rows instead. `describe_data_sync` is about 150 lines of pandas only API (`describe`, `select_dtypes`, `to_markdown`, `nunique`, `.str.len()`).

Follow up to #2008, which only puts the conversion boundary in place.

ghostiee-11:
The schema summary converted the whole frame to pandas and then read 5000 rows off it, so a large polars frame was materialised just to describe a sample of it. I moved the sampling above the conversion, so only the sample is converted and the summary comes out byte-identical.

| | before | after |
| --- | --- | --- |
| `describe_data_sync`, 2M x 15 polars frame | 0.369s | 0.091s |
| process peak RSS | 1319MB | 889MB |
| rows converted by `LumenEditor.render_context` | 200000 | 5000 |

A dask frame raised `TypeError` here before, since its shape comes back as a `Delayed` that none of the branches could act on. It is now computed and summarised.

Fixes #2010

### DIFF
--- a/lumen/ai/editors.py
+++ b/lumen/ai/editors.py
@@ -32,14 +32,13 @@ from ..config import dump_yaml, load_yaml
 from ..filters import WidgetFilter
 from ..pipeline import Pipeline
 from ..transforms.sql import SQLLimit
-from ..util import as_pandas
 from ..views.base import Panel, Table, View
 from .analysis import Analysis
 from .config import FORMAT_ICONS, FORMAT_LABELS
 from .controls import (
     AnnotationControls, CopyControls, ExplainControls, RetryControls,
 )
-from .utils import describe_data, get_data
+from .utils import describe_data, get_frame
 
 if TYPE_CHECKING:
     from panel.chat.feed import ChatFeed
@@ -226,8 +225,7 @@ class LumenEditor(Viewer):
         else:
             sql_limit = None
         if sql_limit:
-            data = as_pandas(pipeline.data)
-            limited = len(data) == sql_limit.limit
+            limited = len(pipeline.data) == sql_limit.limit
             if limited:
                 def unlimit(e):
                     sql_limit.limit = None if e.new else 1_000_000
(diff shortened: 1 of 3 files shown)


ahuang11:
If we sample, min and max will be off right?

ahuang11:
I'm starting to wonder whether narwhals or duckdb should be at the core. duckdb supports all dataframes right? I guess my question is why do we convert back and forth from duckdb to pandas/narwhals etc. Why don't we just use duckdb for everything?

philippjfr (approved):

philippjfr:
> If we sample, min and max will be off right?

My understanding was that we already sampled.

ghostiee-11:
yes that's already the case on main, this PR doesn't change it. describe_data_sync always sampled before `df.describe()`:

```

10000 x 2 frame on main
stats.i -> {'count': 5000, 'min': 1, 'max': 9998}
true       min 0, max 9999
```
this just moves the sampling above the pandas conversion so the same 5000 rows get profiled without materialising the rest.

ghostiee-11:
on duckdb: it does read pandas/polars/arrow directly, checked all three. but it's an engine not a dispatch layer so the data has to come back out, and hvplot/tabulator/perspective/vega all still want pandas and hvplot hasn't moved to narwhals yet. so the conversion doesn't go away, it moves to the view boundary. Also loses the source's own pushdown, a polars scan or remote warehouse would go through duckdb instead of planning itself. Narwhals costs nothing at runtime, it only dispatches.

ahuang11:
hvplot supports duckdb afaik.

ahuang11 (approved):

PR #2046, opened 2026-08-20
Keep integer and boolean dtypes across the pandas fallback

### LINKED ISSUE #2011: Transform fallback to pandas does not preserve nullable integer dtypes
When a transform config has no narwhals equivalent the frame converts to pandas, and `to_pandas()` turns a polars nullable `Int64` into `float64`.

So two adjacent configs on the same transform return different schemas:

```python
DropNA()             # stays polars, id is Int64
DropNA(how='all')    # converts, id becomes float64 with NaN
```

IDs then render as `3.0` and downstream joins or `astype` break. Known and documented in #2008, not fixed there.

ghostiee-11:
Two configurations of one transform returned different schemas because the pandas fallback widened a nullable integer to float64, so an id rendered as 3.0 and, past 2**53, came back as a different number entirely. Plots still take the numpy widening, because datashader rejects a pandas extension dtype.

Fixes #2011

### DIFF
--- a/lumen/util.py
+++ b/lumen/util.py
@@ -21,6 +21,7 @@ import numpy as np
 import pandas as pd
 import panel as pn
 import param
+import pyarrow as pa
 import yaml
 
 from jinja2 import DebugUndefined, Environment, Undefined
@@ -109,6 +110,41 @@ def collect_lazy(df):
 
 DATAFRAME_BACKENDS = ['pandas', 'polars', 'pyarrow']
 
+# The pandas dtype that holds each arrow integer and boolean type without
+# widening it. numpy has no missing value for either, which is the whole
+# problem _to_pandas below exists to solve.
+_NULLABLE_DTYPES = {
+    getattr(pa, f'{prefix}int{width}')(): pd.api.types.pandas_dtype(
+        f'{prefix.upper()}Int{width}'
+    )
+    for prefix in ('', 'u') for width in (8, 16, 32, 64)
+}
+_NULLABLE_DTYPES[pa.bool_()] = pd.BooleanDtype()
+
+
+def _to_pandas(narwhals_df):
+    """Convert to pandas without widening an integer or boolean column.
+
+    Every backend but pandas holds nulls in the column's own type, and
+    ``to_pandas`` lands those on numpy, which has no missing value for an
+    integer or a boolean. An id comes back as 3.0 and, past 2**53, as a
+    different number entirely. The columns that would widen come across
+    through arrow as the pandas nullable dtype instead.
+
+    Only those columns are rerouted. Sending the whole frame through arrow
+    would be shorter and would break more: pyarrow cannot convert the
+    dictionary indices behind a polars categorical, which polars itself
+    converts fine.
+    """
+    df = narwhals_df.to_pandas()
+    for name, dtype in narwhals_df.collect_schema().items():
+        widened = df[name].dtype.kind in 'fO'
+        if widened and (dtype.is_integer() or isinstance(dtype, nw.Boolean)):
+            df[name] = narwhals_df[name].to_arrow().to_pandas(
+                types_mapper=_NULLABLE_DTYPES.get
+            )
+    return df
+
 
 def to_backend(df, backend):
     """Return df as a frame of the named dataframe library.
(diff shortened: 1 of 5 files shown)


lumen/util.py
```diff
@@ -109,6 +110,90 @@ def collect_lazy(df):
...
+    different number entirely. The columns that would widen come across
+    through arrow as the pandas nullable dtype instead.
+
+    Only those columns are rerouted. Sending the whole frame through arrow
+    would be shorter and would break more: pyarrow cannot convert the
+    dictionary indices behind a polars categorical, which polars itself
+    converts fine.
+    """
+    narwhals_df = _cast_unsigned_dictionaries(narwhals_df)
+    df = narwhals_df.to_pandas()
```
ahuang11, line 167: does to_pandas have any kwargs to do this?

ghostiee-11: no, narwhals `to_pandas()` takes no args at all (2.17.0). polars' native one does take `types_mapper`, but then every int column comes back nullable even when it has no nulls, and going through arrow on top of that turns a polars Date into object. the loop keeps it to the columns that actually widened.

lumen/util.py
```diff
@@ -109,6 +110,90 @@ def collect_lazy(df):
...
+# The pandas dtype that holds each arrow integer and boolean type without
+# widening it, because numpy has no missing value for either. Arrow has no
+# other integer type, so every column _to_pandas reroutes has an entry here.
+_NULLABLE_DTYPES = {
+    pa.int8(): pd.Int8Dtype(), pa.int16(): pd.Int16Dtype(),
+    pa.int32(): pd.Int32Dtype(), pa.int64(): pd.Int64Dtype(),
+    pa.uint8(): pd.UInt8Dtype(), pa.uint16(): pd.UInt16Dtype(),
+    pa.uint32(): pd.UInt32Dtype(), pa.uint64(): pd.UInt64Dtype(),
+    pa.bool_(): pd.BooleanDtype(),
+}
```
ahuang11, line 122: Not sure if this is better:

```
1. Map by Arrow type ID

If this is performance-sensitive and these are specifically the supported primitive types, I'd probably use pa.Type IDs:

_NULLABLE_DTYPES = {
    pa.Type_INT8: pd.Int8Dtype(),
    pa.Type_INT16: pd.Int16Dtype(),
    pa.Type_INT32: pd.Int32Dtype(),
    pa.Type_INT64: pd.Int64Dtype(),
    pa.Type_UINT8: pd.UInt8Dtype(),
    pa.Type_UINT16: pd.UInt16Dtype(),
    pa.Type_UINT32: pd.UInt32Dtype(),
    pa.Type_UINT64: pd.UInt64Dtype(),
    pa.Type_BOOL: pd.BooleanDtype(),
}

Then:

dtype = _NULLABLE_DTYPES.get(arrow_type.id)

This makes the intent clearer: you're dispatching on the Arrow type, rather than relying on equality/hash behavior of DataType objects.

2. Generate the mapping instead of spelling it out

If the relationship is systematically "Arrow bit width → pandas nullable dtype", you could reduce the repetition:

_NULLABLE_DTYPES = {
    getattr(pa, f"Type_{kind.upper()}{bits}"): getattr(pd, f"{kind.upper()}{bits}Dtype")()
    for kind in ("int", "uint")
    for bits in (8, 16, 32, 64)
}
_NULLABLE_DTYPES[pa.Type_BOOL] = pd.BooleanDtype()

I wouldn't actually recommend this, though. It's cleverer than the original and makes static inspection/type checking harder.
```

ahuang11 (approved):

lumen/util.py
ghostiee-11: `pa.Type_INT64` isn't public, only `pa.lib.Type_INT64`, so both snippets raise AttributeError at import. And `types_mapper` gets the `DataType` itself, not an id, so an id-keyed dict would need a lambda instead of the bare `.get`. pyarrow's docs for that param say to pass `dict.get`, so I'll leave it as is.

ahuang11: Okay fine to merge

ghostiee-11: Okay, Thanks :)

PR #2049, opened 2026-08-20
Decline to grid a to_dataset result that is not a grid

### LINKED ISSUE #2048: to_dataset fails on the pivot fallback when a sql_transform projects a dim away
`XArraySQLSource.to_dataset` falls back to pivoting the long-form result when the installed xarray-sql has no native `to_dataset`. A projecting `sql_transform` drops a dim from the result, and the rows that dim separated
then collapse onto the same index, so the pivot raises:

```
ValueError: cannot convert a DataFrame with a non-unique MultiIndex into xarray
```

`lumen/tests/sources/test_xarray_sql.py::TestToDataset::test_to_dataset_ignores_projected_away_dim` covers this and fails whenever the fallback branch runs. The native path handles it, so whether the test passes depends on the installed xarray-sql
version rather than on lumen.

The fallback has to decide what a projected-away dim means before it can pivot: aggregate the duplicate rows, or refuse and say so. Right now it does neither and fails inside xarray.

ghostiee-11:
A projecting sql_transform that drops a dim leaves the rows that dim separated on top of each other, and the pivot fallback handed those to xarray, which raised. Callers already fall back to the long-form frame when there is no grid, so this returns None instead.

Fixes #2048

### DIFF
--- a/lumen/sources/xarray_sql.py
+++ b/lumen/sources/xarray_sql.py
@@ -276,34 +276,20 @@ class XArraySQLSource(BaseSQLSource):
         return self.execute(self._build_sql(table, **query))
 
     def to_dataset(self, table, **query):
-        """Return the query result as a gridded ``xr.Dataset``, or None.
+        """Return the query result as a gridded ``xr.Dataset``.
 
-        Uses xarray-sql's ``to_dataset`` (lazy, keeps the data gridded) when
-        the installed version supports it, so a gridded source is never
-        materialized as long-form. Falls back to pivoting the long-form result
-        on the dataset dims for xarray-sql < 0.3.
-
-        None means the result is not a grid, and callers fall back to the
-        long-form frame.
+        Uses xarray-sql's ``to_dataset``, which is lazy and keeps the data
+        gridded, so a gridded source is never materialized as long-form.
         """
         result = self._ctx.sql(self._build_sql(table, **query))
-        if hasattr(result, 'to_dataset'):
-            # Pass only the dims the result actually has a column for. Multiple
-            # data vars need dims passed explicitly (the FROM clause is
-            # ambiguous), but a bounds dim (e.g. nbnds) or a projecting/
-            # aggregating sql_transform can leave a dataset dim out of the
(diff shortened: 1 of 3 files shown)


ahuang11 (approved):

lumen/sources/xarray_sql.py
```diff
@@ -295,7 +298,12 @@ def to_dataset(self, table, **query):
             return result.to_dataset(dims=dims)
         df = result.to_pandas()
         present = [d for d in self._dataset.dims if d in df.columns]
-        return df.set_index(present).to_xarray() if present else df
+        if not present or df.duplicated(subset=present).any():
+            # A projecting sql_transform that drops a dim leaves the rows that
+            # dim separated on top of each other, so the surviving dims no
+            # longer index one row each and there is no grid to build.
+            return None
+        return df.set_index(present).to_xarray()
```
ahuang11: Can't we just require a newer xarray-sql version?

ahuang11 (changes requested):

lumen/sources/xarray_sql.py
ghostiee-11: `pyproject` already pins `xarray-sql >=0.3.1`, so that fallback is dead code and I'm fine deleting it. It only still runs because `pixi.toml` leaves it unpinned and the lock resolves 0.2.2.

A newer version doesn't fix the case though. On 0.3.3 and 0.4.0rc1 the native `to_dataset` takes the same projected query and silently keeps the last row per (lat, lon): 200 rows in, 20 cells out, no error. Filed upstream as xqlsystems/xarray-sql#247.

ahuang11: Please fix the dead code

ghostiee-11: Suree

ghostiee-11:
@ahuang11, ig now can me merged

ahuang11 (approved):

ghostiee-11:
Thanks @ahuang11 :)

PR #2052, opened 2026-08-21
Fix/issue 1895

### LINKED ISSUE #1895: Clean up requirements in lumen
I think there are unnecessary dependencies that can be moved to optional, e.g. https://github.com/holoviz/lumen/blob/main/pyproject.toml#L37

Vamshi1404:
## Description

Clean up requirements in lumen by moving 17 dependencies from core to
optional extras, reducing the base install footprint from 29 to 12 core
dependencies.

### Dependency classification

| Dependency | Moved to | Evidence |
|---|---|---|
| `chardet` | `tests` extra | try/except fallback at `lumen/util.py:644` degrades to `'latin-1'`; used by `lumen/tests/transforms/test_sql.py:137` via `pytest.importorskip` |
| `duckdb` | new `duckdb` extra | only imported in `lumen/sources/duckdb.py:9` and `lumen/ai/vector_store.py:15` |
| `griffe` | new `ai` extra | only imported in `lumen/ai/translate.py:17` |
| `instructor` | new `ai` extra | only imported in `lumen/ai/{config,llm,utils}.py` |
| `intake` | new `intake` extra | only imported in `lumen/sources/intake.py:5` |
| `lxml` | **removed** | zero imports in `lumen/` source; already in `tests` extra |
| `markitdown` | new `ai` extra | only imported in `lumen/ai/{config,vector_store}.py` inside try/except |
| `nbformat` | new `ai` extra | only imported in `lumen/ai/export.py:9` |
| `panel-graphic-walker` | new `graphic-walker` extra | not imported at module load time; only string ref `'graphic_walker'` in `lumen/views/base.py:1758` |
| `panel-splitjs` | new `ai` extra | only imported in `lumen/ai/ui.py:32` |
| `pyarrow` | **removed** | zero direct imports; pandas uses it transparently as a parquet/CSV backend |
| `pydantic` | new `ai` extra | only imported in `lumen/ai/` |
| `pydantic-extra-types` | new `ai` extra | only imported in `lumen/ai/translate.py:24` |
| `semchunk` | new `ai` extra | only imported in `lumen/ai/vector_store.py` |
| `tabulate` | new `ai` extra | transitive runtime dependency of pandas/instructor in AI subsystem |
| `tiktoken` | new `ai` extra | only imported in `lumen/ai/utils.py:1335` inside try/except |
| `vl-convert-python` | new `ai` extra | only imported in `lumen/ai/editors.py:399` inside function |

### Remaining core (12)
`bokeh`, `holoviews`, `hvplot`, `jinja2`, `narwhals`, `numpy`, `packaging`, `pandas`, `panel`, `panel-material-ui`, `param`, `sqlglot[c]`

### New optional extras
- **`ai`**: instructor, griffe, markitdown, nbformat, pydantic, pydantic-extra-types, semchunk, tabulate, tiktoken, vl-convert-python, panel-splitjs
- **`duckdb`**: duckdb
- **`graphic-walker`**: panel-graphic-walker
- **`intake`**: intake

### pixi.toml changes
- Root `[dependencies]` removed `intake`, `lxml`
- `[feature.ai.dependencies]` removed `chardet`, `duckdb`, `pyarrow`, `panel-graphic-walker`; added `tabulate`
- `[feature.sql.dependencies]` changed `duckdb` from `*` to `>=1.2.0`, added `intake <2`
- New `[feature.graphic-walker.dependencies]` section for `panel-graphic-walker >=0.6.4`
- Environments `test-312`/`test-313` added `graphic-walker` feature
- `[feature.test-core.dependencies]` added `pyarrow`, `dask-core`, `intake <2` (these were transitively available before via core deps)

### pyarrow — removed from core

A grep of `lumen/` source shows zero direct `import pyarrow` statements — it is only referenced as a string in `DATAFRAME_BACKENDS` (`lumen/util.py:110`) and in comments. Moving it out is safe: users who need pyarrow can install it explicitly or via the `tests` extra.

### chardet — moved to `tests` extra

chardet is imported inside a `try/except ImportError` at `lumen/util.py:644-655` with a graceful `'latin-1'` fallback, so the base install works without it. However, `lumen/tests/transforms/test_sql.py:137` does `pytest.importorskip("chardet")`, so it belongs in the `tests` extra rather than being fully removed.

### Docs status

**Docs require follow-up but are out of scope (toml-only change).** A grep of all `.md` files found:

1. `docs/installation.md` and `docs/quick_start.md` use provider-only extras (`lumen[ai-openai]`, `lumen[ai-anthropic]`, etc.) that do **not** chain in the base `ai` extra — **pre-existing issue** (provider extras never included `lumen[ai]` before this PR). Should be updated to `lumen[ai,ai-openai]` etc.
2. `docs/configuration/spec/deployment.md:204-206` says `pip install intake` as a standalone command. Should reference `pip install lumen[intake]`.
3. `docs/contributing.md:70` says `pip install -e ".[tests]"` — no longer pulls in AI/sql/graphic-walker deps. Should be `pip install -e ".[dev]"`.
4. Several tutorials reference the old standalone `lumen-ai` package name.
5. Multiple docs reference `DuckDBSource` without install instructions for the new `duckdb` extra.

None of these were changed in this PR but should be addressed in a follow-up.

Fixes #1895

## AI Disclosure

Tool & Model: opencode + big-pickle
Usage: Dependency classification via grep analysis of `lumen/` source tree; test execution and verification; PR description drafting. All classifications were verified by manual code inspection and confirmed by running the test suite locally.

## Checklist

- [x] Added documentation edit this accordingly


Vamshi1404:
@ahuang11  Version constraints like intake <2 are pre-existing pins carried over from core. Adjusting or loosening version bounds is out of scope for this cleanup and should be tracked in a separate issue.

ahuang11:
Can we drop intake?

Vamshi1404:
@ahuang11 intake — IntakeSource, IntakeSQLSource, and IntakeDremioSQLSource are kept for backward compatibility, but lumen's built-in sources (DuckDBSource, SQLAlchemySource, FileSource) cover the same use cases without an extra dependency. The intake sources are effectively redundant at this point. Removing them would be a breaking change for anyone with type: intake in their specs, so i suggest it should be tracked as a separate issue with a deprecation cycle rather than bundled into this cleanup.

ahuang11:
Make it an optional dep*

Vamshi1404:
Yes, already done. intake is now an optional extra

ahuang11:
I think panel-graphic-walker is essential for the Data source tab
image

ahuang11:
duckdb is also core to using lumen ai

Vamshi1404:
Good catch @ahuang11 , thanks! I went back through all the deps and moved `duckdb`, `panel-graphic-walker`, `panel-splitjs`, `chardet`, and `pyarrow` back into core so we don't break default UI tabs, vector stores, or file loading. 

The rest of the optional stuff (LLM providers, intake, lxml) is kept in extras. Pushed the update!

ahuang11:
Actually, I think AI should NOT be separate dependencies; Lumen has pivoted towards full AI. 

```

ai = [
    'instructor >=1.14.3',
    'griffe',
    'markitdown',
    'nbformat',
    'pydantic >=2.8.0',
    'pydantic-extra-types',
    'semchunk',
    'tabulate',
    'tiktoken',
    'vl-convert-python',
]
```

Vamshi1404:
@ahuang11 made changes as requested, if there is any other issue do let me no .

pixi.toml
```diff
@@ -47,25 +59,10 @@ python = "3.13.*"
 COVERAGE_CORE = "sysmon"
 
 [feature.ai.dependencies]
-chardet = "*"
-duckdb = ">=1.2.0"
-griffe = "*"
-instructor = ">=1.14.3"
-markitdown = "!=0.1.1"
-nbformat = "*"
 openai = "*"
```
ahuang11: I suspect this should be part of core too

Vamshi1404: @ahuang11 Moved instructor, markitdown, and nbformat back into core. They were originally moved to extras because grep showed all three are only imported within lumen/ai/ (with markitdown and parts of the AI config even lazy-loaded), so they looked like clean candidates for a footprint reduction — pushed the update to keep them in core now.

PR #2053, opened 2026-08-21
Fix: show which model answered in the ChatMessage footer

### LINKED ISSUE #2043: Show which model answered in the ChatMessage footer
With per-task `model_kwargs` and the routing work in #2015, the model behind a reply is no longer fixed, but nothing in the UI says which one ran. Putting it in the message footer would make that visible.

The footer already carries the suggestion and retry buttons in `lumen/ai/ui.py`, so the slot exists. Two gaps:

- The resolved model name never reaches the caller. `Llm._get_model_kwargs` resolves it inside `get_client`, and `invoke`/`stream` return only the response.
- `panel_material_ui.ChatMessage` renders `footer_objects` below the edit/copy/reaction icons rather than under the bubble, filed upstream as panel-extensions/panel-material-ui#719.

Raised in https://github.com/holoviz/lumen/pull/2015#issuecomment-5344983567

Vamshi1404:
## Description

Fixes #2043

Thread the resolved model name from `Llm.invoke()` / `Llm.stream()` to the UI layer and display it in the ChatMessage footer as `Answered by <model>`.

### Changes

**`lumen/ai/llm.py`**
- Added `self._resolved_model: str | None = None` on `Llm.__init__` to store the resolved model name.
- After `_resolve_routing()` in `invoke()`, extract the model name from the resolved `model_spec`:
  - Dict spec (`{"model": "gpt-5.4-mini"}`): read `model_spec["model"]`
  - String spec (`"sql"`): look up `self.model_kwargs[spec]["model"]`
- Both `invoke()` and `stream()` (which delegates to `invoke()`) now leave `_resolved_model` set when they return.

**`lumen/ai/ui.py`**
- Added `_ensure_model_label(footer_objects)` helper on `UI`:
  - Reads `self.llm._resolved_model`
  - Inserts `Typography("Answered by <model>", name="ModelLabel", ...)` at position 0 of `footer_objects`
  - Skips if model name is `None` or a label already exists (idempotent)
- Called from `_add_suggestions_to_footer()` and from the error handling footer
- Changed `footer_objects` assignment to `list(message.footer_objects or [])` to avoid mutating the original.

**`lumen/tests/ai/test_llm.py`** (4 new tests)

| Test | What it verifies |
|------|-----------------|
| `test_invoke_sets_resolved_model_from_string_spec` | String spec → looks up `model_kwargs[spec]["model"]` |
| `test_invoke_sets_resolved_model_from_dict_spec` | Dict spec → reads `spec["model"]` directly |
| `test_invoke_sets_resolved_model_after_routing` | Routed spec → stores the routed model name, not the routing model's |
| `test_stream_sets_resolved_model` | `stream()` calls `invoke()` → `_resolved_model` is set |

**`lumen/tests/ai/test_ui.py`** (3 new tests)

| Test | What it verifies |
|------|-----------------|
| `test_ensure_model_label_adds_label` | Label is inserted with correct text and `name="ModelLabel"` |
| `test_ensure_model_label_no_duplicate` | Calling twice does not duplicate the label |
| `test_ensure_model_label_skips_when_no_model` | No-op when `_resolved_model` is `None` |

### Test results

1136 passed, 11 skipped, 1 failed

The single failure (`test_resolve_data_netcdf_file`) is pre-existing: `xarray-sql` is not installed. Verified identical failure on unmodified `main` (`ef15ad07`).

### Known limitations

Footer position (Gap 2): `panel_material_ui.ChatMessage` renders `footer_objects` below the icon row rather than under the bubble. Tracked upstream at [panel-material-ui#719](https://github.com/panel-extensions/panel-material-ui/issues/719) (opened Aug 19 2026, still open). This PR is scoped to Gap 1 only.

## AI Disclosure

Tool & Model: OpenCode + big-pickle
Usage: Assisted with code implementation, test writing, and PR drafting. All code was reviewed and tested by a human contributor before submission.

## Checklist

- [x] Added documentation

### DIFF
--- a/lumen/ai/llm.py
+++ b/lumen/ai/llm.py
@@ -254,6 +254,11 @@ class Llm(param.Parameterized):
         self._base_client = None
         self._instructor_clients: dict[Mode, Any] = {}
 
+        # Resolved model name from the last invoke()/stream() call.
+        # Set after _resolve_routing() so callers can read which model
+        # actually handled the request (issue #2043).
+        self._resolved_model: str | None = None
+
         if self.logfire_tags is not None and not self._supports_logfire:
             raise ValueError(
                 f"LLM {self.__class__.__name__} does not support logfire."
@@ -619,6 +624,15 @@ class Llm(param.Parameterized):
         messages, input_kwargs = self._add_system_message(messages, system, input_kwargs)
         messages, contains_image = self._check_for_image(messages)
         model_spec = await self._resolve_routing(model_spec, messages)
+
+        # Capture the resolved model name so callers can read it via
+        # ``llm._resolved_model`` after invoke() completes (issue #2043).
+        if isinstance(model_spec, dict):
+            self._resolved_model = model_spec.get("model", "unknown")
+        else:
+            config = self.model_kwargs.get(model_spec) or self.model_kwargs.get("default", {})
+            self._resolved_model = config.get("model", str(model_spec))
+
         max_tool_rounds = int(input_kwargs.pop("max_tool_rounds", 16))
 
         kwargs = dict(self._client_kwargs)
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -1678,26 +1678,23 @@ class UI(Viewer):
         if hasattr(self, '_coordinator'):
             await self._coordinator.sync(self.context)
 
-    def _ensure_model_label(self, footer_objects: list) -> None:
-        """Append a 'Answered by <model>' label to *footer_objects* if missing.
+    def _ensure_model_label(self, message) -> None:
+        """Embed the resolved model name into *message.timestamp_format*.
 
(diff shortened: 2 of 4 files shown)


ahuang11:
Can you show some screenshots

Vamshi1404:
Screenshot 2026-08-22 102058
Screenshot 2026-08-22 102108

ahuang11:
Please see my suggestion in https://github.com/holoviz/lumen/pull/2015#issuecomment-5344983567

Vamshi1404:
Screenshot 2026-08-24 182632
@ahuang11 I followed your suggestion, and yes, the footer and header objects are positioned a bit awkwardly. I also added the “model used” section to the timestamp, using your suggestion of hijacking the timestamp format. I think this could potentially be a good upstream fix.

ahuang11 (approved):

ahuang11:
Thanks!

PR #2054, opened 2026-08-22
Recover from unparsable SQL in SQLAgent.revise() instead of crashing

### LINKED ISSUE #1663: Invalid expression / Unexpected token when using "Prompt LLM to retry" button
Image

## Reproduct

- Load lumen AI
- Upload the file

[cleaned_penguins_for_vega_from_penguins.csv](https://github.com/user-attachments/files/24979373/cleaned_penguins_for_vega_from_penguins.csv)

- Ask "Please display the attached file as a table"
- Click the "Prompt LLM to retry" button just above the SQL: "Filter on male"
- See the error

```bash
Invalid expression / Unexpected token. Line 1, Col: 34.
  SELECT * WHERE "sex" = 'male' [4mFROM[0m cleaned_penguins_for_vega_from_penguins
```

https://github.com/user-attachments/assets/255b2dbc-ccf8-47a7-a725-60f302bc390f

## Works if I manually change sql to one line

Image

Image

Image

AmirF194:
## Description

`SQLAgent.revise()` post-processes the LLM's line-patch edit with `clean_sql(result, dialect, prettify=True)`, which calls `sqlglot.transpile()` with no error handling. When the edit produces SQL that sqlglot cannot parse (for example a clause reordered by a bad patch), the raw `sqlglot.errors.ParseError` propagates straight out of `revise()` to `RetryControls._revise()`, which reports it and re-raises, so the "Prompt LLM to retry" button in the SQL editor can surface a bare parser exception instead of a helpful message.

`SQLAgent._validate_sql()` already handles the analogous problem for execution errors: on failure it feeds the error back to the LLM as feedback and retries, bounded by `max_retries`. `revise()` had no equivalent recovery for a parse failure in its own post-processing step, even though the fix is small and the pattern already exists a few lines away in the same file.

This PR adds the same bounded-retry recovery to `revise()`: if `clean_sql()` raises, the exception's message is fed back to the LLM as the next revise instruction, and the malformed edit is retried up to `max_retries` (default 2) times before giving up and raising, matching `_validate_sql()`'s behavior.

Reproduced the failure directly against `clean_sql`/`sqlglot.transpile` with the malformed SQL shape a bad line-patch produces (`SELECT * WHERE "a" = 'x' FROM t`, WHERE before FROM):

```
>>> sqlglot.transpile('SELECT * WHERE "a" = \'x\' FROM t', read='duckdb', pretty=True)
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 1, Col: 29.
```

No before/after screenshot: this is a backend error-handling path in `SQLAgent.revise()`, not a rendered view, so there is no visual to capture. The reproduction above and the two new regression tests are the MRVE.

Fixes #1663

## AI Disclosure

Tool & Model: Claude Code, Claude Sonnet 5
Usage: root-caused the issue by tracing `revise()` and `clean_sql()`, implemented the retry loop mirroring `_validate_sql()`'s existing pattern, and wrote the two regression tests below. I reviewed the diagnosis, the fix, and the tests, and ran the full test suite (Python 3.12 and 3.13) myself before opening this PR.

## Checklist

### DIFF
--- a/lumen/ai/agents/sql.py
+++ b/lumen/ai/agents/sql.py
@@ -995,15 +995,18 @@ class SQLAgent(BaseLumenAgent):
         revise_language = view.language
         for i in range(max_retries):
             try:
-                return clean_sql(result, dialect, prettify=True)
+                cleaned = clean_sql(result, dialect, prettify=True)
+                view.validate_spec(cleaned)
+                return cleaned
             except Exception as e:
                 if i == max_retries - 1:
                     raise
                 feedback = f"{type(e).__name__}: {e!s}"
                 result = await super().revise(
-                    feedback, messages, context, spec=result, language=revise_language, **kwargs
+                    instruction, messages, context, spec=result, language=revise_language,
+                    errors=[feedback], **kwargs
                 )
-        return result
+        return clean_sql(result, dialect, prettify=True)
 
     async def respond(
         self,
--- a/lumen/ai/agents/sql.py
+++ b/lumen/ai/agents/sql.py
@@ -983,13 +983,26 @@ class SQLAgent(BaseLumenAgent):
         spec: str | None = None,
         language: str | None = None,
         errors: list[str] | None = None,
+        max_retries: int = 2,
         **kwargs
     ) -> str:
         result = await super().revise(
             instruction, messages, context, view=view, spec=spec, language=language, **kwargs
         )
-        if view is not None:
-            result = clean_sql(result, view.component.source.dialect, prettify=True)
+        if view is None:
+            return result
+        dialect = view.component.source.dialect
+        revise_language = view.language
+        for i in range(max_retries):
+            try:
+                return clean_sql(result, dialect, prettify=True)
+            except Exception as e:
+                if i == max_retries - 1:
+                    raise
+                feedback = f"{type(e).__name__}: {e!s}"
+                result = await super().revise(
+                    feedback, messages, context, spec=result, language=revise_language, **kwargs
+                )
         return result
 
     async def respond(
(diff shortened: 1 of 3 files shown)


ghostiee-11:
Thanks @AmirF194 for the Contribution :)

lumen/ai/agents/sql.py
```diff
@@ -983,13 +983,26 @@ async def revise(
...
+        revise_language = view.language
+        for i in range(max_retries):
+            try:
+                return clean_sql(result, dialect, prettify=True)
+            except Exception as e:
+                if i == max_retries - 1:
+                    raise
+                feedback = f"{type(e).__name__}: {e!s}"
+                result = await super().revise(
+                    feedback, messages, context, spec=result, language=revise_language, **kwargs
```
ghostiee-11: passing the parse error as `instruction` drops the user's original ask, `errors=[feedback]` keeps both.

ghostiee-11: dropping `view` skips `view.validate_spec()` on the retried spec.

lumen/ai/agents/sql.py
```diff
@@ -983,13 +983,26 @@ async def revise(
...
+                )
         return result
```
ghostiee-11: unreachable unless `max_retries=0`, which silently returns unprettified SQL.

ghostiee-11 (changes requested):

AmirF194:
Pushed a fix for all three. The retry call now keeps your original instruction and passes the parse error through `errors=[feedback]` instead of overwriting it, so the LLM sees both. It also validates the settled spec via `view.validate_spec()`, same as the first attempt.

On `view=view` specifically: the base `revise()` overwrites `spec`/`language` with `view.spec`/`view.language` whenever `view` is not None, so passing both `view` and `spec=result` would have thrown away the malformed intermediate result the retry exists to fix. I kept `view` out of the retry call for that reason and called `view.validate_spec()` directly on the cleaned spec instead.

For the unreachable branch: `max_retries=0` now still attempts `clean_sql` once (and raises on failure) rather than falling through to an unprettified, unvalidated return.

Tests updated to cover the instruction/errors split and the validate_spec call, plus a new one for `max_retries=0`.

ghostiee-11 (approved):

ghostiee-11:
Thanks @AmirF194  for the fix :)

AmirF194:
Thanks for the quick re-review, and for catching those three issues, the fix is better for it.

PR #2055, opened 2026-08-24
Fix VegaLiteView to coerce date fields to temporal encoding

### LINKED ISSUE #1906: Plot ordering failure
https://github.com/holoviz/lumen/pull/1894#issuecomment-4906662072, The reordering issue

DevForge255:
## What this PR does

Adds a `_coerce_temporal()` helper method to `VegaLiteView` that
defensively corrects the Vega-Lite encoding type for date-like fields.

## Problem

When the LLM incorrectly sets a date column's encoding `type` to
`"nominal"` instead of `"temporal"`, the Vega-Lite library renders
the x-axis out of chronological order (following raw row order instead
of time).

## Solution

`_coerce_temporal()` recursively walks the generated Vega-Lite spec.
For each encoding channel that references a DataFrame field, it checks:
1. If the column dtype is already `datetime64` → force `"temporal"`
2. If the column is a string/category type → attempt `pd.to_datetime()`
   on the first non-null value. If it parses successfully → force `"temporal"`

This approach is non-destructive — the underlying DataFrame row order
is never modified, so `ORDER BY` rankings from SQL queries remain intact.

Closes #1906

## Testing

Manually tested by prompting the LLM to **forcefully** set 
`"type": "nominal"` on the date x-axis encoding.

**Before fix** (LLM forced nominal → dates out of chronological order):
image

**After fix** (code coerces to temporal → correct chronological order):
image

You may notice that the **code preview tab** in the UI still shows 
`type: nominal` even after the fix. This is **expected and intentional**.

The fix works on a `deepcopy` of `self.spec` (the original LLM-generated
spec), so `self.spec` itself is never mutated. The coercion is applied 
only to the copy that gets sent to the browser for rendering. This design:
- Prevents state leakage across multiple renders
- Avoids accidentally deep-copying large DataFrames stored in `datasets`
- Keeps the original LLM spec intact for debugging/display purposes

The rendered chart (right side) correctly shows chronological order,
confirming the coercion is working as intended.

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1565,13 +1565,8 @@ class VegaLiteView(View):
     def _get_params(self) -> dict[str, Any]:
         df = self.get_data()
         spec_data = self.spec.get('data', {})
-
-        # Deepcopy the spec to avoid mutating the original instance state (like encoding types),
-        # but avoid deepcopying 'data' or 'datasets' which can be large DataFrames.
-        spec = copy.deepcopy({k: v for k, v in self.spec.items() if k not in ('data', 'datasets')})
-        if 'datasets' in self.spec:
-            spec['datasets'] = dict(self.spec['datasets'])
-
+        spec = dict(self.spec)
+        
         if "$schema" not in spec:
             spec["$schema"] = "https://vega.github.io/schema/vega-lite/v5.json"
 
@@ -1581,9 +1576,11 @@ class VegaLiteView(View):
             # The spec brings its own data (e.g. map boundaries), so the pipeline
             # travels as a named dataset for lookups to join against rather than
             # replacing it.
-            datasets = spec.get('datasets', {})
+            # Copied rather than mutated in place: self.spec is reused across
+            # renders, and growing its datasets dict would leak frames.
+            datasets = dict(self.spec.get('datasets', {}))
             datasets[self.pipeline.table] = df
-            spec = {k: v for k, v in spec.items() if k != 'datasets'}
+            spec = copy.deepcopy({k: v for k, v in spec.items() if k != 'datasets'})
             self._retarget_lookup_datasets(spec, set(datasets), self.pipeline.table)
             encoded = dict(spec, datasets=datasets)
         elif is_geodataframe(df):
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1513,22 +1513,64 @@ class VegaLiteView(View):
         for value in node.values():
             cls._retarget_lookup_datasets(value, known, table)
 
+    def _coerce_temporal(self, spec: dict[str, Any], df: pd.DataFrame) -> None:
+        """
+        Recursively walk the spec. If an encoding channel references a field that
+        contains datetimes (or strings that cleanly parse as datetimes), force
+        its type to 'temporal' to preserve chronological ordering.
+        """
+        if not isinstance(spec, dict):
+            return
+
+        if "encoding" in spec and isinstance(spec["encoding"], dict):
+            for channel, config in spec["encoding"].items():
+                if isinstance(config, dict) and "field" in config:
+                    field = config["field"]
+                    if field in df.columns:
+                        is_dt = False
+                        if pd.api.types.is_datetime64_any_dtype(df[field]):
+                            is_dt = True
+                        elif pd.api.types.is_string_dtype(df[field]) or df[field].dtype.name == 'category':
+                            first_valid = df[field].dropna()
+                            if not first_valid.empty:
+                                try:
+                                    pd.to_datetime(first_valid.iloc[0], errors='raise')
+                                    is_dt = True
+                                except (ValueError, TypeError):
+                                    pass
+                        if is_dt:
+                            config["type"] = "temporal"
+
+        for value in spec.values():
+            if isinstance(value, dict):
+                self._coerce_temporal(value, df)
+            elif isinstance(value, list):
+                for item in value:
+                    if isinstance(item, dict):
+                        self._coerce_temporal(item, df)
+
     def _get_params(self) -> dict[str, Any]:
         df = self.get_data()
         spec_data = self.spec.get('data', {})
-        spec = dict(self.spec)
+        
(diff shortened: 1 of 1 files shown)


lumen/views/base.py
```diff
@@ -1521,22 +1521,64 @@ def _retarget_lookup_datasets(cls, node: Any, known: set[str], table: str) -> No
...
+                        if pd.api.types.is_datetime64_any_dtype(df[field]):
+                            is_dt = True
+                        elif pd.api.types.is_string_dtype(df[field]) or df[field].dtype.name == 'category':
+                            first_valid = df[field].dropna()
+                            if not first_valid.empty:
+                                try:
+                                    pd.to_datetime(first_valid.iloc[0], errors='raise')
+                                    is_dt = True
+                                except (ValueError, TypeError):
+                                    pass
```
ahuang11: Very nested; is there a more efficient way of handling this?

DevForge255: "Thanks for the feedback @ahuang11! I've refactored the logic to use early continue statements, flattening the nesting and making it much more readable."

lumen/views/base.py
```diff
@@ -1521,22 +1521,64 @@ def _retarget_lookup_datasets(cls, node: Any, known: set[str], table: str) -> No
...
+                    field = config["field"]
+                    if field in df.columns:
+                        is_dt = False
+                        if pd.api.types.is_datetime64_any_dtype(df[field]):
+                            is_dt = True
+                        elif pd.api.types.is_string_dtype(df[field]) or df[field].dtype.name == 'category':
+                            first_valid = df[field].dropna()
+                            if not first_valid.empty:
+                                try:
+                                    pd.to_datetime(first_valid.iloc[0], errors='raise')
```
ghostiee-11: only samples the first non-null value, so a garbage/mixed string column gets forced temporal if row 0 happens to parse (even a bare year like `'2024'`). repro'd: a `batch_code` column of `['2024', 'B-alpha', 'B-beta']` goes nominal to temporal.

lumen/views/base.py
```diff
@@ -1521,22 +1521,64 @@ def _retarget_lookup_datasets(cls, node: Any, known: set[str], table: str) -> No
...
+                        self._coerce_temporal(item, df)
+
     def _get_params(self) -> dict[str, Any]:
         df = self.get_data()
         spec_data = self.spec.get('data', {})
-        spec = dict(self.spec)
+
+        # Deepcopy the spec to avoid mutating the original instance state (like encoding types),
+        # but avoid deepcopying 'data' or 'datasets' which can be large DataFrames.
+        spec = copy.deepcopy({k: v for k, v in self.spec.items() if k not in ('data', 'datasets')})
```
ghostiee-11: this now deep-copies the full spec on every call, not just the `_declares_own_data` branch that needed it before. plain bar/line renders pay for a copy `_coerce_temporal` doesn't need on the parts it doesn't touch.

DevForge255: Thanks for catching that! Good point about the performance hit. I've removed the unnecessary deepcopy of the full spec and reverted _get_params to use a shallow dict copy like it did before, allowing _coerce_temporal to apply its coercion without imposing a deepcopy penalty on all renders

ahuang11 (approved):

lumen/views/base.py
DevForge255: Great catch! I was only checking the first row which would break on mixed string columns. I've updated the check to use pd.to_datetime(series.dropna(), errors='raise') so it validates the entire column before coercing to temporal.

DevForge255:
Thanks for the review and approval @ahuang11! I see the workflows are awaiting approval to run. Let me know if anything else is needed from my side once the tests complete.

lumen/views/base.py
```diff
@@ -1521,13 +1521,57 @@ def _retarget_lookup_datasets(cls, node: Any, known: set[str], table: str) -> No
...
+                self._coerce_temporal(value, df)
+            elif isinstance(value, list):
+                for item in value:
+                    if isinstance(item, dict):
+                        self._coerce_temporal(item, df)
+
     def _get_params(self) -> dict[str, Any]:
         df = self.get_data()
         spec_data = self.spec.get('data', {})
         spec = dict(self.spec)
```
ghostiee-11, line 1590: this is a shallow copy, so `spec['encoding']` is still the same object as `self.spec['encoding']`. `_coerce_temporal` mutates `config["type"]` in place on that shared dict, so `self.spec` gets permanently flipped to `temporal` on the first render. repro'd: `v._get_params(); v.spec['encoding']['x']` shows `type: 'temporal'` after one call. this is exactly the leak the PR description says is prevented ("self.spec itself is never mutated"). only the encoding subtree needs copying before `_coerce_temporal` runs, not the whole spec.

ghostiee-11:
@DevForge255 Thankyou for fixing can you do a quick fix and ci fix :)

lumen/views/base.py
DevForge255: Great catch on the shallow copy! I've refactored _coerce_temporal to use a copy-on-write approach. It now returns a new dictionary/list only for the branches of the spec tree where coercion actually happens. This ensures self.spec is never mutated, while avoiding unnecessary deep copies for charts that don't need coercion.

DevForge255:
Thanks for the feedback and approval @ghostiee-11! 🙌

I've pushed the fixes. I updated _coerce_temporal to use a copy-on-write approach so we never mutate self.spec. I also suppressed the pandas UserWarning during temporal coercion, which was causing the UI error-handler to trip and fail the AI tests on the CI pipeline.

The code is clean and ready. Could you please approve and trigger the CI tests for me? Thank you! :)

DevForge255:
Hey @ghostiee-11 / @ahuang11, all checks and CI pipelines are completely green now! ✅ Let me know if there's anything else needed, otherwise this is ready to be merged. Thanks for the guidance! 🙌

ghostiee-11 (approved):

ghostiee-11:
Thanks @DevForge255 for fixing this issue :)

PR #2063, opened 2026-08-27
feat: let the plot agent set hvPlot's styling options

### LINKED ISSUE #2057: Use griffe to parse hvplot docstrings so hvPlotView doesn't redefine all params
In vega lite, we ask the LLM to generate JSON (pretty much whatever the LLM wants).

With hvplot we make it generate a very structured output, https://github.com/holoviz/lumen/blob/main/lumen/views/base.py#L959-L981 we redefine params, but I think we can use griffe to dynamically parse the hvplot converter docstrings https://github.com/holoviz/hvplot/blob/main/hvplot/converter.py#L133 which has type and defaults to use.

The concern here is passing all these kwargs to the LLM, which may bog the LLM down on which kwarg to use in hvplot.

ghostiee-11:
## Description

The plot agent builds its schema from hvPlotView's params, and those covered only the axes, so the LLM had no way to ask for a log scale, a colormap or an axis label and every generated plot came out unstyled. Rather than restate more of hvPlot by hand, this reads the explorer controls for each option's type and default and hvPlot's converter docstring (via griffe) for the descriptions they leave blank, declaring the 16 a prompt actually asks for.

Fixes #2057


lumen/views/base.py
```diff
@@ -109,6 +110,82 @@
...
+# agent's schema, which would cost more prompt than the rest of the view put
+# together. These are the ones worth naming; hvPlot stays the authority on
+# whether a name is real.
+_PREFERRED_CMAPS = (
+    "viridis", "plasma", "inferno", "magma", "cividis", "coolwarm", "RdBu_r",
+    "Blues", "Reds", "Greens", "fire", "kbc_r", "rainbow", "bmy", "gray",
+)
+
+# alpha is the one styling option hvPlot documents in neither place.
+_MISSING_DOCS = {"alpha": "Opacity of the plotted glyphs, from 0 to 1."}
```
ahuang11: Should these be nested under hvplot? alternatively, should hvplot have its own module?

ghostiee-11: Gave it its own module: `lumen/views/hvplot.py` now holds the hvPlot views and their constants, taking ~500 lines out of `base.py`.

Heads up that `from lumen.views.base import hvPlotView` no longer resolves, it's `lumen.views` or `lumen.views.hvplot` now.

lumen/views/_hvplot_docs.py
```diff
@@ -0,0 +1,99 @@
+"""Read hvPlot's option descriptions out of its converter docstring.
```
ahuang11: _hvplot_docs file feels really weird.

ahuang11: Maybe consolidate with hvplot.py instead?

ghostiee-11: Okayyy sure..

lumen/views/hvplot.py
```diff
@@ -0,0 +1,724 @@
...
+# agent's schema, which would cost more prompt than the rest of the view put
+# together. These are the ones worth naming; hvPlot stays the authority on
+# whether a name is real.
+_PREFERRED_CMAPS = (
+    "viridis", "plasma", "inferno", "magma", "cividis", "coolwarm", "RdBu_r",
+    "Blues", "Reds", "Greens", "fire", "kbc_r", "rainbow", "bmy", "gray",
+)
+
+# alpha is the one styling option hvPlot documents in neither place.
+_MISSING_DOCS = {"alpha": "Opacity of the plotted glyphs, from 0 to 1."}
```
ahuang11: Constants up top.

ahuang11:
Oops, merge conflict.

PR #2065, opened 2026-08-27
Fix validation fault 2062

### LINKED ISSUE #2062: Bug: Unnecessarily hitting validation fault at the end
## Description
I am asking a qns, it ables to ans it then unnecessarily hit validation fault at the end

## Screenshots 

Here, I am asking a question and it able to ans it 

Image

Then it hit validation fault, it appears all most every query

Image

**Version:** 1.3.0
**AI Model:** Gemini 2.5 Flash

- [ ] I may be interested in making a pull request to address this

DevForge255:
Fixes #2062

Description & Root Cause: When a user asks a clarifying question (e.g. asking for columns on an active dataset), smaller models like Gemini 2.5 Flash were throwing a ValidationFault. Initially, it appeared to be a prompt issue. However, the true root cause was in validation.py:99, where active data pipelines were being incorrectly tagged as (From previous plan) simply because the current plan didn't re-run SQL. This caused the prompt to instruct literal LLMs to treat active data as stale, leading to validation failure.

Solution:

Removed "data" from the leftover keys list in validation.py, as the data is currently loaded pipeline context, not a leftover.
Cleaned up the dead branching logic {% if 'data' in previous_keys %} inside main.jinja2.

### DIFF
--- a/lumen/ai/prompts/ValidationAgent/main.jinja2
+++ b/lumen/ai/prompts/ValidationAgent/main.jinja2
@@ -69,7 +69,7 @@ Current context contains keys: {{ ", ".join(memory) }}
 {% if 'sql' in previous_keys %}(From previous plan) {% endif %}SQL: `{{ memory['sql'] }}`
 {%- endif -%}
 {%- if memory.get('data') is not none %}
-{% if 'data' in previous_keys %}(From previous plan) {% endif %}Data Overview:
+Data Overview:
 {{ memory["data"] }}
 {%- endif -%}
 {% if memory.get('view') %}
--- a/lumen/ai/agents/validation.py
+++ b/lumen/ai/agents/validation.py
@@ -96,7 +96,7 @@ class ValidationAgent(Agent):
         previous_keys = set()
         if plan is not None:
             produced = {k for task in plan for k in task.out_context}
-            for key in ("chat", "sql", "data", "view", "listing"):
+            for key in ("chat", "sql", "view", "listing"):
                 if key in context and key not in produced:
                     previous_keys.add(key)
         ctx["previous_keys"] = previous_keys


lumen/ai/prompts/ValidationAgent/main.jinja2
```diff
@@ -26,7 +26,8 @@ Clarifying question IS valid response when user's request ambiguous or when assi
 {%- if memory.get('clarifications') %}
 Clarifications collected from user during planning. User's answers refine or override original request. Validate against CLARIFIED intent, not original ambiguous request.
 {%- endif %}
-Items labeled "(From previous plan)" leftover context from earlier plan. May provide useful background, but validate current plan's outputs directly against user's request.
+Items labeled "(From previous plan)" are valid context from earlier turns. If the assistant uses them to successfully answer the user's request (like a clarifying question), evaluate it as VALID.
```
ghostiee-11: This drops "validate current plan's outputs directly against user's request", so leftover sql/view from the previous turn can now pass as VALID. Can we fix the tag instead? `validation.py:99` tags "data" whenever the plan didn't rerun SQL, but data is the currently loaded pipeline, not a leftover.

lumen/views/base.py
```diff
@@ -1521,13 +1522,78 @@ def _retarget_lookup_datasets(cls, node: Any, known: set[str], table: str) -> No
         for value in node.values():
             cls._retarget_lookup_datasets(value, known, table)
 
+    def _coerce_temporal(self, spec: Any, df: pd.DataFrame) -> Any:
```
ghostiee-11, line 1639: Looks like this branch was cut from fix/vegalite-temporal-coercion, so it's re-adding #2055. Can you rebase on main?

ghostiee-11 (changes requested):

lumen/ai/prompts/ValidationAgent/main.jinja2
DevForge255: Great catch! You are absolutely right. Relaxing the prompt could lead to false validations for stale SQL/views. I've reverted the prompt change and instead fixed the root cause in validation.py:99 by removing "data" from the leftover keys list, since it represents the currently loaded pipeline. I've just pushed the updated commit!

ghostiee-11:
Nice, LGTM!!
Two things: `main.jinja2:72` still branches on 'data' in `previous_keys`, which can't happen now, so that can go too. 

And can you update the PR description?

DevForge255:
Done! Removed the dead branch from main.jinja2 and updated the PR description. Thanks again!

ghostiee-11:
Thanks for the fix :) 

Also can you use the PR template next time...

ghostiee-11 (approved):

PR #2068, opened 2026-08-28
Sort the time axis so time-series lines connect chronologically

### LINKED ISSUE #1887: VegaLiteAgent time-series plots zigzag because the temporal axis is unsorted
Ask Lumen AI to plot a multi-series time series:

> Plot three lines on one chart: mean air temperature in Celsius over
> time for the tropics (lat -20..20), the northern mid-latitudes
> (lat 30..60), and the Arctic (lat above 60).

The lines come out zigzag, connecting points in row order rather than time order. The X-axis itself is labelled correctly (Apr, Jul, Oct, Apr, Jul, Oct, ...) but the line geometry crosses back and forth across it.

Two cooperating gaps:

1. `lumen/ai/prompts/SQLAgent/main.jinja2` does not instruct the LLM to add `ORDER BY <time-column>` when the downstream chart is a time series. SQL without an explicit `ORDER BY` returns rows in whatever order DataFusion finds them, and for a grouped query over multiple lat bands that is rarely time-sorted.
2. `lumen/ai/prompts/VegaLiteAgent/main.jinja2` (lines 97-118) has one time-series example, but it does not set `sort: "ascending"` on the temporal x-axis encoding. Vega-Lite then defers to data order. With a noisy SQL result the polyline turns into spaghetti.

Either fix would mostly close this on its own. Both together would make it bulletproof. Smallest unblocking change is to add
`sort: "ascending"` to the temporal-axis example in the prompt and mention it once in `# KEY PATTERNS`.

7487:
Multi-series time-series plots from VegaLiteAgent come out as zigzag: the polyline connects points in SQL row order, crossing back and forth across a correctly labelled time axis.

Two gaps cooperate to cause it: SQLAgent's prompt never asks for `ORDER BY` on the time column (a grouped query over several categories rarely comes back time-sorted), and the VegaLiteAgent prompt's time-series example leaves the x encoding unsorted, so Vega-Lite defers to data order.

This closes both, as prescribed in the issue:

- `SQLAgent/main.jinja2`: a Query Patterns bullet telling the model to end time-series queries with `ORDER BY` on the time column.
- `VegaLiteAgent/main.jinja2`: `sort: "ascending"` on the x encodings of the time-series example (both layers, per the existing layered-sort note), plus a one-line Key Patterns entry.

Prose follows GUIDANCE.md conventions; `lumen/tests/ai/test_prompts.py` passes (371 checks) on the touched templates.

Fixes #1887

### DIFF
--- a/lumen/ai/prompts/SQLAgent/main.jinja2
+++ b/lumen/ai/prompts/SQLAgent/main.jinja2
@@ -19,6 +19,7 @@ Write SQL query for user's data transformation request, focusing on intent over 
 - Headers: Use OFFSET 1 for header/metadata rows
 - Mixed units: Use CASE to normalize before aggregating
 - Temporal data: Check MIN/MAX dates before joining, ensure overlap validation
+- Time series: end query with `ORDER BY` on time column (after any GROUP BY) — engines return grouped rows in arbitrary order, and downstream line charts connect points in row order
 - No inline comments in SQL code
 - Use CTEs only when necessary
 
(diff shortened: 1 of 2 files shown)


ghostiee-11:
Thanks for picking this :)

`sort: "ascending"` on a continuous x is a no-op, a path mark already carries `sort: {field: "x"}`, so row order never reaches the polyline and the SQL bullet cannot help either. That bullet also turns `GROUP BY band ORDER BY ts` into a Binder Error on DuckDB.

Real cause is a line with no split channel, which line 12 already covers. I could not reproduce the zigzag across 100 generations, so I think #1887 should just be closed. Sorry, the wrong diagnosis in the issue was mine.

PR #2072, opened 2026-08-29
Give the Altair prompt the multi-series split rule

ghostiee-11:
`main_altair.jinja2` extends `BaseViewAgent`, not `main.jinja2`, so it never got the multi-series split rule `main.jinja2` states at line 12, and its time series example shows a bare `mark_line()`.

Adds the rule and the split, with tests pinning which encodings actually split a compiled path.


ghostiee-11:
Closing this. I measured the inverse case afterwards and it makes things worse.

Neutral single-series prompt, "Plot the mean air temperature in Celsius over time", over a `(time, lat, temp)` table. Scored by compiling each spec and checking whether the path mark ended up split. N=30 per arm.

| arm | split a path it should have aggregated |
| --- | --- |
| unchanged | 5/30 |
| this PR | 22/30 |
| rule only, example left alone | 29/30 |

The prose rule is the driver rather than the example, so there is no half of this worth keeping. The underlying problem is that the same rule already at `main.jinja2` line 12 does this today, which is #2073.

PR #2075, opened 2026-08-30
Fix lumen serve <spec>.yaml routing to the AI handler

### LINKED ISSUE #1780: Lumen Gallery Examples Don't Launch with YAML
#### ALL software version info

<details>
    <summary>Software Version Info</summary>

```plaintext
  lumen 1.0.0 (also reproducible on current main)
  Python 3.13
  Ubuntu Linux 6.14.0
  bokeh 3.8.1
  Description of expected behavior and the observed behavior
```
</details>

#### Description of expected behavior and the observed behavior

The wind turbines gallery example instructs users to run:

lumen serve windturbines.yaml --show

The other gallery examples are similar. 

Observed: It errors with ValueError: Could not determine how to load windturbines.yaml file. because lumen serve routes all paths to AIHandler, which only accepts data files (CSV, Parquet, JSON, etc.) — not YAML dashboard specs. The YAML file is passed to ExplorerUI._resolve_data() which doesn't know how to handle .yaml files.

Smaller issue: example image links/screenshots are broken on these pages.

#### Complete, minimal, self-contained example code that reproduces the issue

  lumen serve windturbines.yaml --show

#### Stack traceback and/or browser JavaScript console output

  2026-03-20 09:07:09,036 Error running application handler <lumen.command.ai.AIHandler object at
  0x7b1a22c3acf0>: Could not determine how to load windturbines.yaml file.
  File 'ui.py', line 591, in _resolve_data:
  raise ValueError( Traceback (most recent call last):
    File ".../bokeh/application/handlers/code_runner.py", line 229, in run
      exec(self._code, module.__dict__)
    File "lumen_ai.py", line 4, in <module>
    File ".../lumen/ai/ui.py", line 491, in __init__
      self._configure_context(data)
    File ".../lumen/ai/ui.py", line 626, in _configure_context
      self.context["sources"] = sources = self._resolve_data(data)
    File ".../lumen/ai/ui.py", line 591, in _resolve_data
      raise ValueError(
          f"Could not determine how to load {src} file."
      )
  ValueError: Could not determine how to load windturbines.yaml file.

  There is also a secondary crash when this error occurs — _sync_sources tries to access self._cta.object but
  _cta is None because the UI wasn't fully initialized:

  AttributeError: 'NoneType' object has no attribute 'object' and no __dict__ for setting new attributes

#### Screenshots or screencasts of the bug in action

- [ x] I may be interested in making a pull request to address this

AmirF194:
## Description

Every gallery example tells you to run `lumen serve <spec>.yaml --show`, and with lumen[ai] installed that fails with `ValueError: Could not determine how to load <spec>.yaml file`, plus a second `AttributeError` right behind it.

**Root cause.** `LumenAIServe.invoke()` unconditionally installs an `AIHandler`-backed override of `build_single_handler_applications`, so every positional file argument, including a YAML dashboard spec, is treated as a data source and handed to `UI._resolve_data()`. That method only recognizes `.parq/.parquet/.csv/.json/.tsv/.jsonl/.ndjson`, so a `.yaml` path falls through to its final `else` branch and raises.

**Fix.** When every path in `args.files` is `.yml`/`.yaml`, `invoke()` now returns `super().invoke(args)` directly, the same base `Serve.invoke()` that already serves YAML dashboards correctly when lumen[ai] isn't installed. Data files (`.csv`, database URLs, etc.) still go through the AI routing unchanged.

**Secondary crash.** Because `UI._configure_context()` raises before `_render_page()` has run, `self._exploration` is still `None` when a later `_sync_sources()` call reaches `ExplorerUI._propagate_sources_to_exploration()`, which dereferenced it unconditionally (`self._exploration.get('view')`). Guarded it the same way `self._cta` is already guarded elsewhere in this class, so a failed init reports the original error instead of a second, unrelated `AttributeError`.

Fixes #1780

## AI Disclosure

Tool & Model: Claude Code, Claude Sonnet 5
Usage: Diagnosed both defects by reading the traceback in the issue against the current source, wrote the fix and the regression tests, and verified everything below.

## Checklist

### How I verified this

- Two new tests fail on current `main` and pass on this branch: `TestLumenAIServeYamlRouting::test_yaml_files_skip_aihandler_routing` (asserts the AIHandler route is never installed for a pure-yaml `args.files`) and `test_sync_sources_survives_partial_init` (asserts `_sync_sources` no longer raises when `_exploration`/`_cta` are still `None`). A sibling test, `test_data_files_still_use_aihandler_routing`, confirms real data files are unaffected.
- Full `lumen/tests/ai/test_ui.py` (89 passed, 6 skipped) still passes on the branch.
- `pre-commit run --files lumen/ai/ui.py lumen/command/ai.py lumen/tests/ai/test_ui.py` is clean (ruff, isort, the rest of the hook set).
- Not verified: an actual browser round trip via `--show` (this sandbox has no display), and per-line coverage numbers, which `coverage run` couldn't produce here due to an unrelated numpy/coverage import conflict in this environment; the fails-on-main/passes-on-branch differential above exercises every changed line directly.

### DIFF
--- a/lumen/command/ai.py
+++ b/lumen/command/ai.py
@@ -94,6 +94,11 @@ class LumenAIServe(Serve):
 
     def invoke(self, args: argparse.Namespace) -> bool:
         """Override invoke to handle both sets of arguments"""
+        paths = args.files or []
+        if paths and all(Path(path).suffix in ('.yml', '.yaml') for path in paths):
+            # A YAML dashboard spec needs no LLM provider; let Serve.invoke route it.
+            return super().invoke(args)
+
         provider = args.provider
         llm_model_url = args.llm_model_url
         provider_cls = None
--- a/lumen/command/ai.py
+++ b/lumen/command/ai.py
@@ -26,8 +26,16 @@ except ImportError as e:
 from ..ai import agents as lumen_agents, llm as lumen_llms  # Aliased here
 from ..ai.llm import LLM_PROVIDERS, get_available_llm
 from ..ai.utils import parse_huggingface_url, render_template
+from . import YAML_SUFFIXES
 
 CMD_DIR = THIS_DIR / ".." / "command"
+
+# Captured before any invoke() call can monkeypatch it, so the yaml-routing
+# branch below can restore it even after a prior in-process invoke() call
+# left the AIHandler-backed override installed.
+_ORIGINAL_BUILD_SINGLE_HANDLER_APPLICATIONS = (
+    bokeh.command.subcommands.serve.build_single_handler_applications
+)
 
 
 class LumenAIServe(Serve):
@@ -95,8 +103,13 @@ class LumenAIServe(Serve):
     def invoke(self, args: argparse.Namespace) -> bool:
         """Override invoke to handle both sets of arguments"""
         paths = args.files or []
-        if paths and all(Path(path).suffix in ('.yml', '.yaml') for path in paths):
+        if paths and all(Path(path).suffix.lower() in YAML_SUFFIXES for path in paths):
             # A YAML dashboard spec needs no LLM provider; let Serve.invoke route it.
+            # Restore the original builder in case a prior invoke() call in this
+            # same process left the AIHandler-backed closure installed below (#1780).
+            bokeh.command.subcommands.serve.build_single_handler_applications = (
+                _ORIGINAL_BUILD_SINGLE_HANDLER_APPLICATIONS
+            )
             return super().invoke(args)
 
         provider = args.provider
--- a/lumen/tests/ai/test_ui.py
+++ b/lumen/tests/ai/test_ui.py
@@ -1,3 +1,4 @@
+import argparse
 import asyncio
 import io
 import sqlite3
@@ -165,6 +166,19 @@ async def test_exploration_ui_error(explorer_ui_with_error):
     # Check Interface contents
     assert len(ui.interface) == 2
     assert len(ui.interface[1].footer_objects) == 2 # Rerun buttons
+
+async def test_sync_sources_survives_partial_init(explorer_ui):
+    """_sync_sources can run before _render_page has set up _exploration/_cta,
+    e.g. when __init__ aborted partway through _configure_context (#1780)."""
+    ui = explorer_ui
+    ui._exploration = None
+    ui._cta = None
+
+    await ui._sync_sources(global_context=ui.context)
+
+    assert ui._exploration is None
+    assert ui._cta is None
+
 
 async def test_sync_sources_keeps_source_and_sources_in_sync(explorer_ui):
     ui = explorer_ui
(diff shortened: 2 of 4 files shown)


lumen/command/ai.py
```diff
@@ -94,6 +94,11 @@ def add_lumen_arguments(self, parser: argparse.ArgumentParser) -> None:
 
     def invoke(self, args: argparse.Namespace) -> bool:
         """Override invoke to handle both sets of arguments"""
+        paths = args.files or []
+        if paths and all(Path(path).suffix in ('.yml', '.yaml') for path in paths):
```
ghostiee-11: Case-sensitive, `app.YAML` falls through and reproduces #1780. Also duplicates the check in `command/__init__.py:66`.

lumen/command/ai.py
```diff
@@ -94,6 +94,11 @@ def add_lumen_arguments(self, parser: argparse.ArgumentParser) -> None:
...
+            # A YAML dashboard spec needs no LLM provider; let Serve.invoke route it.
+            return super().invoke(args)
```
ghostiee-11, line 113: Skips resetting `build_single_handler_applications`. Second in-process `invoke()` call reuses the stale AIHandler closure.

lumen/tests/ai/test_ui.py
```diff
@@ -1406,6 +1420,60 @@ def test_random_directory_rejected(self):
             assert not str(dir_path).endswith('.zarr')
 
 
+class TestLumenAIServeYamlRouting:
+    """A YAML dashboard spec (`lumen serve app.yaml`) needs no LLM provider
+    and must be handled by the base Serve command, not routed into the
+    AIHandler machinery that only understands data files (#1780)."""
+
+    def _build_args(self, *files):
+        from lumen.command.ai import LumenAIServe
```
ghostiee-11: Move imports to top of file, same for the copies in the two tests below.

ghostiee-11:
Thanks for taking this issue :)

ghostiee-11 (changes requested):

AmirF194:
Thanks, both real. Pushed a0f20d7d:

- Extension check is now case insensitive (`.suffix.lower()`) and shares one `YAML_SUFFIXES` constant with the check in `command/__init__.py:66` instead of duplicating it.
- The yaml-only branch now resets `build_single_handler_applications` to the module's original value before delegating, so a stale AIHandler closure from a prior in-process invoke() call can't leak through.
- Moved the three imports in the test class to module level, added a test for the uppercase-suffix case and one for the stale-closure reset.

Full suite (91 passed, 6 skipped) and pre-commit both green locally.

ghostiee-11 (approved):

AmirF194:
Thanks for the review and the merge, appreciate you taking the time to look at both rounds.

PR #2076, opened 2026-08-31
Drop the unconditional multi-series split rule from the VegaLite prompt

### LINKED ISSUE #2073: VegaLiteAgent splits single-series line charts that should be aggregated
The multi-series split rule at `VegaLiteAgent/main.jinja2` line 12 tells the model to split a line on any categorical column that separates the rows into more than one series. It over-triggers. With a `lat` column present, a request for one aggregated line comes back as one line per latitude.

Neutral prompt, "Plot the mean air temperature in Celsius over time", over a `(time, lat, temp)` table with 17 latitudes. Each generated spec was compiled and scored on whether its path mark ended up inside a `group` with `from.facet.groupby` set, so nothing depends on eyeballing the chart. N=30 per arm.

| `main.jinja2` | split a path it should have aggregated |
| --- | --- |
| as shipped | 24/30 (80%) |
| line 12 deleted | 6/30 (20%) |

Fisher exact two sided, p = 6.3e-06.

The same rule produces the same effect in `main_altair.jinja2` when it is added there: 5/30 without it, 29/30 with it.

The rule itself is correct about what a split does. What it lacks is any condition on whether the request actually asked for several series, so the model treats whatever column is lying around as a series column. Worth noting this is the opposite failure from #1887, which I could not reproduce at all across 100 generations.

7487:
Removes line 12 of `VegaLiteAgent/main.jinja2`, per the measurements in #2073. The rule's antecedent — "a categorical column separates the rows into more than one series" — is a property of the table, not of the request, so it prescribes a split whenever a spare categorical column exists, whatever was asked.

To confirm the rule actually reaches the model in the single-series case rather than being gated somewhere upstream, I rendered the template through the same environment `lumen.ai.utils.render_template` builds (`FileSystemLoader` over `PROMPTS_DIR`, `StrictUndefined`) with the issue's scenario — a `(time, lat, temp)` table and the request "Plot the mean air temperature in Celsius over time":

- the rule lands verbatim at line 12 of the rendered prompt; there is no Jinja guard around it in `main.jinja2` or any parent block (`BaseViewAgent`, `Agent`, `Actor`), and nothing keys it to the request
- with `lat` in the table its antecedent is already satisfied, so the rendered prompt instructs a `color`/`detail` split on a request for one aggregated line

Deletion rather than an intent-conditioned rewrite, because deletion is the only variant with measurements behind it (24/30 → 6/30 wrong splits in #2073; the inverse in #2072, 5/30 → 29/30 when the rule is added to the Altair prompt), and GUIDANCE.md wants rules added against observed failures — the zigzag this one guarded against was not reproducible (#1887, 0/100). The multi-line example and the Legends section still carry the split pattern for requests that do ask for several series.

`lumen/tests/ai/test_prompts.py` (371 checks), `test_vega_lite.py`, and `test_agents.py` pass on the branch.

Fixes #2073

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lumen/ai/prompts/VegaLiteAgent/main.jinja2
+++ b/lumen/ai/prompts/VegaLiteAgent/main.jinja2
@@ -9,7 +9,6 @@ Every visualization must:
 - Use `data: {name: {{ memory['table'] }}}` with exact column names from source; a choropleth joined to external boundaries is the one exception (see Key Patterns)
 - Wrap all marks in `layer` array, each with own `mark` and `encoding`
 - Focus on ONE primary metric or relationship per chart
-- If a categorical column separates the rows into more than one series, split on it with `color` (or `detail` when no legend is wanted). A `line` or `area` mark with no split joins points across series and produces a zigzag: two products sharing the same months, drawn as one line, alternate between them.
 - Give the chart its own narrative title *inside the spec*: `title: {text: ..., subtitle: ...}` (see title example). This is the only title that renders above the plot; the `title` field on the chart entry is a tab label and never appears in the chart itself.
   - `title.text` states the finding, not the variables: "Revenue fell 12% after the March price change", NOT "Revenue by Month". If it contains no claim, it is not a title yet.
   - `title.subtitle` adds a second fact from the data ("Enterprise accounts drove the whole decline"). Never describe the chart's own construction — "Stacked bars show revenue by segment" tells the reader what they can already see. Keep it under ~60 characters or it clips.
--- a/lumen/ai/prompts/VegaLiteAgent/main.jinja2
+++ b/lumen/ai/prompts/VegaLiteAgent/main.jinja2
@@ -9,6 +9,7 @@ Every visualization must:
 - Use `data: {name: {{ memory['table'] }}}` with exact column names from source; a choropleth joined to external boundaries is the one exception (see Key Patterns)
 - Wrap all marks in `layer` array, each with own `mark` and `encoding`
 - Focus on ONE primary metric or relationship per chart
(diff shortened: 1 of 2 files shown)


lumen/ai/prompts/VegaLiteAgent/main.jinja2
```diff
@@ -9,7 +9,6 @@ Every visualization must:
 - Use `data: {name: {{ memory['table'] }}}` with exact column names from source; a choropleth joined to external boundaries is the one exception (see Key Patterns)
 - Wrap all marks in `layer` array, each with own `mark` and `encoding`
 - Focus on ONE primary metric or relationship per chart
-- If a categorical column separates the rows into more than one series, split on it with `color` (or `detail` when no legend is wanted). A `line` or `area` mark with no split joins points across series and produces a zigzag: two products sharing the same months, drawn as one line, alternate between them.
 - Give the chart its own narrative title *inside the spec*: `title: {text: ..., subtitle: ...}` (see title example). This is the only title that renders above the plot; the `title` field on the chart entry is a tab label and never appears in the chart itself.
```
ghostiee-11, line 13: This drops the only guard against the #1887 zigzag entirely, not just the over-triggering case #2073 measured. Condition it instead of removing it, split or aggregate, never leave rows un-aggregated and un-split.

```suggestion
- If rows share an x-value only because of an un-aggregated categorical column, either aggregate across it (e.g. `transform: [{aggregate: [...], groupby: [x]}]`) for a single overall series, or split it with `color`/`detail` for a breakdown. Leaving it un-aggregated and un-split joins points across categories into one zigzagging line.
- Give the chart its own narrative title *inside the spec*: `title: {text: ..., subtitle: ...}` (see title example). This is the only title that renders above the plot; the `title` field on the chart entry is a tab label and never appears in the chart itself.
```

ghostiee-11 (approved):

ghostiee-11:
Thanks @7487 for the fix :)

PR #2077, opened 2026-08-31
Preserve GeoDataFrame CRS at DuckDBSource ingest points

### LINKED ISSUE #1904: GeoDataFrame CRS is dropped through DuckDBSource's WKB roundtrip
### Summary

`DuckDBSource.get()` returns a `GeoDataFrame`, but its CRS is always `None`, even when the underlying `GEOMETRY` data is in a known CRS (e.g. EPSG:4326). The geometry is round-tripped through WKB (`ST_AsWKB(...)` then `GeoSeries.from_wkb(...)`), and WKB carries no CRS, so it is lost.

This matters now that #1903 routes geometry columns to maps: `get_dataframe_schema` reports a `crs`/`geographic` flag that the map agents use to choose between a basemap map (DeckGL) and a flat 2D plot (hvPlot). With the CRS dropped, geographic (lat/lon) data reads as `geographic: False` and mis-routes.

### Reproduce

```python
from lumen.util import get_dataframe_schema

# a duckdb file whose `countries` table has a GEOMETRY column in EPSG:4326
src = DuckDBSource(uri="geo.db")
gdf = src.get("countries")
print(gdf.crs)  # -> None   (expected EPSG:4326)

props = get_dataframe_schema(gdf)["items"]["properties"]["geometry"]
print(props)    # -> {..., 'crs': None, 'geographic': False}
```

### Cause

`DuckDBSource._fetch_df` rebuilds geometry with:

```python
df[col] = gpd.GeoSeries.from_wkb(df[col].apply(bytes))
```

`from_wkb` defaults to `crs=None`, and DuckDB's `GEOMETRY` type is SRID-less, so the CRS cannot be recovered from the fetched bytes.

### Suggested fix

Track the CRS at the Lumen level rather than trying to read it back from DuckDB: capture it when the data is ingested (`read_geo_file` / the `GeoDataFrame` handed to `DuckDBSource` already knows it), persist it as table metadata on the source, and re-apply it in `_fetch_df` via `GeoSeries.from_wkb(..., crs=...)`.

### Impact

- Correct DeckGL-vs-hvPlot routing for geographic data (#1903).
- Enables reprojection and correct multi-layer overlays.

Found while testing #1903 (geopandas end-to-end). Repro verified locally.

7487:
The reapply half of this is already in place since #1996: `_fetch_df` rebuilds geometry from WKB with the CRS taken from the column type (`GEOMETRY('EPSG:4326')`) or the `geometry_crs` param. What was still missing is the capture half at the ingest points that know the CRS but dropped it:

- **A GeoDataFrame handed to `DuckDBSource` (`from_df` or `mirrors`) crashed outright** with `Not implemented Error: Data type 'geometry' not recognized` — DuckDB's pandas scanner does not understand the geometry dtype. Both paths now route through a shared `_ingest_table`, which materializes geometry columns via WKB + `ST_GeomFromWKB` into a native GEOMETRY table (a view over the registered frame would be invisible to the cursors queries run on) and records the frame's CRS on `geometry_crs` for `_fetch_df` to reapply. This also unblocks `from_df` callers like the source agent and ingest results when an action returns geographic data.
- **`UI._resolve_geo_source` discarded the CRS `read_geo_file` captures**: it forwarded the `initializers` from `source_params` but not `geometry_crs`, so a geospatial file passed at startup (the gh-1900 path) came back with `crs=None` and `get_dataframe_schema` reported `geographic: False`, mis-routing lat/lon data to a flat plot instead of a basemap.

Before/after on the startup path:

```python
src = UI._resolve_data(['counties.geojson'])[0]  # file is EPSG:4326
src.get('counties').crs   # None -> EPSG:4326
```

An explicitly set `geometry_crs` still wins over the ingested one, and a frame without geometry columns takes the exact same `from_df().to_view()` path as before.

Tests cover all three paths (`from_df`, `mirrors`, startup file); the skip guard for a missing spatial extension only triggers on spatial-related errors so a regression of the geometry-dtype crash still fails instead of skipping. Ran `test_duckdb.py` (79 passed) and `test_ui.py` (88 passed, 5 pre-existing skips) locally; `ruff`/`isort` clean against the repo baseline.

Fixes #1904

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -217,7 +217,7 @@ class DuckDBSource(BaseSQLSource):
         geometry columns pass through WKB and ST_GeomFromWKB into a native
         GEOMETRY table instead. WKB carries no CRS, so each column's CRS is
         recorded on its GEOMETRY('<crs>') column type for `_fetch_df` to
-        reapply after fetching (gh-1904).
+        reapply after fetching.
         """
         geom_cols = geometry_columns(df)
         if not geom_cols:
--- a/lumen/sources/duckdb.py
+++ b/lumen/sources/duckdb.py
@@ -34,6 +34,11 @@ if TYPE_CHECKING:
 GEOMETRY_CRS = re.compile(r"GEOMETRY\('(.+)'\)")
 
 
+def _quote_ident(ident: str) -> str:
+    """Quote a SQL identifier, escaping embedded double quotes."""
+    return '"{}"'.format(ident.replace('"', '""'))
+
+
 class DuckDBSource(BaseSQLSource):
     """
     DuckDBSource provides a simple wrapper around the DuckDB SQL
@@ -77,8 +82,8 @@ class DuckDBSource(BaseSQLSource):
     geometry_crs = param.String(default=None, allow_None=True, doc="""
         CRS to reapply to geometry columns after the WKB roundtrip through
         DuckDB, used as a fallback when the column type carries no CRS of its
-        own. Populated from the source data at ingest; may also be set
-        explicitly for a known dataset.""")
+        own. May be set explicitly for a known dataset; an ingested frame
+        instead records its CRS on the GEOMETRY column type itself.""")
 
     read_only = param.Boolean(default=None, doc="""
         Whether to open the DuckDB database in read-only mode.""")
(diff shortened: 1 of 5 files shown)


lumen/sources/duckdb.py
```diff
@@ -188,21 +189,55 @@ def update(e, table=table):
                     df = as_pandas(e.new)
                     for col in df.select_dtypes(include=['string']).columns:
                         df[col] = df[col].astype(object)
-                    self._connection.from_df(df).to_view(table)
+                    self._ingest_table(table, df)
```
ghostiee-11, line 197: No try/except here, unlike the init loop below. A mirror flipping between geometry and non-geometry data raises an uncaught CatalogException (view vs table).

lumen/sources/duckdb.py
```diff
@@ -188,21 +189,55 @@ def update(e, table=table):
...
+        selected = ', '.join(
+            f'ST_GeomFromWKB("{c}") AS "{c}"' if c in geom_cols else f'"{c}"'
+            for c in df.columns
+        )
+        self._connection.execute('INSTALL spatial;\nLOAD spatial;')
+        self._connection.register(f'{name}__wkb', wkb)
+        # Materialized as a real table: a view over the registered frame would
+        # be invisible to the cursors queries run on.
+        self._connection.execute(
+            f'CREATE OR REPLACE TABLE "{name}" AS SELECT {selected} FROM "{name}__wkb"'
```
ghostiee-11: Name and column go into raw SQL unescaped for embedded quotes, unlike the existing to_view path.

lumen/sources/duckdb.py
```diff
@@ -188,21 +189,55 @@ def update(e, table=table):
...
+        )
+        self._connection.unregister(f'{name}__wkb')
+        if crs is not None and not self.geometry_crs:
```
ghostiee-11: geometry_crs is set once and never revisited. A source with multiple geometry tables of different CRS silently mislabels every table after the first.

lumen/sources/duckdb.py
7487: Good catch — the flip raised in both directions, since CREATE OR REPLACE only replaces a relation of its own kind. Fixed at the root in `_ingest_table` rather than wrapping the callback: it now drops a leftover relation of the other kind before creating (45114a6), so neither the mirror callback nor the init loop can hit the CatalogException — and a try/except in the callback would have silently dropped the update instead. Covered by `test_duckdb_ingest_flips_between_geometry_and_plain`.

lumen/sources/duckdb.py
7487: Fixed in 45114a6: identifiers now go through a `_quote_ident` helper (embedded quotes doubled) everywhere they are interpolated — here and in the `_fetch_df` WKB wrapper, which had the same hole on the read side. `test_duckdb_geometry_ingest_quotes_identifiers` round-trips a column named `na"me`.

lumen/sources/duckdb.py
7487: Right — first-wins mislabeled every later table. DuckDB can carry the CRS on the column type itself (`GEOMETRY('EPSG:4326')`), which `_fetch_df` already prefers over the param, so 45114a6 bakes each column's CRS into the type at ingest via `ST_GeomFromWKB(...)::GEOMETRY('<crs>')` and drops the param mutation entirely; `geometry_crs` stays as an explicit fallback only. `test_duckdb_from_df_geodataframes_keep_distinct_crs` covers two tables with different CRS (4326 + 3857).

ghostiee-11 (approved):

PR #2081, opened 2026-09-02
Feat validation heuristic

### LINKED ISSUE #2080: ValidationAgent: previous_keys conflates stale context with reused-but-still-relevant context
ValidationAgent._gather_prompt_context (lumen/ai/agents/validation.py) computes previous_keys as any of chat/sql/view/listing present in context but missing from the current plan's combined task.out_context. A plan that reuses a prior sql/view as input without a task redeclaring it in out_context gets the same previous_keys label as genuinely stale, unrelated context.

#2074's gating (`and 'sql' not in previous_keys`, same for view/listing/chat) fully skips strict validation whenever a key lands in previous_keys, so a plan that legitimately depends on a reused sql/view now has that value validated only by the generic response check, not the per-key correctness check. That's an opposite-direction regression from the false positives #2074 fixes: a genuinely wrong GROUP BY or chart mark could pass validation as long as the task that produced it isn't re-run this plan.

Worth distinguishing "stale/irrelevant" from "reused/still authoritative" rather than gating purely on out_context membership.

Related to #2074.

DevForge255:
## Description
This PR implements a granular, rule-based heuristic to intelligently tag `previous_keys` (stale context) in the `ValidationAgent`. 

**Motivation:**
Previously, any data (`sql`, `view`, `chat`) not produced in the current plan was blanket-tagged as `previous_keys`. This created a dilemma: if we completely gated out validation for `previous_keys`, the agent would fail to validate old context that was being reused as an input for a new prompt. On the other hand, if we didn't gate it, leftover context triggered false-positive `ValidationFault`s when strictly evaluated against completely new, unrelated queries.

**Changes:**
- Added category-specific heuristic keywords (SQL, View, Chat, General) in `validation.py`.
- The `_gather_prompt_context` now extracts the user's latest query text and tokenizes it.
- If the user uses a continuation keyword (like *"color"*, *"filter"*, *"it"*, *"elaborate"*), the agent dynamically identifies which contexts the user is trying to reuse/modify and keeps them **active** (i.e., does not tag them as `previous_keys`), while safely marking unrelated context as stale.

**Tests Run:**
- Successfully ran `pytest lumen/tests/ai/test_validation_agent.py` and `pytest lumen/tests/ai/test_prompts.py`. All tests passed.

Fixes #2080 

## AI Disclosure

Tool & Model: Antigravity IDE (Gemini)
Usage: Used to brainstorm the architectural approach for the heuristic tagging, write the Python regex parsing logic for the context gatherer, and draft this PR description.

## Checklist

### DIFF
--- a/lumen/ai/agents/validation.py
+++ b/lumen/ai/agents/validation.py
@@ -1,5 +1,3 @@
-import re
-
 from typing import Any, NotRequired
 
 import param
@@ -50,6 +48,33 @@ class ValidationOutputs(ContextModel):
     validation_result: str
 
 
+def get_plan_required_keys(plan):
+    """
+    Check which context keys are required by the tasks in the plan.
+
+    Iterates over each task's ``input_schema`` annotations and checks
+    whether any task declares ``sql``, ``view``, ``chat``, or ``listing``
+    as an input dependency.
+
+    Parameters
+    ----------
+    plan : list[ActorTask]
+        The list of tasks produced by the planner.
+
+    Returns
+    -------
+    dict[str, bool]
+        A dictionary mapping each context key to ``True`` if any task
+        in the plan requires it, ``False`` otherwise.
+    """
+    context_keys = {"sql", "view", "chat", "listing"}
+    all_input_keys = set()
+    for task in plan:
+        annotations = set(task.input_schema.__annotations__.keys())
+        all_input_keys |= annotations
+    return {key: key in all_input_keys for key in context_keys}
+
+
 class ValidationAgent(Agent):
     """
     ValidationAgent focuses solely on validating whether the executed plan
@@ -94,39 +119,15 @@ class ValidationAgent(Agent):
         """
         ctx = await super()._gather_prompt_context(prompt_name, messages, context, **kwargs)
 
-        # Extract the latest user query
-        user_messages = [msg for msg in reversed(messages) if msg.get("role") == "user"]
-        query_text = content_to_text(user_messages[0].get("content", "")).lower() if user_messages else ""
-        query_words = set(re.findall(r'\b\w+\b', query_text))
-
-        # Define category-specific continuation keywords
-        general_keywords = {"it", "this", "that", "these", "those", "also", "and", "now", "previous"}
-        sql_keywords = {"filter", "sort", "limit", "top", "include", "group"}
-        view_keywords = {"color", "label", "chart", "plot", "axis", "instead", "make"}
-        chat_keywords = {"elaborate", "translate", "explain", "why", "point"}
-
-        # Determine which categories are requested to be kept
-        keep_sql = bool(query_words.intersection(sql_keywords | general_keywords))
-        keep_view = bool(query_words.intersection(view_keywords | general_keywords))
-        keep_chat = bool(query_words.intersection(chat_keywords | general_keywords))
-        keep_listing = bool(query_words.intersection(general_keywords))
-
         plan = context.get("plan")
         previous_keys = set()
         if plan is not None:
+            required_keys = get_plan_required_keys(plan)
             produced = {k for task in plan for k in task.out_context}
             for key in ("chat", "sql", "view", "listing"):
                 if key in context and key not in produced:
-                    # Check if we should KEEP this key (i.e., NOT mark as stale)
-                    if key == "sql" and keep_sql:
+                    if required_keys.get(key, False):
                         continue
-                    if key == "view" and keep_view:
-                        continue
-                    if key == "chat" and keep_chat:
-                        continue
-                    if key == "listing" and keep_listing:
-                        continue
-
                     previous_keys.add(key)
         ctx["previous_keys"] = previous_keys
         return ctx
--- a/lumen/ai/agents/validation.py
+++ b/lumen/ai/agents/validation.py
@@ -6,7 +6,7 @@ from panel_material_ui import Button
 from pydantic import Field
 
 from ..config import PROMPTS_DIR
-from ..context import ContextModel, TContext
+from ..context import ContextModel, TContext, input_dependency_keys
 from ..llm import Message
 from ..models import BaseModel
 from ..utils import content_to_text, log_debug
@@ -70,8 +70,7 @@ def get_plan_required_keys(plan):
     context_keys = {"sql", "view", "chat", "listing"}
     all_input_keys = set()
     for task in plan:
-        annotations = set(task.input_schema.__annotations__.keys())
-        all_input_keys |= annotations
+        all_input_keys |= input_dependency_keys(task.input_schema)
     return {key: key in all_input_keys for key in context_keys}
 
 
@@ -122,7 +121,9 @@ class ValidationAgent(Agent):
         plan = context.get("plan")
         previous_keys = set()
         if plan is not None:
-            required_keys = get_plan_required_keys(plan)
+            required_keys = get_plan_required_keys(
+                task for task in plan if task.actor is not self
+            )
             produced = {k for task in plan for k in task.out_context}
             for key in ("chat", "sql", "view", "listing"):
                 if key in context and key not in produced:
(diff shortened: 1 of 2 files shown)


DevForge255:
Hi @ghostiee-11, 

The PR is ready! I noticed that one of the CI checks (`unit:test-312:windows-latest`) failed with a DuckDB `httpfs` extension IO Error in `test_ui.py`. 

Since my changes are only in `validation.py` and all other tests passed, this looks like an unrelated environment/network issue with the Windows runner downloading the extension. Could you please re-run that specific failed job when you have a moment?

lumen/ai/agents/validation.py
```diff
@@ -92,12 +94,39 @@ async def _gather_prompt_context(self, prompt_name, messages, context, **kwargs)
         """
         ctx = await super()._gather_prompt_context(prompt_name, messages, context, **kwargs)
 
+        # Extract the latest user query
+        user_messages = [msg for msg in reversed(messages) if msg.get("role") == "user"]
+        query_text = content_to_text(user_messages[0].get("content", "")).lower() if user_messages else ""
+        query_words = set(re.findall(r'\b\w+\b', query_text))
+
+        # Define category-specific continuation keywords
+        general_keywords = {"it", "this", "that", "these", "those", "also", "and", "now", "previous"}
```
ghostiee-11: Using common words like `and` keeps unrelated stale context active; could this use plan provenance instead of keywords?

lumen/ai/agents/validation.py
```diff
@@ -92,12 +94,39 @@ async def _gather_prompt_context(self, prompt_name, messages, context, **kwargs)
...
+        sql_keywords = {"filter", "sort", "limit", "top", "include", "group"}
+        view_keywords = {"color", "label", "chart", "plot", "axis", "instead", "make"}
+        chat_keywords = {"elaborate", "translate", "explain", "why", "point"}
+
+        # Determine which categories are requested to be kept
+        keep_sql = bool(query_words.intersection(sql_keywords | general_keywords))
```
ghostiee-11: Clear reuse such as "use the same query" still marks `sql` stale, so this does not fully resolve #2080.

lumen/ai/agents/validation.py
```diff
@@ -92,12 +94,39 @@ async def _gather_prompt_context(self, prompt_name, messages, context, **kwargs)
         """
         ctx = await super()._gather_prompt_context(prompt_name, messages, context, **kwargs)
 
+        # Extract the latest user query
```
ghostiee-11: Could you add parameterized regression tests for stale and reused context that fail on `main`?

ghostiee-11 (changes requested):
The heuristic still misclassifies both unrelated stale context and explicit reuse, and needs regression coverage before merge.

DevForge255:
Hi @ghostiee-11, I've addressed both feedback points:

1. **Replaced keyword-based heuristic with plan provenance** — instead of 
   guessing from user query words, we now check `task.input_schema.__annotations__` 
   directly to see which context keys the plan actually requires.

2. **Added parameterized regression tests** — `test_get_plan_required_keys` 
   (5 cases) and `test_previous_keys_provenance` (4 cases) that would fail on 
   main but pass on this branch.

lumen/ai/agents/validation.py
DevForge255: Great point! You're absolutely right — common words like `and` are too noisy 
and would incorrectly keep unrelated stale context active.

Honestly, I wasn't fully aware of the plan's internal structure (`input_schema` 
annotations on each task) before your suggestion — your hint about "plan 
provenance" was really eye-opening and helped me understand the architecture 
much better!

I've replaced the entire keyword-based approach by inspecting 
`task.input_schema.__annotations__` for each task in the plan. A context key 
is only kept active if at least one task declares it as an input dependency. 
Much cleaner and more reliable than guessing from text. Thanks a lot! 🙏

lumen/ai/agents/validation.py
DevForge255: Completely agree. Phrases like "use the same query" were not caught by the 
keyword heuristic, so `sql` was incorrectly marked stale. The new plan 
provenance approach fixes this at the root — instead of parsing user text, 
we check whether the plan's tasks actually need `sql` as an input. 
If `SQLAgent` is in the plan, `sql` stays active. No text parsing needed at all.

lumen/ai/agents/validation.py
DevForge255: Done! Added two sets of parameterized tests in `test_validation_agent.py`:

1. `test_get_plan_required_keys` — 5 cases covering SQLAgent-only, ViewAgent-only, 
   combined plans, ChatAgent, and empty input schema.

2. `test_previous_keys_provenance` — 4 regression cases (stale sql, reused sql, 
   all stale, mixed) that verify the tagging logic end-to-end.

All 9 new tests pass on this branch and would fail on `main` since 
`get_plan_required_keys` doesn't exist there. Thanks for the suggestion! 🙏

DevForge255:
Hi @ghostiee-11, I've addressed all three review comments:
- Replaced keyword heuristics with plan provenance (`input_schema` annotations)
- Added 9 parameterized regression tests that fail on main

Could you please re-review when you get a chance? 🙏

DevForge255:
Thanks @ghostiee-11 for the direct fixes! 🙏

I learned a couple of things from your changes:

1. I wasn't aware of the existing `input_dependency_keys()` helper — using 
   `.__annotations__` directly was fragile in comparison.

2. The self-exclusion (`task.actor is not self`) was a blind spot for me — 
   since `ValidationAgent` is part of the plan itself, its own `ValidationInputs` 
   would have incorrectly marked everything as "required", so nothing would 
   ever be stale.

3. Testing with real `SQLAgent`/`ChatAgent` instances instead of mock TypedDicts 
   is much more robust.

Really appreciate you taking the time to polish this! 🙌

ghostiee-11:
Yup...
Your fix was also good, Just fixed somethings on top of your fix:)
Thanks for taking this up.

ghostiee-11 (approved):

DevForge255:
Also, congratulations on completing GSoC! 🎉

If you're open to it, I'd love to stay in touch and learn from your 
experience with this codebase. Are you planning to continue contributing 
to lumen, or moving on to other projects?

Since you've worked deeply on this codebase, I'd love some guidance if you 
have a moment:
- Are there any areas of the codebase you'd recommend exploring next 
  (e.g. coordinator flow, context propagation, prompt design)?
- Any habits or practices that helped you ramp up quickly as a contributor?
- Is there anything I should be doing differently in how I approach issues 
  or PRs?
No pressure at all — any advice would be really valuable as I'm trying 
to grow as a contributor here! 😊

ghostiee-11:
Thanks 
I will keep on contributing to Lumen :) 
Love to work over this and manage it. Wanna see Lumen grow exponentially and people using it for good ways :)

Will recommend firstly try to read the docs of lumen and then, try running lumen ai serve and try using different-different functionality of this :) 
This is a really cool platform

ghostiee-11:
I will suggest you, try using lumen more and more and then brainstorm any feature for making it accessible/enhance lumen in any way and can impact users :)

ghostiee-11:
Anyways, Thanks for contributing :) 
Happy to see you active here..

DevForge255:
That's really great to hear — lumen is genuinely a cool platform and 
having passionate contributors like you makes a huge difference! 🙌

Thanks for the advice! I've already been running `lumen-ai serve` locally 
and exploring the codebase, but I think I need to use it more as an actual 
user — not just a contributor — to really understand the pain points.

I'll start reading the docs more carefully and try to brainstorm features 
that could make lumen more accessible. Will definitely ping you if I find 
something interesting to work on! 😊

---

Also, I'm aiming to apply for GSoC next year with lumen/holoviz. 
Since you just went through the whole journey, would love your perspective:

- How early did you start contributing before GSoC applications opened?
- What made your proposal stand out — was it the number of PRs, the quality 
  of contributions, or something else?
- Is there any specific area of lumen you'd suggest focusing on that could 
  make for a strong GSoC project idea?
- Any mistakes you'd avoid if you were starting over?

Totally understand if you're busy — even a couple of lines would mean a lot! 🙏

PR #2083, opened 2026-09-04
Fix hvPlotUIView controls clipping on short windows (#2066)

### LINKED ISSUE #2066: hvPlotUIView clips the explorer controls on a short window with no way to scroll
## Description

`hvPlotUIView` renders `hvDataFrameExplorer` straight into the exploration panel, with nothing to scroll it. The explorer's control stack is tall and its height is fixed by the surrounding card, so on a short window the lower controls are cut off and cannot be reached at all.

At a 1280x560 viewport the `Advanced` tab label is cut mid-word, `By` and `Groupby` are entirely off-screen, and the plot's x-axis is not visible. Nothing scrolls:

```
document.documentElement.scrollHeight  560
document.documentElement.clientHeight  560
window.scrollTo(0, 5000)  ->  window.scrollY stays 0
```

No inner container scrolls either, so those controls are simply lost rather than merely off-screen. At 1440x900 the same view fits, which is why this is easy to miss.

## Reproduce

```
lumen-ai serve cities.csv --provider openai --model gpt-4.1 --agents hvplot
```

Ask for any hvPlot chart, then resize the browser window so its height is around 560px. `hvPlotUIView` is opt-in via `--agents hvplot`; a dashboard with `type: hvplot_ui` shows the same thing.

Note that reaching this needs #2064, without which the hvPlot agent cannot render at all.

## Expected

The explorer scrolls within its panel, or is given a height that keeps its controls reachable.

rimsha152007-dotcom:
## Description
Added a _fit_explorer() helper that wraps the explorer's layout in a pn.Column with sizing_mode='stretch_both' and scroll='y-auto', and applies min_height=500 to the explorer's internal layout so it always has enough room to render properly.get_panel() is left untouched and still returns the raw hvGridExplorer / hvDataFrameExplorer object as before. The scroll/sizing fix is instead applied in __panel__(), which is where the view is actually rendered for display — so the fix only affects the rendered output, not the public contract of get_panel().

Fixes #2066 

## AI Disclosure

Tool & Model:Gemini 3.6 Flash
Usage: Assisted in CI fix and testing.

### DIFF
--- a/lumen/views/base.py
+++ b/lumen/views/base.py
@@ -1162,13 +1162,20 @@ class hvPlotUIView(hvPlotBaseView):
         return (data,), dict(params, **self.kwargs)
 
     def __panel__(self):
-        panel = self.get_panel()
+        layout = self.get_panel()
+        explorer = layout[0] if isinstance(layout, pn.Column) else layout
         def ui(*events):
             gridded = self._source_dataset()
-            panel._data = gridded if gridded is not None else self.get_data()
-            panel._plot()
-            return panel
+            explorer._data = gridded if gridded is not None else self.get_data()
+            explorer._plot()
+            return layout
         return pn.bind(ui, self.param.rerender)
+
+    def _fit_explorer(self, explorer):
+        layout = getattr(explorer, '_layout', None)
+        if layout is not None:
+            layout.param.update(height=None, min_height=500, sizing_mode='stretch_both')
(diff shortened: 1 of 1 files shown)


ghostiee-11:
Use https://github.com/holoviz-dev/holoviz-skills/blob/main/contributing-to-holoviz/skills/pr-description/SKILL.md to write PR desc

ghostiee-11 (changes requested):
Fix Ci and update the PR desc firstly :)

PR #2085, opened 2026-09-06
fix: preserve artifacts for a session (#1988)

### LINKED ISSUE #1988: Preserve artifacts for a session
when Browser tab reloads, all things generated disappear. Preserve the artifacts for a session.

https://panel.holoviz.org/how_to/server/reconnect.html

rimsha152007-dotcom:
## Description
Adds an opt-in `persist_session=True` param that saves explorations
to SQLite as you work, and restores them when you return to the
same session.

**Stable session IDs** — replaced `id(self)` (changes every time)
with a UUID read from/written to the URL (`?session_id=...`), so a
session can be found again later.

**Reconnect handling** — added a `reconnect` param, set
`pn.config.reconnect`/`notifications` in `_configure_session`, and
wrapped `_busy` with `state.block_expiration()` so brief connection
drops or long-running plans don't kill the session outright.

**Persistence store (`logs.py`)** — new `explorations` table plus
`upsert_exploration()`, `load_session()`, `delete_exploration()`,
`delete_stale(ttl_seconds)`, following the same upsert pattern as
the existing chat `logs` table.

**Serializable explorations (`ui.py`)** — added `exploration_id` and
`to_spec()`/`from_spec()` to `Exploration`, reusing each
source's/view's own `to_spec()`. Only `context`, `views`, and
serialized `conversation` are saved — the live `plan` and LLM
objects are excluded, since they can't be reconstructed.

**Save on change** — new `_persist_state()`, called from
`_add_exploration`, `_delete_exploration`, `_cleanup_exploration`,
and after each finished plan in `_update_conversation`. `_destroy`
now persists before clearing, and skips in-memory cleanup entirely
when `persist_session=True`.

**Restore on load** — `_configure_session` cleans up stale rows and
registers `state.onload(self._restore_session)`. `_restore_session`
rebuilds each saved `Exploration`, its `Tabs`/`MultiSplit` view, and
its parent/child links, then restores the chat feed and active
exploration from the most recent entry.

### New params on `UI`
- `reconnect` (default `True`)
- `persist_session` (default `False`, opt-in)
- `session_ttl` (default `86400`s)

Fixes #1988 

## AI Disclosure

Tool & Model: claude sonnet 5 medium 
Usage:Assisted in understanding existing code base, implementation 

### DIFF
--- a/lumen/ai/ui.py
+++ b/lumen/ai/ui.py
@@ -5,6 +5,7 @@ import atexit
 import os
 import tempfile
 import traceback
+import uuid
 
 from contextlib import contextmanager
 from functools import partial
@@ -13,6 +14,7 @@ from pathlib import Path
 from types import SimpleNamespace
 from typing import Any
 
+import panel as pn
 import param
 
 from panel.chat.feed import PLACEHOLDER_SVG
@@ -448,6 +450,16 @@ class UI(Viewer):
     logs_db_path = param.String(default=None, constant=True, doc="""
         The path to the log file that will store the messages exchanged with the LLM.""")
 
+    reconnect = param.Selector(default=True, objects=[True, False, "prompt"], doc="""
+    Whether to automatically re-connect to the server if the connection drops.""")
+
+    session_ttl = param.Integer(default=86400, doc="""
+    Seconds to keep persisted explorations before cleaning them up.""")
+
+    persist_session = param.Boolean(default=False, doc="""
+    Whether to persist explorations so they survive a page reload.
+    Requires logs_db_path to be set.""")
+
     notebook_preamble = param.String(default='', doc="""
         Preamble to add to exported notebook(s).""")
 
@@ -775,10 +787,26 @@ class UI(Viewer):
         Configure the Panel session state.
         """
         log_debug("New Session: \033[92mStarted\033[0m", show_sep="above")
-        self._session_id = id(self)
+        self._session_id = self._resolve_session_id()
+        if state.curdoc and state.curdoc.session_context:
+            pn.config.reconnect = self.reconnect
+            pn.config.notifications = True
         state.onload(self._initialize_new_llm)
         if state.curdoc and state.curdoc.session_context:
             state.on_session_destroyed(self._destroy)
+
+    def _resolve_session_id(self) -> str:
+        args = state.session_args or {}
+        sid = args.get("session_id")
+        if sid:
+            return sid[0].decode() if isinstance(sid[0], bytes) else str(sid[0])
+        sid = uuid.uuid4().hex
+        def _set_query():
+            if state.location:
+                state.location.update_query(session_id=sid)
+
+        state.onload(_set_query)
+        return sid
 
     def _configure_logs(self, interface):
         if not self.logs_db_path:
@@ -1533,12 +1561,18 @@ class UI(Viewer):
         self._idle.clear()
         with edit_readonly(self._page):
             self._page.busy = True
+        server = bool(state.curdoc and state.curdoc.session_context)
         try:
+            if server:
+                state.block_expiration()
             yield
         finally:
+            if server:
+                state.unblock_expiration()
             self._idle.set()
             with edit_readonly(self._page):
                 self._page.busy = old_busy
+
 
     def _update_help_getting_started(self):
         """Update the Getting Started help text based on whether data sources are connected."""
@@ -1864,7 +1898,6 @@ class UI(Viewer):
         server = bool(state.curdoc and state.curdoc.session_context)
         return self._create_view(server=server).servable(title, **kwargs)
 
-
 class ChatUI(UI):
     """
     ChatUI provides a high-level entrypoint to start chatting with your data
@@ -1890,7 +1923,6 @@ class ChatUI(UI):
     provider_choices = param.Dict(default={}, doc="""
         Available LLM providers to show in the configuration dialog.""")
 
-
 class Exploration(param.Parameterized):
 
     context = param.Dict(allow_refs=True)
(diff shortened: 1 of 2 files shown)


lumen/ai/ui.py
```diff
@@ -1908,10 +1940,43 @@ class Exploration(param.Parameterized):
...
         return self.view
 
+    def to_spec(self) -> dict:
+        ctx = self.context or {}
+        return {
+            "title": self.title,
+            "subtitle": self.subtitle,
+            "sources": [src.to_spec() for src in ctx.get("sources", [])],
+            "visible_slugs": sorted(ctx.get("visible_slugs", [])),
+            "views": [v.to_spec() for v in (self.plan.views if self.plan else [])],
```
ghostiee-11, line 1955: `plan.views` contains LumenEditor objects, which do not implement `to_spec()`, so persistence fails once outputs exist. Could this use the editor serialization API?

lumen/ai/ui.py
```diff
@@ -1908,10 +1940,43 @@ class Exploration(param.Parameterized):
...
+            "conversation": [msg.serialize() for msg in self.conversation],
+            "exploration_id": self.exploration_id,
+        }
+
+    @classmethod
+    def from_spec(cls, spec: dict, view=None) -> Exploration:
```
ghostiee-11, line 1961: The saved views are never restored, leaving every restored exploration with empty tabs. Could we rebuild the editors and repopulate the tabs here?

lumen/ai/ui.py
```diff
@@ -1908,10 +1940,43 @@ class Exploration(param.Parameterized):
...
 
     def __panel__(self):
         return self.view
 
+    def to_spec(self) -> dict:
+        ctx = self.context or {}
+        return {
+            "title": self.title,
+            "subtitle": self.subtitle,
+            "sources": [src.to_spec() for src in ctx.get("sources", [])],
```
ghostiee-11, line 1953: `Source.to_spec()` may include passwords, tokens, private keys, or credentials in connection URLs. We should not store these unencrypted in SQLite.

lumen/ai/ui.py
```diff
@@ -2401,6 +2571,9 @@ async def _delete_exploration(self, item):
                 self._explorations.update_item(parent, items=[it for it in parent["items"] if it is not item])
                 self._explorations.value = parent
             self.interface.objects = []
+            if self.persist_session and self._logs:
+               self._logs.delete_exploration(item["view"].exploration_id)
```
ghostiee-11, line 2575: This deletes only the selected exploration; nested descendants remain in SQLite and later reappear at the root. Could deletion remove the complete subtree?

lumen/ai/ui.py
```diff
@@ -1908,10 +1940,43 @@ class Exploration(param.Parameterized):
...
+            "exploration_id": self.exploration_id,
+        }
+
+    @classmethod
+    def from_spec(cls, spec: dict, view=None) -> Exploration:
+        sources = [Source.from_spec(s) for s in spec.get("sources", [])]
+        context = {"sources": sources, "visible_slugs": set(spec.get("visible_slugs", []))}
+        if sources:
+            context["source"] = sources[-1]
+        conversation = [
```
ghostiee-11, line 1966: `serialize()` returns only message content, so restored Agent messages become User messages and lose metadata. Could we persist and restore the message fields?

ghostiee-11:
Thanks for picking up this issue @rimsha152007-dotcom :)

rimsha152007-dotcom:
Thanks so much for the thorough review, @ghostiee-11 these are all
really valuable catches, especially the credentials one. I'll work to fix each issue.

ghostiee-11 (changes requested):

rimsha152007-dotcom:
Hi @ghostiee-11,

I've updated the PR to address all feedback.

View Serialization & Restoration: Updated to_spec() to call LumenEditor._serialize_component() directly on underlying components. Restored views in from_spec() are now properly wrapped in a Plan object so editor tabs populate on reload.

Message Metadata: to_spec() now captures {"user": msg.user, "object": msg.object} and from_spec() reinstantiates ChatMessage(object=..., user=...), preserving sender identity and styling.

Subtree Cleanup: delete_exploration() now recursively deletes child/descendant records before deleting the parent node
.
Credential Redaction: Added _redact_source_spec() to sanitize sensitive keys and connection URIs prior to persistence.

Trade-offs:

Redaction: Redacting passwords means restored sources with inline credentials will require environment variables or manual re-entry to reconnect. Let me know if this trade-off works for v1.

_deserialize_component: LumenEditor._deserialize_component() expects an instantiated Component, so from_spec() uses issubclass(component_cls, Panel) instead. We can update _deserialize_component to accept type[Component] if you'd prefer to call it directly.

lumen/ai/ui.py
```diff
@@ -1908,10 +1957,63 @@ class Exploration(param.Parameterized):
...
+                "yaml_spec": yaml_spec,
+                "spec_dict": spec_dict,
+            })
+        return {
+            "title": self.title,
+            "subtitle": self.subtitle,
+            "sources": [_redact_source_spec(src.to_spec()) for src in ctx.get("sources", [])],
+            "visible_slugs": sorted(ctx.get("visible_slugs", [])),
+            "views": views,
+            "conversation": [{"user": msg.user, "object": msg.object} for msg in self.conversation],
```
ghostiee-11, line 1984: `to_spec()` can end up with a raw Panel object in `ChatMessage.object` (e.g. the planner's steps Card lands in `interface.objects` most turns). `json.dumps` in `upsert_exploration` then throws and gets swallowed by the bare `except Exception: log_debug`, so the save silently no-ops. Worth serializing message objects to a string/dict first, or letting this exception surface instead of eating it.

lumen/ai/ui.py
```diff
@@ -2351,6 +2453,86 @@ def _configure_session(self):
...
+
+        with hold():
+            for row in rows:
+                spec = row["spec"]
+
+                # Rebuild the live Exploration object from its saved spec
+                exploration = Exploration.from_spec(spec)
+
+                # Rebuild its visual container the same way _add_exploration does
+                tabs = Tabs(dynamic=True, sizing_mode="stretch_both")
```
ghostiee-11, line 2493: This builds a fresh empty `Tabs` and assigns it to `exploration.view`, but nothing pushes the reconstructed `LumenEditor` views from `plan.views` into it, since `_add_views` (or its watcher) is never called here. After restore the exploration shows up in the nav tree but the pane is empty.

lumen/ai/ui.py
```diff
@@ -775,11 +792,27 @@ def _configure_session(self):
...
-        self._session_id = id(self)
+        self._session_id = self._resolve_session_id()
+        if state.curdoc and state.curdoc.session_context:
+            pn.config.reconnect = self.reconnect
+            pn.config.notifications = True
         state.onload(self._initialize_new_llm)
         if state.curdoc and state.curdoc.session_context:
             state.on_session_destroyed(self._destroy)
 
+    def _resolve_session_id(self) -> str:
```
ghostiee-11, line 803: This trusts `session_id` straight from the URL query param with no ownership check. Anyone who gets the URL (bookmark, shared link, support ticket) can load and overwrite/delete another user's session. Probably needs to be tied to a signed cookie or the browser's own state rather than an open query param.

lumen/ai/ui.py
```diff
@@ -1890,6 +1928,17 @@ class ChatUI(UI):
     provider_choices = param.Dict(default={}, doc="""
         Available LLM providers to show in the configuration dialog.""")
 
+_REDACT_KEYS = {"password", "token", "api_key", "secret", "access_key", "private_key"}
+
+def _redact_source_spec(spec: dict) -> dict:
```
ghostiee-11, line 1933: This only strips a fixed top-level key set plus an exact `"uri"` match. `SQLAlchemySource` puts its credential in `url`, not `uri`, so that slips through, and anything nested (`conn_kwargs`, `storage_options`, etc.) isn't scanned at all. As written this can write plaintext creds into the persisted SQLite spec.

PR #2090, opened 2026-09-12
Add Mosaic visualization support

### LINKED ISSUE #1985: Propose MosaicAgent as the chart-generating agent
## Summary

@ahuang11 this follows up on the Mosaic/vgplot idea you raised a while back.

I've been contributing to [Vega-Lite](https://github.com/vega/vega-lite), and Lumen happens to use Vega-Lite too, through `VegaLiteAgent`, for chart generation. While working on Vega-Lite stuff I came across [Mosaic](https://github.com/uwdata/mosaic) (the `vgplot` library) and ended up contributing there as well. The more time I spent in Mosaic's codebase, the more it felt like its architecture lines up almost perfectly with how Lumen is already built.

Lumen is built on DuckDB. Mosaic is also built around DuckDB, and its whole design is basically: keep the data in the database, only send query results to the browser instead of shipping raw rows over. `VegaLiteAgent` doesn't do that. It pulls the full result out of the database and embeds it as inline JSON in the chart spec.

So I started thinking, since Mosaic and Lumen already line up so well on the data side, why not build a `MosaicAgent` for Lumen and let it be the agent Lumen reaches for when generating charts?

## How it fits in

Lumen already has this pattern for Vega-Lite: an agent that gets the LLM to write a spec, and a view that renders it. I built `MosaicAgent` to follow the same shape, just targeting Mosaic's spec format instead.

```mermaid
flowchart TD
    U[User asks for a chart] --> A[MosaicAgent]
    A -->|LLM writes a Mosaic spec| S[mosaic-spec YAML]
    S --> V[MosaicView]
    V --> C[Interactive chart]
```

That's the version I've built so far. `MosaicWidget`'s Python API actually gives you two ways to hand it data:
- `con=<a duckdb connection>`: it queries directly against a connection you hand it
- `data={name: df}`: it makes its own fresh `duckdb.connect()` internally and loads your DataFrame into that

Right now I'm using `data=`. There's a further step we could take on top of that: switch to `con=` and point it at Lumen's existing DuckDB connection directly, so there's no copying involved at all. That's not quite as simple as it sounds though, since a Lumen pipeline can apply filters and SQL transforms on top of the raw table, so we'd need to register the pipeline's actual (filtered/transformed) query as a view rather than just handing over the connection and pointing at the raw table. It also only works when the pipeline is backed by `DuckDBSource` specifically; plenty of Lumen sources (Snowflake, Intake, plain files, etc.) don't expose a DuckDB connection at all, so there'd need to be a fallback for those. Still very doable, just more design work than what I've built so far.

```mermaid
flowchart TD
    subgraph Built["Built so far"]
        direction LR
        P1[Pipeline data] -->|copy into a DataFrame| D1[data=]
        D1 --> M1[MosaicWidget's own DuckDB]
    end

    subgraph Next["Possible next step, DuckDBSource only"]
        direction LR
        P2[Pipeline's filtered/transformed query] -->|register as a view| V2[View in Lumen's DuckDB]
        V2 -->|con=| M2[Lumen's existing DuckDB connection]
    end

    Built ~~~ Next
```

Everything else still falls back to the copy path above, since only `DuckDBSource` has a connection to hand over in the first place.

## What this actually involves

- `MosaicAgent`: writes a Mosaic spec, same idea as `VegaLiteAgent` but a different output format
- `MosaicView`: renders it via the [`mosaic-widget`](https://pypi.org/project/mosaic-widget/) widget
- some prompt examples so the LLM actually writes valid specs
- `mosaic-widget` as a dependency

SuMayaBee:
## Description

Adds Mosaic visualization support to Lumen through the published `panel-mosaic-viz` Panel extension.

@philippjfr suggested creating the Panel extension first, so this PR integrates that extension into Lumen and lets Lumen render Mosaic charts.

Fixes #1985 

## AI Disclosure

Tool & Model: OpenAI Codex (GPT-5)  
Usage: Used to help implement, test, and document the integration. I reviewed and tested all changes.

## Checklist

- [x] Added documentation


SuMayaBee:
@ahuang11 I have run the lumen with mosaic with the classic iris dataset, with those qns:

### 1. Create a scatter plot of sepal length versus sepal width, colored by species.
Screenshot_2026-09-16_21-55-21

### 2. Make a box plot of petal length for each species.
Screenshot_2026-09-16_21-57-34

### 3. Show a histogram of petal width, split by species.
Screenshot_2026-09-16_22-00-18

### 4. Create a correlation heatmap for all numeric columns.
Screenshot_2026-09-16_22-02-31

### 5. Make a pair plot of sepal and petal measurements, colored by species.
Screenshot_2026-09-16_22-05-01

### 6. Show the average sepal length and petal length for each species as a grouped bar chart.
Screenshot_2026-09-16_22-06-59

ahuang11:
Can you show usage with mosaic's interactivity e.g.

image

image

SuMayaBee:
@ahuang11 There is some usage with mosaic's interactivity on iris dataset.

1. Make the scatter plot interactive. Add a brush selection, a PetalWidthCm histogram, and a data table. Update the histogram and table when I select points in the scatter plot.

https://github.com/user-attachments/assets/e4da6f5c-4e73-4703-8f26-eaca71b174f5




2. Add a slider called “Maximum petal length”. Filter the scatter plot, histogram, and table so they show only rows whose PetalLengthCm is at or below the selected value.

https://github.com/user-attachments/assets/7f1a43a7-55be-4974-9ef5-ebaa2bf9dc71




3. Add a Species dropdown filter. When I choose a species, update the scatter plot, histogram, and table to show only that species.

https://github.com/user-attachments/assets/00812dc8-5747-4cbf-b819-ae6ecb3d325c

ahuang11:
Nice! Any way to stretch the container so it takes up the whole space?

SuMayaBee:
> Nice! Any way to stretch the container so it takes up the whole space?

Let me see, what can be done to achieve that.

ghostiee-11:
Damn!! This is cool..

ahuang11:
Is this ready now?

SuMayaBee:
> Is this ready now?

It's not ready yet, I am working on it

